PR #7, opened 2020-01-09
Implement contact models library

GiulioRomualdi:
This PR introduces the contact models library.

In details:
1. An `abstract` interface named `ContactModel` has been implemented. This class will be useful to implement different contact models
2. The  `ContinuousContactModel` implements the a continuous contact model. contactModel
3. `C++17` is now required. (i.e. the library widly uses `std::any`)


src/ContactModels/tests/ContinousContactModelTest.cpp
```diff
@@ -0,0 +1,188 @@
...
+    toEigen(linkVelocity.getAngularVec3()).setRandom();
+
+    double springCoeff = 2000.0;
+    double damperCoeff = 100.0;
+
+    double length = 0.12;
+    double width = 0.09;
+
+    ContinuousContactModel model({{"length", length}, {"width", width}},
+                                 {{"spring_coeff", springCoeff}, {"damper_coeff", damperCoeff}});
```
traversaro: This syntax is quite nice!

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContactModel.h
```diff
@@ -0,0 +1,112 @@
...
+protected:
+    iDynTree::Wrench m_contactWrench; /**< Contact wrench between the robot and the environment
+                                         expressed in mixed representation */
+
+    /** Autonomous dynamics of the contact model rate of change (i.e. given a non linear system\f$
+     * \dot{x} = f + g u\f$ the autonomous is \a f */
+    iDynTree::Vector6 m_autonomousDynamics;
+
+    /** Autonomous dynamics of the contact model rate of change (i.e. given a non linear system\f$
+     * \dot{x} = f + g u\f$ the autonomous is \a g */
```
traversaro: This comment seems to be a copy paste?

traversaro:
Ah, a minor comment: it may be possible the `unordered_set<string, any>` structure for methods that are supposed to be called at each control loop may may be not efficient in term of computation time, so if you ever want to optimize the code this could be one of the things to check.

S-Dafarra:
> it may be possible the unordered_set<string, any> structure for methods that are supposed to be called at each control loop may may be not efficient in term of computation time

Is it? Search, insertion, and removal have average constant-time complexity (https://en.cppreference.com/w/cpp/container/unordered_set)

traversaro:
> 
> 
> > it may be possible the unordered_set<string, any> structure for methods that are supposed to be called at each control loop may may be not efficient in term of computation time
> 
> Is it? Search, insertion, and removal have average constant-time complexity (https://en.cppreference.com/w/cpp/container/unordered_set)

Constant is not negligible, especially if any of those operation uses dynamic memory allocation. To use an hyperbole, If the constant time complexity is `100 ms`, it would be indeed constant, but also unusable  in practice. However I think is perfectly fine to ignore this for now and come back to it to see if it is actually relevant if we need to speed up the code.

S-Dafarra:
> Constant is not negligible, especially if any of those operation uses dynamic memory allocation. 

As long as you don't add or remove elements, I don't think it does. There could be a difference when you need to scan all the elements given that they may be stored in non-contiguous portions of memory. In that case, vectors are always the best pick.

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContinuousContactModel.h
```diff
@@ -0,0 +1,141 @@
...
+    ContinuousContactModel(const std::unordered_map<std::string, std::any>& parameters);
+
+
+    /**
+     * Set the internal state of the model.
+     * @param state std::unordered_map containing the state of the system. To compute the contact
+     * wrench between two system (i.e. a link of a robot and the environment) the following data has
+     * to be available. If not an std::runtime_error is thrown.
+     *   - \a frame_transform (iDynTree::Transform): transformation between the link and the inertial frame;
+     *   - \a null_force_transform (iDynTree::Transform): transformation corresponding to a null force;
```
S-Dafarra: In theory, there could be infinite transforms for which you get a null force right?

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContinuousContactModel.h
```diff
@@ -0,0 +1,141 @@
...
+     *   - \a twist (iDynTree::Twist): twist (expressed in mixed representation) of the link
+     */
+    void setState(const std::unordered_map<std::string, std::any>& state) final;
+
+    /**
+     * Set the parameters may depends on time/state (i.e. they are not considered constant)
```
S-Dafarra:
```suggestion
     * Set the parameters that may depend on time/state (i.e. they are not considered constant)
```

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContinuousContactModel.h
```diff
@@ -0,0 +1,141 @@
...
+     * @param state std::unordered_map containing the mutable parameters of the system. To compute the contact
+     * wrench between two system (i.e. a link of a robot and the environment) the following data has
```
S-Dafarra:
```suggestion
     * wrench between two systems (i.e. a link of a robot and the environment) the following data have
```

src/ContactModels/src/ContinuousContactModel.cpp
```diff
@@ -0,0 +1,239 @@
...
+                                 "named frame_transform");
+
+    if (!getVariable(state, "null_force_transform", m_nullForceTransform))
+        throw std::runtime_error("[ContinuousContactModel::setState] Unable to get the variable "
+                                 "named null_force_transform");
+
+    // the parameters has been update the previous quantities has to be evaluated again
+    m_isContactWrenchComputed = false;
+    m_isControlMatrixComputed = false;
+    m_isAutonomusDynamicsComputed = false;
```
S-Dafarra: To be honest I don't like too much that child classes have to set a parameter that is used in the implementation of the base class. IMHO, the setting of these parameters to false should be handled by the base class or through a method in the base class, like ``invalidateBuffers()`` or similar.

src/ContactModels/src/ContinuousContactModel.cpp
```diff
@@ -0,0 +1,239 @@
...
+    auto nullForceRotation(iDynTree::toEigen(m_nullForceTransform.getRotation()));
+
+    // compute the force
+    force = std::abs(rotation(2, 2)) * area * (m_springCoeff * (nullForcePosition - position)
+                                               - m_damperCoeff * linearVelocity);
+
+    // compute the torque
+    auto skewRe1 = iDynTree::skew(rotation.col(0));
+    auto skewRe2 = iDynTree::skew(rotation.col(1));
+    torque = std::abs(rotation(2, 2)) * area / 12 * (m_length * m_length
```
S-Dafarra: Maybe it would be better to have a link to reference how this formula is obtained.

src/ContactModels/tests/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+  ## Variable Handler
+  add_library(ContinousContactModelTestMain ContinousContactModelTestMain.cpp)
+  target_link_libraries(ContinousContactModelTestMain PRIVATE Catch2::Catch2)
+
+  add_executable(ContinousContactModelTest ContinousContactModelTest.cpp)
+  target_link_libraries(ContinousContactModelTest ContinousContactModelTestMain
+    Catch2::Catch2 BipedalLocomotionControllers::ContactModels)
+  target_compile_definitions(ContinousContactModelTest PRIVATE CATCH_CONFIG_FAST_COMPILE CATCH_CONFIG_DISABLE_MATCHERS)
+
+  add_test(NAME ContinousContactModelTest COMMAND ContinousContactModelTest)
```
S-Dafarra: It would be cool to have some CMake macro for all this Catch2 machinery.

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContinuousContactModel.h
```diff
@@ -0,0 +1,141 @@
...
+ * ground. Each point in the contact surface is subjected to an infinitesimal force \f$ f \f$ given
+ * by \f$ f=k(x_0 - x)- b x\f$, where \f$k\f$ is the spring coefficient and \f$b\f$ is the damper
+ * coefficient. \f$x\f$ and \f$x_0\f$ are, respectively, the position of the point placed in the
+ * contact surface and the point con corresponding to a null contact force written in the inertial
+ * frame. Furthermore, during the contact scenario, we assume the link acts as a rigid body. While
+ * the environment will deform. The ground characteristics are isotropic, and it can be approximated
+ * as a continuum of springs and dampers.
+ * The contact surface is supposed to be <b>rectangular</b> and the frame placed on the contact
+ * surfaced is centered in the middle of the surface, with the \a z-axis pointing upwards and the \a
+ * x-axis pointing forward (parallel to one edge of the rectangle).
```
S-Dafarra: Maybe you can define here what do you mean by length and width. In addition, I would mention that this model assumes infinite friction.

src/ContactModels/tests/ContinousContactModelTest.cpp
prashanthr05: initializers list? :D

src/ContactModels/tests/CMakeLists.txt
GiulioRomualdi: Now we have it

src/ContactModels/tests/ContinousContactModelTest.cpp
GiulioRomualdi: Yep!

CMakeLists.txt
```diff
@@ -3,14 +3,17 @@
 # GNU Lesser General Public License v2.1 or any later version.
 
 cmake_minimum_required(VERSION 3.5)
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
```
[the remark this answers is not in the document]
GiulioRomualdi: Fixed in 0c04bfe85e095b93813d663cb4a8dfbcd14d2d37

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContactModel.tpp
```diff
@@ -0,0 +1,33 @@
...
+    const auto& var = variables.find(variableName);
+
+    if (var == variables.end())
+    {
+        std::cerr << "[ContactModel::getVariable] The variables named " << variableName
+                  << " cannot be found." << std::endl;
+        return false;
+    }
+
+    variable = std::any_cast<T>(var->second);
```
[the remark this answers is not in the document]
GiulioRomualdi: Let's move to return `false` since the `exceptions` are not well handled by the language  
(as already discussed with @diegoferigo  and @S-Dafarra )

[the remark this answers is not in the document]
GiulioRomualdi: Fixed in 9d9c1c3f65001bab15385997ba62ffb76303aac5

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContinuousContactModel.h
```diff
@@ -0,0 +1,141 @@
...
+     * @param state std::unordered_map containing the mutable parameters of the system. To compute the contact
+     * wrench between two system (i.e. a link of a robot and the environment) the following data has
+     * to be available. If not an std::runtime_error is thrown.
+     *   - \a spring_coeff (double): spring coefficient associated to the model
+     *   - \a damper_coeff (double): damper coefficient associated to the model
+     */
+    void setMutableParameters(const std::unordered_map<std::string, std::any>& parameters) final;
+
+    /**
+     * Compute the force in a particular point in the contact surface.
```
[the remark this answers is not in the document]
GiulioRomualdi: It is the force applied by the environment on the robot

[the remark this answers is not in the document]
GiulioRomualdi: Fixed in db6e50cd6a64759c3f67397606aedcedb4a43f08

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContinuousContactModel.h
```diff
@@ -0,0 +1,141 @@
...
+     * @param x x-coordinate of the point expressed in a frame placed in the contact surface whose
+     * origin is the center of the surface [in meters]
+     * @param y y-coordinate of the point expressed in a frame placed in the contact surface whose
+     * origin is the center of the surface [in meters]
+     */
+    iDynTree::Force getForceAtPoint(const double& x, const double& y);
+
+    /**
+     * Compute the torque about the origin of the frame place in the contact surface whose
```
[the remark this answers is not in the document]
GiulioRomualdi: Fixed in db6e50cd6a64759c3f67397606aedcedb4a43f08

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContactModel.h
```diff
@@ -0,0 +1,112 @@
...
+    iDynTree::Matrix6x6 m_controlMatrix;
+
+    bool m_isContactWrenchComputed; /**< If true the contact wrench has been already computed */
+    bool m_isAutonomusDynamicsComputed; /**< If true the autonomous dynamics has been already computed */
```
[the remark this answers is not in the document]
GiulioRomualdi: Ops :cry:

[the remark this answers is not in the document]
GiulioRomualdi: Fixed in 430db97a3be33621074ccc0aecd220a890759173

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContactModel.h
GiulioRomualdi: Fixed in 089bce74224f3dc736bc984651332dfa1ef37b5b

src/ContactModels/tests/CMakeLists.txt
GiulioRomualdi: Fixed in https://github.com/dic-iit/bipedal-locomotion-controllers/pull/7/commits/b98eeaeb3f9a2085a6e3a09700570b1ed711733e

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContinuousContactModel.h
GiulioRomualdi: In theory yes because it depends on the velocity. I can update the documentation as 
```
     *   - \a null_force_transform (iDynTree::Transform): transformation corresponding to a null force when the velocity of the system is zero;
```

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContinuousContactModel.h
```diff
@@ -0,0 +1,141 @@
...
+    ContinuousContactModel(const std::unordered_map<std::string, std::any>& immutableParameters,
+                           const std::unordered_map<std::string, std::any>& mutableParameters);
+
+    /**
+     * Constructor. It instantiate the value of the mutable parameters and the immutable parameters.
+     * @param parameters std::unordered_map containing the immutable and mutable parameters of the
+     * system. The list of the required parameters could be found in @ref
+     * BipedalLocomotionController::ContactModels::setMutableparameters
+     */
+    ContinuousContactModel(const std::unordered_map<std::string, std::any>& parameters);
```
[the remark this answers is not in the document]
GiulioRomualdi: It sounds good. In this case, we should rethink the implementation of the class.
For instance the constructor of the `ContinuousContactModel` should become
```cpp
ContinuousContactModel::ContinuousContactModel()
     : m_twist(iDynTree::Twist::Zero())
     , m_frameTransform(iDynTree::Transform::Identity())
     , m_nullForceTransform(iDynTree::Transform::Identity())
{
     // reset matrix and vectors
     m_autonomousDynamics.zero();
     m_controlMatrix.zero();
}
```

an initialize method should be implemented 
```cpp
template <class Derived>
bool ContinuousContactModel::initialize(std::waeak_ptr<IParameterHanlder<derived> hanlder)
{
     // get the parameters
     return true;
}
```

At this point, I would avoid using the `std::unordered_map<std::string, std::any>` for passing not only the parameters but also the state.  In this case the implementation of the other function will be required. i.e. 

```cpp
void ContinuousContactModel::setNullForceFrameTransform(const iDynTree::Transform &transform)
{
    m_nullForceTransform = transform;
}
```

In this way, we also go to the direction suggested by @traversaro  in https://github.com/dic-iit/bipedal-locomotion-controllers/pull/7#issuecomment-573560373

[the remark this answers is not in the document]
S-Dafarra: Honestly, I don't see anything wrong in using the map. I would say it would be better to avoid having templates around. It could unnecessarily complexify the interface.

To this end, you could use an alias so that you are free to change the type of handler without problems in the future. Or a new derived class specific to the ``COntinuosContactModel``, something like ``ContinuosContactModelParameters`` which inherit from ``IParametersHandler``.

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContinuousContactModel.h
S-Dafarra: > In theory yes because it depends on the velocity.

I don't think I understood your remark :thinking:. In my opinion, it is not clear how to choose that transform.

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContinuousContactModel.h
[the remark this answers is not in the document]
traversaro: > Honestly, I don't see anything wrong in using the map. I would say it would be better to avoid having templates around. It could unnecessarily complexify the interface.

I agree.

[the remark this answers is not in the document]
GiulioRomualdi: I know that having a template function may complexify the interface however is the only way to use the `ContinuousContactModel` with a different implementation of the handler. Till now we have only one implementation of the handler but hopefully, in the future, a `toml` implementation will be added #26 
I  really would like to avoid using CRPT in `IParameterHandler` but I don't know to update the parameter library.

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContinuousContactModel.h
GiulioRomualdi: For us is the position given by the planner. In general, I don't know

src/ContactModels/src/ContinuousContactModel.cpp
GiulioRomualdi: I can update the `doxygen` documentation adding the explanation of the model

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContinuousContactModel.h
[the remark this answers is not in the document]
traversaro: I was thinking that at some point you would use the `IParameterHandler` to convert to `std::unordered_map<std::string, std::any>`, but perhaps it is indeed to complex.

GiulioRomualdi:
I updated the code accordingly to https://github.com/dic-iit/bipedal-locomotion-controllers/pull/28

I notice that we are implementing the `initialize()` method for each class and in general the class cannot be used till the `initialize()` function is called. 
Probably we can generalize the concept by designing a class `block` and all the other classes will inherit from that class.

This is an example of the class
```cpp
class Block
{
protected:
    enum class State
    {
         NotInitialized, Initialized, ...
    };
   State m_state;
public:
   virtual bool initialize(std::weak_ptr<IParameterHandler> handler) = 0;
};
```
We can discuss this on another issue

@S-Dafarra you can review again the code

DanielePucci:
CC @gabrielenava

src/ContactModels/src/ContinuousContactModel.cpp
GiulioRomualdi: I agree!

GiulioRomualdi: I implemented the solution you suggested in https://github.com/dic-iit/bipedal-locomotion-controllers/issues/33#issuecomment-604329915

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContinuousContactModel.h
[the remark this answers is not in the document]
GiulioRomualdi: I think that this comment has been addressed thanks to:
- https://github.com/dic-iit/bipedal-locomotion-controllers/commit/01d406c794e872a75eb967e3fd7f5236761cf955
- https://github.com/dic-iit/bipedal-locomotion-controllers/pull/7/commits/7a64ab0605ad7d7383b02f33dcd2ce0f66c2f30c

GiulioRomualdi:
I updated the code to be compliant with the latest version of `master`. And I think I also handled all @S-Dafarra comments.  

Let me know if we can merge this PR since it is blocking #46 :smile:

src/ContactModels/include/BipedalLocomotionControllers/ContactModels/ContactModel.h
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Get variable from a set of variables
+     * @param variables map containing variables
+     * @param variableName name of the variable
+     * @param variable variable
+     * @return true/false in case of success/failure
+     */
+    template <class T>
+    bool getVariable(const std::unordered_map<std::string, std::any>& variables,
```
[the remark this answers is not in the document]
GiulioRomualdi: I would do it later with another PR

S-Dafarra:
Go for me!

PR #17, opened 2020-02-20
Implement isEmpty() method in ParameterHandler library

GiulioRomualdi:
The `isEmpty()` method could be really useful to check if the handler contains any parameters.

### DIFF
--- a/src/ParametersHandler/YarpImplementation/include/BipedalLocomotionControllers/ParametersHandler/YarpImplementation.h
+++ b/src/ParametersHandler/YarpImplementation/include/BipedalLocomotionControllers/ParametersHandler/YarpImplementation.h
@@ -84,6 +84,12 @@ public:
     std::string toString() const;
 
     /**
+     * Check if the handler contains parameters
+     * @return true of the the handler does not contains any parameters, false otherwise
+     */
+    bool isEmpty() const;
+
+    /**
      * Destructor
      */
     ~YarpImplementation() = default;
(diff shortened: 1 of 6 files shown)


src/ParametersHandler/YarpImplementation/include/BipedalLocomotionControllers/ParametersHandler/YarpImplementation.h
```diff
@@ -83,6 +83,12 @@ class YarpImplementation : public IParametersHandler<YarpImplementation>
      */
     std::string toString() const;
 
+    /**
+     * Check if the handler contains parameters
+     * @return true of the the handler does not contains any parameters, false otherwise
```
S-Dafarra:
```suggestion
     * @return true if the handler does not contain any parameters, false otherwise
```

src/ParametersHandler/YarpImplementation/src/YarpImplementation.cpp
```diff
@@ -30,3 +30,9 @@ std::string YarpImplementation::toString() const
 {
     return m_container.toString();
 }
+
+bool YarpImplementation::isEmpty() const
+{
+    // if the toString method return an empty string means that the handler is empty
```
S-Dafarra:
```suggestion
    // if the toString method returns an empty string means that the handler is empty
```

src/ParametersHandler/include/BipedalLocomotionControllers/ParametersHandler/IParametersHandler.h
```diff
@@ -66,6 +66,14 @@ template <class Derived> class IParametersHandler
      */
     std::string toString() const;
 
+    /**
+     * Check if the handler contains parameters
+     * @return true of the the handler does not contains any parameters, false otherwise
```
S-Dafarra:
```suggestion
     * @return true if the handler does not contain any parameters, false otherwise
```

PR #46, opened 2020-05-20
Implement the DynamicalSystem class

GiulioRomualdi:
This PR introduces the `DynamicalSystem` class. The idea is really similar to the one implemented in [`iDynTree::optimalcontrol`](https://robotology.github.io/idyntree/master/classiDynTree_1_1optimalcontrol_1_1DynamicalSystem.html). However, differently from the `iDynTree` implementation, I add the support to non-linear systems whose state space does not belong to Rⁿ. This can be done by concatenating `matrices`, `vectors` (in general classes) using `std::tuple<>`.

As @traversaro commented in https://github.com/dic-iit/bipedal-locomotion-framework/pull/46#issuecomment-632015393, I try to give a detailed explanation of the PR.
This PR introduces two main concepts the **dynamical system** and the **integrator**. As @S-Dafarra suggested in https://github.com/dic-iit/bipedal-locomotion-framework/pull/16#discussion_r393013646 these two classes will be required to implement the simulator.

## DynamicalSystem
A dynamical system is a system in which a mathematical function describes the time dependence of a state subjected to certain inputs in a geometrical space. In general, the state of the dynamical system may belong to different spaces. The simplest example is the set of the real vectors (Rⁿ ), however other dynamical systems may be described by using different kinds of sets. For instance, if we want to consider the dynamics of a rotation matrix we cannot rely on vectors since an element of the state space belongs to SO(3). Another complex system is a floating base robot. 
As we know the dynamics of this kind of system can be described by:
<p align="center">
M ( q ) d ν / d t +  h(q, ν) = Bτ + ∑Jᵀ F  
</p>

where ν is a triplet containing the base linear and angular velocity and the joints velocity. q is a triplet containing the joint positions, base position, and a rotation matrix.
In this case, the state space is much more complex than a simple Rⁿ.

In light of the previous considerations, I decided to add the possibility to define dynamical systems whose state belongs to different state spaces. So I defined the `DynamiacalSystem` as a template class where the templates parameters are:
- `State`: this type must be and `std::tuple` containing the list of vector/matrices (in general classes) that are used to represent the state space;
- `StateDerivative`: this type must be and `std::tuple` containing the list of vector/matrices (in general classes) that are used to represent the state space derivative;
- `Input`: this type must be and `std::tuple` containing the list of vector/matrices (in general classes) that are used to represent the input space.

The `DynamicalSystem` class is virtual and it contains a pure virtual method called `dynamics` that has to be implemented in a custom Dynamical system.

### `LinearTimeInvariantSystem`
I also implemented the `LinearTimeInvariantSystem` class that inherits from the `DynamicalSystem`.
A Linear time-invariant system describes a model driven by the following dynamical equation:
<p align="center">
d x / d t = A x + B u
</p>

where x and u belongs to a vectorial space. In this case the custom types are:
```cpp
using StateType = std::tuple< iDynTree::VectorDynSize >
using StateDerivativeType = std::tuple< iDynTree::VectorDynSize >
using InputType = std::tuple< iDynTree::VectorDynSize >
```

### `FloatingBaseDynamicalSystem` 

The `FloatingBaseDynamicalSystem` inherits from a generic `DynamicalSystem` and it describes a system driven by: 

<p align="center">
M ( q ) d ν / d t +  h(q, ν) = Bτ + ∑Jᵀ F  
</p>

Here the:
- `DynamicalSystem::StateType` is described by an `std::tuple` containing:
   - `iDynTree::Vector6`: the base velocity expressed in mixed representation;
    - `iDynTree::VectorDynsize`: the joint velocities [in rad/s];
    - `iDynTree::Position`: position of the base w.r.t. the inertial frame
    - `iDynTree::Rotation`: rotation matrix IRb. Matrix that transform a vector whose coordinates are expressed in the base frame in the inertial frame;
    - `iDynTree::VectorDynsize`: the joint positions [in rad].
- `DynamicalSystem::StateDerivativeType` is described by an `std::tuple` containing:
    - `iDynTree::Vector6`: the base acceleration expressed in mixed representation;
    - `iDynTree::VectorDynsize`: the joint accelerations [in rad/s^2];
    - `iDynTree::Vector3`: base velocity w.r.t. the inertial frame;
    - `iDynTree::Matrix3x3`: rate of change of the rotation matrix IR˙b. whose coordinates are expressed in the base frame in the inertial frame;
    - `iDynTree::VectorDynsize`: the joint velocities [in rad/s].
- `DynamicalSystem::InputType` is described by an std::tuple containing:
    -  `iDynTree::VectorDynsize`: the joint torques [in Nm];
    - `std::vector<ContactWrench>`: List of contact wrenches.

Here you can file the `DynamicalSystem` inherit graph generated using `Doxygen`.
classBipedalLocomotion_1_1System_1_1DynamicalSystem__inherit__graph

## Integrator
The integrator is a class that takes a dynamic system and computes the integral. To be compatible with every kind of dynamical system, it is a template class whose template parameter is a DynamicalSystem. The `Integrator` is also a virtual class since we can have different kinds of integrators (fixed steps/variable steps). The user has to implement the `integrate` function in the derived class. 
For the time being, I only implemented the `ForwardEuler` integrator. It should be enough to implement the first draft of the simulator (see here #16).

Here you can file the `Integrator` inherit graph generated using `Doxygen`.
<p align="center">
Integrator
</p>

@traversaro @S-Dafarra @MiladShafiee @Giulero

### DIFF
--- a/cmake/BipedalLocomotionFrameworkFindDependencies.cmake
+++ b/cmake/BipedalLocomotionFrameworkFindDependencies.cmake
@@ -147,3 +147,7 @@ framework_dependent_option(FRAMEWORK_COMPILE_Planners
 framework_dependent_option(FRAMEWORK_COMPILE_ContactModels
   "Compile ContactModels library?" ON
   FRAMEWORK_HAS_Eigen3 OFF)
+
+framework_dependent_option(FRAMEWORK_COMPILE_System
+  "Compile Estimators library?" ON
+  "FRAMEWORK_HAS_Eigen3;FRAMEWORK_COMPILE_ContactModels" OFF)
--- a/cmake/BipedalLocomotionFrameworkFindDependencies.cmake
+++ b/cmake/BipedalLocomotionFrameworkFindDependencies.cmake
@@ -149,5 +149,5 @@ framework_dependent_option(FRAMEWORK_COMPILE_ContactModels
   FRAMEWORK_HAS_Eigen3 OFF)
 
 framework_dependent_option(FRAMEWORK_COMPILE_System
-  "Compile Estimators library?" ON
+  "Compile System library?" ON
   "FRAMEWORK_HAS_Eigen3;FRAMEWORK_COMPILE_ContactModels" OFF)
--- a/src/System/include/BipedalLocomotion/System/ContactWrench.h
+++ b/src/System/include/BipedalLocomotion/System/ContactWrench.h
@@ -10,6 +10,7 @@
 
 #include <iDynTree/Core/Wrench.h>
 #include <iDynTree/Model/Indices.h>
+#include <BipedalLocomotion/ContactModels/ContactModel.h>
 
 namespace BipedalLocomotion
 {
@@ -23,15 +24,17 @@ namespace System
 class ContactWrench
 {
     iDynTree::FrameIndex m_frame; /**< Useful for identifying the variable in the Model */
-    iDynTree::Wrench m_wrench; /**< Wrench expressed in mixed representation */
+    std::shared_ptr<ContactModels::ContactModel> m_contactModel; /**< Contact model */
 
 public:
 
     /**
      * Constructor
      * @param index index of the frame
+     * @param wrench the contact wrench
      */
-    ContactWrench(const iDynTree::FrameIndex& index);
+    ContactWrench(const iDynTree::FrameIndex& index,
+                  std::shared_ptr<ContactModels::ContactModel> model);
 
     /**
      * Get the index of the frame
@@ -47,15 +50,9 @@ public:
 
     /**
      * Get the contact wrench
-     * @return a reference to the parameter
-     */
-    iDynTree::Wrench& wrench() noexcept;
-
-    /**
-     * Get the contact wrench
      * @return a const reference to the parameter
      */
-    const iDynTree::Wrench& wrench() const noexcept;
+    const std::shared_ptr<ContactModels::ContactModel>& contactModel() const noexcept;
 };
 
 } // namespace System
--- a/src/System/include/BipedalLocomotion/System/ContactWrench.h
+++ b/src/System/include/BipedalLocomotion/System/ContactWrench.h
@@ -1,0 +1,64 @@
+/**
+ * @file ContactWrench.h
+ * @authors Giulio Romualdi
+ * @copyright 2020 Istituto Italiano di Tecnologia (IIT). This software may be modified and
+ * distributed under the terms of the GNU Lesser General Public License v2.1 or any later version.
+ */
+
+#ifndef BIPEDAL_LOCOMOTION_SYSTEM_CONTACT_WRENCH_H
+#define BIPEDAL_LOCOMOTION_SYSTEM_CONTACT_WRENCH_H
+
+#include <iDynTree/Core/Wrench.h>
+#include <iDynTree/Model/Indices.h>
+
+namespace BipedalLocomotion
+{
+
+namespace System
+{
+
+/**
+ * A wrench excerted on a link due to an external contact.
+ */
+class ContactWrench
+{
+    iDynTree::FrameIndex m_frame; /**< Useful for identifying the variable in the Model */
+    iDynTree::Wrench m_wrench; /**< Wrench expressed in mixed representation */
+
+public:
+
+    /**
+     * Constructor
+     * @param index index of the frame
+     */
+    ContactWrench(const iDynTree::FrameIndex& index);
+
+    /**
+     * Get the index of the frame
+     * @return a reference to the index of the frame
+     */
+    iDynTree::FrameIndex& index() noexcept;
+
+    /**
+     * Get the index of the frame
+     * @return a const reference to the index of the frame
+     */
+    const iDynTree::FrameIndex& index() const noexcept;
+
+    /**
+     * Get the contact wrench
+     * @return a reference to the parameter
+     */
+    iDynTree::Wrench& wrench() noexcept;
+
+    /**
+     * Get the contact wrench
+     * @return a const reference to the parameter
+     */
+    const iDynTree::Wrench& wrench() const noexcept;
+};
+
+} // namespace System
+} // namespace BipedalLocomotion
+
+#endif // BIPEDAL_LOCOMOTION_SYSTEM_CONTACT_WRENCH_H
--- a/src/System/include/BipedalLocomotion/System/ContactWrench.h
+++ b/src/System/include/BipedalLocomotion/System/ContactWrench.h
@@ -52,7 +52,7 @@ public:
      * Get the contact wrench
      * @return a const reference to the parameter
      */
-    const std::shared_ptr<ContactModels::ContactModel>& contactModel() const noexcept;
+    const std::weak_ptr<ContactModels::ContactModel> contactModel() const noexcept;
 };
 
 } // namespace System
--- a/src/System/include/BipedalLocomotion/System/DynamicalSystem.h
+++ b/src/System/include/BipedalLocomotion/System/DynamicalSystem.h
@@ -19,41 +19,39 @@ namespace BipedalLocomotion
 namespace System
 {
 
-// Declaration of a template
-template <typename StateTypeList, typename DerivativeTypeList, typename InputTypeList>
-class DynamicalSystem
-{
-    static_assert(dependent_false<StateTypeList>::value,
-                  "Unable to create the following dynamical system.");
-};
-
 /**
  * DynamicalSystem defines a continuos time dynamical system, i.e. \f$\dot{x}=f(x, u, t)\f$. Please
- * inherit publicly from this class  in order to define your custom dynamical system.
- * @tparam StateTypes types used for describing the state (i.e. it may be a list of
- * vectors/matrices).
- * @tparam StateDerivativeTypes types used for describing the state derivative (i.e. it may be a
- * list of vectors/matrices).
- * @tparam InputTypes types used for describing the input (i.e. it may be a list
- * of vectors/matrices or in general classes).
+ * inherit publicly from this class in order to define your custom dynamical system.
+ * @tparam State type used for describing the state (i.e. it has to be a std::tuple vectors/matrices
+ * or in general classes).
+ * @tparam StateDerivative type used for describing the state derivative (i.e. it has to be a
+ * std::tuple of vectors/matrices or in general classes).
+ * @tparam Input type used for describing the input (i.e. it has to be a std::tuple of
+ * vectors/matrices or in general classes).
  */
-template <typename... StateTypes, typename... StateDerivativeTypes, typename... InputTypes>
-class DynamicalSystem<std::tuple<StateTypes...>,
-                      std::tuple<StateDerivativeTypes...>,
-                      std::tuple<InputTypes...>>
+template <typename State, typename StateDerivative, typename Input>
+class DynamicalSystem
 {
-public:
+    static_assert(is_specialization<State, std::tuple>::value);
+    static_assert(is_specialization<StateDerivative, std::tuple>::value);
+    static_assert(is_specialization<Input, std::tuple>::value);
 
-    using StateType = std::tuple<StateTypes...>;
-    using StateDerivativeType = std::tuple<StateDerivativeTypes...>;
-    using InputType = std::tuple<InputTypes...>;
+public:
+    using StateType = State; /**< State space type */
+    using StateDerivativeType = StateDerivative; /**< State space derivative type */
+    using InputType = Input; /**< Input type */
 
 protected:
-    StateType m_initialState;
-    InputType m_controlInput;
+    StateType m_initialState; /**< Value of the initial state */
+    InputType m_controlInput; /**< Value of the control input */
 
 public:
 
+    /**
+     * Initialize the Dynamical system.
+     * @param handler pointer to the parameter handler.
+     * @return true in case of success/false otherwise.
+     */
     virtual bool initalize(std::weak_ptr<ParametersHandler::IParametersHandler> handler);
 
     /**
@@ -83,11 +81,12 @@ public:
     virtual bool setControlInput(const InputType& controlInput);
 
     /**
-     * Computes the system dynamics. It return \f$f(x, u, t)\f&.
+     * Computes the system dynamics. It return \f$f(x, u, t)\f$.
      * @note The control input has to be set separately with the method setControlInput.
      * @param state tuple containing a const reference to the state elements.
      * @param time the time at witch the dynamics is computed.
      * @param stateDynamics tuple containing a reference to the element of the state derivative
+     * @warning Please implement the function in your custom dynamical system.
      * @return true in case of success, false otherwise.
      */
     virtual bool dynamics(const StateType& state,
--- a/src/System/include/BipedalLocomotion/System/DynamicalSystem.h
+++ b/src/System/include/BipedalLocomotion/System/DynamicalSystem.h
@@ -20,7 +20,7 @@ namespace System
 {
 
 /**
- * DynamicalSystem defines a continuos time dynamical system, i.e. \f$\dot{x}=f(x, u, t)\f$. Please
+ * DynamicalSystem defines a continuous time dynamical system, i.e. \f$\dot{x}=f(x, u, t)\f$. Please
  * inherit publicly from this class in order to define your custom dynamical system.
  * @tparam State type used for describing the state (i.e. it has to be a std::tuple vectors/matrices
  * or in general classes).
--- a/src/System/include/BipedalLocomotion/System/DynamicalSystem.h
+++ b/src/System/include/BipedalLocomotion/System/DynamicalSystem.h
@@ -1,0 +1,93 @@
+/**
+ * @file DynamicalSystem.h
+ * @authors Giulio Romualdi
+ * @copyright 2020 Istituto Italiano di Tecnologia (IIT). This software may be modified and
+ * distributed under the terms of the GNU Lesser General Public License v2.1 or any later version.
+ */
+
+#ifndef BIPEDAL_LOCOMOTION_SYSTEM_DYNAMICAL_SYSTEM_H
+#define BIPEDAL_LOCOMOTION_SYSTEM_DYNAMICAL_SYSTEM_H
+
+#include <memory>
+#include <tuple>
+
+#include <BipedalLocomotion/GenericContainer/TemplateHelpers.h>
+#include <BipedalLocomotion/ParametersHandler/IParametersHandler.h>
+
+namespace BipedalLocomotion
+{
+namespace System
+{
+
+// Declaration of a template
+template <typename StateTypeList, typename DerivativeTypeList, typename InputTypeList>
+class DynamicalSystem
+{
+    static_assert(dependent_false<StateTypeList>::value,
+                  "Unable to create the following dynamical system.");
+};
+
+/**
+ * DynamicalSystem defines a continuos time dynamical system, i.e. \f$\dot{x}=f(x, u, t)\f$. Please
+ * inherit publicly from this class  in order to define your custom dynamical system.
+ * @tparam StateTypes types used for describing the state (i.e. it may be a list of
+ * vectors/matrices).
+ * @tparam StateDerivativeTypes types used for describing the state derivative (i.e. it may be a
+ * list of vectors/matrices).
+ * @tparam InputTypes types used for describing the input (i.e. it may be a list
+ * of vectors/matrices or in general classes).
+ */
+template <typename... StateTypes, typename... StateDerivativeTypes, typename... InputTypes>
+class DynamicalSystem<std::tuple<StateTypes...>,
+                      std::tuple<StateDerivativeTypes...>,
+                      std::tuple<InputTypes...>>
+{
+protected:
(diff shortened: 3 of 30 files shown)


traversaro:
I did not checked in detail the proposal, but in some cases heavily use compile time polymorphism may prevent the easy reuse and introspection of the code at runtime. For this reason, it may be useful for review to:
* Specify the intended use of this interface. Just integration of a single system, or other uses such as combination of multiple system in a bigger diagram, or planning/optimal control? Having a single use case is perfectly fine (it typically simplifies a lot the design problem) but it should clearly stated. 
* Provide at least a sketch of how a generic piece of code will interface with the `DynamicalSystem`, for example a simple forward-euler "integrator". 

As an historic example (even if in their case the design goals were probably different) back in 2015 Drake started moving a lot of their "system" code to use compile time polymorphism, in particular to represent interconnection of primitive systems, see this PR and related issue for an example: 
* https://github.com/RobotLocomotion/drake/pull/1512/files
* https://github.com/RobotLocomotion/drake/issues/1522

However then in 2016 when TRI started working on Drake, for their design goal the compile time polymorphism abstraction was not flexible enough (at least for their goals, that may be different from yours), and they started working on "System 2.0" using traditional composition and compile time polymorphism, that is what is still the System abstraction of Drake nowadays. For an example, the first PR introducing the "System 2.0" is: 
* https://github.com/RobotLocomotion/drake/pull/2308

To c

traversaro:
The bipedal-locomotin-framework is already full of classes that runs synchronously with the fixed time step of the controller and have inputs, state and outputs, that are basically discrete dynamical systems, so it could make sense to call the class that in this PR is `DynamicalSystem` as `ContinuousDynamicalSystem` .

GiulioRomualdi:
Following @traversaro suggestion https://github.com/dic-iit/bipedal-locomotion-framework/pull/46#issuecomment-632015393, I improved the description of the PR in the first comment: https://github.com/dic-iit/bipedal-locomotion-framework/pull/46#issue-420998933.

src/System/include/BipedalLocomotion/System/ForwardEuler.h
```diff
@@ -0,0 +1,81 @@
...
+
+    template <std::size_t I = 0, typename... Tp, typename... Td>
+        inline typename std::enable_if < I<sizeof...(Tp), void>::type
+        addArea(const std::tuple<Tp...>& dx, const double& dT, std::tuple<Td...>& x)
+    {
+        static_assert(sizeof...(Tp) == sizeof...(Td));
+
+        iDynTree::toEigen(std::get<I>(x)) += iDynTree::toEigen(std::get<I>(dx)) * dT;
+        addArea<I + 1>(dx, dT, x);
+    }
```
traversaro, line 49: What is the point of this functions? Why they are called `addArea` ?

src/System/include/BipedalLocomotion/System/FloatingBaseDynamicalSystem.h
```diff
@@ -0,0 +1,116 @@
...
+                                                                      std::vector<ContactWrench>>>
+{
+    static constexpr size_t m_baseDoFs = 6; /**< Number of degree of freedom associated to the
+                                               floating base */
+
+    std::shared_ptr<iDynTree::KinDynComputations> m_kinDyn; /**< Pointer to an existing instance of
+                                                               kinDynComputations object */
+    std::size_t m_actuatedDoFs{0}; /**< Number of actuated degree of freedom */
+
+    iDynTree::Vector3 m_gravity; /**< Gravity vector */
```
traversaro, line 69: Personal experience: offer a way of setting this, as the hardcoded -9.81 will bite you eventually .

src/System/include/BipedalLocomotion/System/ForwardEuler.h
traversaro: I think I got it know, it is to iterate over the elements of the tuple at compile time. Probably a short one line comment or a most descriptive name would help.

src/System/include/BipedalLocomotion/System/DynamicalSystem.h
```diff
@@ -0,0 +1,107 @@
...
+#include <BipedalLocomotion/GenericContainer/TemplateHelpers.h>
+#include <BipedalLocomotion/ParametersHandler/IParametersHandler.h>
+
+namespace BipedalLocomotion
+{
+namespace System
+{
+
+/**
+ * DynamicalSystem defines a continuos time dynamical system, i.e. \f$\dot{x}=f(x, u, t)\f$. Please
```
traversaro, line 23:
```suggestion
 * DynamicalSystem defines a continuous time dynamical system, i.e. \f$\dot{x}=f(x, u, t)\f$. Please
```

traversaro:
Thanks a lot, with the additional comments it make much more sense. In general I am a bit dubious of the flexibility of an highly templated interface, but we will discover if it is working or not just by using it.

traversaro:
Ah, you are probably are already aware of this, but if you represent $SO(3)$ with a $\mathbb{R}^{3\times3}$, if the underlying integrator works on $\mathbb{R}^{3\times3}$, then it is possible that the numerical integration error drives the state $\mathbb{R}^{3\times3}$ outside of the constraint $R R^T = 1$. Within the proposed infrastructure you can deal with this by adding a stabilization term in the dynamics.

GiulioRomualdi:
@S-Dafarra the PR is now ready to be reviewed. As soon as we merge it I will open another PR for the simulator

src/System/include/BipedalLocomotion/System/ForwardEuler.h
traversaro: @GiulioRomualdi I am still curious about why the method is called `addArea`.

cmake/BipedalLocomotionFrameworkFindDependencies.cmake
```diff
@@ -147,3 +147,7 @@ framework_dependent_option(FRAMEWORK_COMPILE_Planners
 framework_dependent_option(FRAMEWORK_COMPILE_ContactModels
   "Compile ContactModels library?" ON
   FRAMEWORK_HAS_Eigen3 OFF)
+
+framework_dependent_option(FRAMEWORK_COMPILE_System
+  "Compile Estimators library?" ON
```
S-Dafarra, line 152: Copy paste? :grin:

src/System/include/BipedalLocomotion/System/ContactWrench.h
```diff
@@ -0,0 +1,61 @@
...
+ * @authors Giulio Romualdi
+ * @copyright 2020 Istituto Italiano di Tecnologia (IIT). This software may be modified and
+ * distributed under the terms of the GNU Lesser General Public License v2.1 or any later version.
+ */
+
+#ifndef BIPEDAL_LOCOMOTION_SYSTEM_CONTACT_WRENCH_H
+#define BIPEDAL_LOCOMOTION_SYSTEM_CONTACT_WRENCH_H
+
+#include <iDynTree/Core/Wrench.h>
+#include <iDynTree/Model/Indices.h>
```
S-Dafarra, line 12: Note that ``idyntree-model`` is linked privately.

src/System/include/BipedalLocomotion/System/ContactWrench.h
```diff
@@ -0,0 +1,61 @@
...
+     * Get the index of the frame
+     * @return a const reference to the index of the frame
+     */
+    const iDynTree::FrameIndex& index() const noexcept;
+
+    /**
+     * Get the contact wrench
+     * @return a const reference to the parameter
+     */
+    const std::shared_ptr<ContactModels::ContactModel>& contactModel() const noexcept;
```
S-Dafarra, line 55: Since you are setting the contact model from a shared pointer, it is better to return a weak pointer here to avoid the circular issue.

src/System/include/BipedalLocomotion/System/DynamicalSystem.h
```diff
@@ -0,0 +1,107 @@
...
+ * std::tuple of vectors/matrices or in general classes).
+ * @tparam Input type used for describing the input (i.e. it has to be a std::tuple of
+ * vectors/matrices or in general classes).
+ */
+template <typename State, typename StateDerivative, typename Input>
+class DynamicalSystem
+{
+    static_assert(is_specialization<State, std::tuple>::value);
+    static_assert(is_specialization<StateDerivative, std::tuple>::value);
+    static_assert(is_specialization<Input, std::tuple>::value);
```
S-Dafarra, line 37: Maybe you can add some message about the error.

src/System/include/BipedalLocomotion/System/FixStepIntegrator.h
```diff
@@ -0,0 +1,73 @@
...
+namespace System
+{
+
+/**
+ * Fixed step integrator base class. Please inherit publicly from this class in order to specify a
+ * custom integration method.
+ * @tparam DynamicalSystemDerived a class derived from DynamicalSystem
+ */
+template <typename DynamicalSystemDerived>
+class FixStepIntegrator : public Integrator<DynamicalSystemDerived>
```
S-Dafarra, line 27: A picky useless comment. It should be "FixedStepIntegrator" :grin:

S-Dafarra, line 27: Does it make sense to inherit from ``Advanceable`` too?

S-Dafarra: By further thinking about it and after a quick F2F it appeared it would make more sense to have an object including both a dynamical system and a fixed step integrator as ``Advanceable``, but this is out of scope.

src/System/include/BipedalLocomotion/System/ForwardEuler.h
GiulioRomualdi: I'm sure that we can improve the name. The method is named `addArea` because it adds the area under the curve to compute the integral
image

Since I don't like this name, I'm open to suggestions

cmake/BipedalLocomotionFrameworkFindDependencies.cmake
GiulioRomualdi: Fixed in 743c4b383cf7b41f7e5c1acb7270261fac1e0f16

src/System/include/BipedalLocomotion/System/ContactWrench.h
GiulioRomualdi: Fixed in a8f8df6fd4f2ab22e28ca491413d071d76a5a29c

src/System/include/BipedalLocomotion/System/ContactWrench.h
GiulioRomualdi: The function now return an `std::weak_ptr` 5db82f3b770ba5b6cb6727f4c00aaee3d0dea56d

src/System/include/BipedalLocomotion/System/DynamicalSystem.h
GiulioRomualdi: Explanation improved in 5db82f3b770ba5b6cb6727f4c00aaee3d0dea56d

src/System/include/BipedalLocomotion/System/FixStepIntegrator.h
GiulioRomualdi: Done in 0e5374b3ff057d8893d28f38d38d5b8e266fbf7d

GiulioRomualdi:
@S-dafarra I'm ready for the second part of the review

src/System/include/BipedalLocomotion/System/ForwardEuler.h
```diff
@@ -0,0 +1,93 @@
...
+        static_assert(sizeof...(Tp) == sizeof...(Td));
+    }
+
+    template <std::size_t I = 0, typename... Tp, typename... Td>
+        inline typename std::enable_if < I<sizeof...(Tp), void>::type
+        addArea(const std::tuple<Tp...>& dx, const double& dT, std::tuple<Td...>& x)
+    {
+        static_assert(sizeof...(Tp) == sizeof...(Td));
+
+        iDynTree::toEigen(std::get<I>(x)) += iDynTree::toEigen(std::get<I>(dx)) * dT;
```
S-Dafarra, line 48: Notice that this requires Eigen publicly. It also limits you basically to only ``iDynTree`` types. Does it make sense to assume Eigen types here directly? Or assume that elements can be summed?

src/System/include/BipedalLocomotion/System/LinearTimeInvariantSystem.h
```diff
@@ -0,0 +1,75 @@
...
+    bool m_isInitialized{false};
+
+public:
+    /**
+     * Set the system matrices.
+     * @param A the A matrix.
+     * @param B the B matrix.
+     * @return true in case of success, false otherwise.
+     */
+    bool setSystemMatrices(const iDynTree::MatrixDynSize& A, const iDynTree::MatrixDynSize& b);
```
S-Dafarra, line 52: Please use consistent capitalization between ``A`` and ``b``.

src/System/src/FloatingBaseSystemDynamics.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+#include <BipedalLocomotion/System/FloatingBaseSystemDynamics.h>
+
+using namespace BipedalLocomotion::System;
+
+FloatingBaseDynamicalSystem::FloatingBaseDynamicalSystem()
+{
+    // set the gravity vector
+    m_gravity.zero();
+    m_gravity(2) = -9.81;
```
S-Dafarra, line 20: Here you are assuming a specific orientation for the inertial frame.

src/System/src/FloatingBaseSystemDynamics.cpp
```diff
@@ -0,0 +1,182 @@
...
+    iDynTree::toEigen(m_knownCoefficent).tail(m_actuatedDoFs) += iDynTree::toEigen(jointTorques);
+
+    // resize the joint acceleration
+    jointAcceleration.resize(m_actuatedDoFs);
+
+    // compute the generalized robot acceleration solving the linear system. Here we assume that the
+    // mass matrix is positive definite (check here for further informations:
+    // https://eigen.tuxfamily.org/dox/group__TutorialLinearAlgebra.html)
+    iDynTree::toEigen(m_generalizedRobotAcceleration)
+        = iDynTree::toEigen(m_massMatrix).llt().solve(iDynTree::toEigen(m_knownCoefficent));
```
S-Dafarra, line 173: Consider keeping the door open for regularization terms and/or reflected inertias.

src/System/tests/IntegratorTest.cpp
```diff
@@ -0,0 +1,94 @@
...
+// Catch2
+#include <catch2/catch.hpp>
+
+#include <BipedalLocomotion/System/ForwardEuler.h>
+#include <BipedalLocomotion/System/LinearTimeInvariantSystem.h>
+
+using namespace BipedalLocomotion::System;
+
+template <typename T, typename U>
+bool areVectorsEqual(const T& vector1, const U& vector2, const double& tollerance = 0)
```
S-Dafarra, line 20: Tolerance has a single "l" :grin:

src/System/include/BipedalLocomotion/System/ForwardEuler.h
traversaro: Ah, I got it now.

src/System/include/BipedalLocomotion/System/ForwardEuler.h
GiulioRomualdi: You're right. However, in our case, we have to sum `iDynTree` elements that are defined in the `SystemDynamics` class. And the value of the state is directly retrieved from `ForwardEuler<DynamicalSystemDerived>::oneStepIntegration()`. See [here](https://github.com/dic-iit/bipedal-locomotion-framework/blob/0e5374b3ff057d8893d28f38d38d5b8e266fbf7d/src/System/include/BipedalLocomotion/System/ForwardEuler.tpp#L33).
I think that the best way to solve the problem is to define the sum operator for the `iDynTree` object (Probably passing thought `Eigen`. This may require some time

src/System/src/FloatingBaseSystemDynamics.cpp
GiulioRomualdi: Done in ac3f5e3d45a577ae242b057ec6f8952cf6b74110

src/System/include/BipedalLocomotion/System/FixedStepIntegrator.tpp
```diff
@@ -0,0 +1,74 @@
...
+    {
+        std::cerr << "[FixedStepIntegrator::integrate] The final time has to be greater than the "
+                     "initial one."
+                  << std::endl;
+        return false;
+    }
+
+    int iterations = std::ceil((finalTime - initialTime) / m_dT);
+
+    typename DynamicalSystemDerived::StateType nextState;
```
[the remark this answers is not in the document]
GiulioRomualdi: Done in fc8958744ddcf63d5271705372c8e107ec57b3cd

src/System/include/BipedalLocomotion/System/FixedStepIntegrator.tpp
```diff
@@ -0,0 +1,74 @@
...
+
+    if (initialTime > finalTime)
+    {
+        std::cerr << "[FixedStepIntegrator::integrate] The final time has to be greater than the "
+                     "initial one."
+                  << std::endl;
+        return false;
+    }
+
+    int iterations = std::ceil((finalTime - initialTime) / m_dT);
```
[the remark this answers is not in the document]
GiulioRomualdi: Done in 7ffcfcc8ff643c83bc65215a70bfc77303434f47

src/System/include/BipedalLocomotion/System/ForwardEuler.h
S-Dafarra: Ok, I guess that the easiest way out would be to include Eigen publicly also in the CMakeLists and add a line of documentation saying that only ``iDynTree`` types are supported at the moment. Unfortunately, this would limit you if you want to use it with ``CasADi`` types.

src/System/tests/IntegratorTest.cpp
GiulioRomualdi: Fixed in 1be19bfe1042e0669070bd43a1be4cce022b2fbb

src/System/src/FloatingBaseSystemDynamics.cpp
GiulioRomualdi: Yes. However in cbff248063101d95a3f146ae129d015f6c0bc6d0 I introduced the `FloatingBaseDynamicalSystem::setGravityVector()` method that allows defining a custom gravity vector. This goes also in the direction of https://github.com/dic-iit/bipedal-locomotion-framework/pull/46#discussion_r429522585

src/System/include/BipedalLocomotion/System/LinearTimeInvariantSystem.h
GiulioRomualdi: Done in 4ec34ee4e01a08cfe3cb1be98a50bb57bf5231db

src/System/include/BipedalLocomotion/System/ForwardEuler.h
GiulioRomualdi: For the time being, I would go for this 
> Ok, I guess that the easiest way out would be to include Eigen publicly also in the CMakeLists and add a line of documentation saying that only iDynTree types are supported at the moment.

Even if the CasADi compatibility It would be really nice to have. I can open an issue as soon as we merge this PR so we can discuss this specific problem in a separate location.

Related issue: https://github.com/robotology/idyntree/issues/510#issuecomment-647474988

GiulioRomualdi: Otherwise for loops may solve the problem.

S-Dafarra: > Otherwise for loops may solve the problem.

At the cost of being slower since you are not exploiting BLAS related commands. I prefer the solution of https://github.com/dic-iit/bipedal-locomotion-framework/pull/46#discussion_r443512307

GiulioRomualdi: In this case, we may obtain something like 

```cpp
if constexpr (is_data_available<T>::value && is_size_available<T>::value)
{
    auto v = std::get<I>(x);
    auto dv = std::get<I>(dx);
    for(int i = 0; i < v.size(); i++)
       v.data()[i] += dv.data()[i] * dT;
}
```

GiulioRomualdi: Done in 21b76ab0b00340ddf9f090f4eece300b164dd4a8 and 8a3c44b020142bb68236057212a05ef82144af23. The dependency is already mentioned in the documentation 
https://github.com/dic-iit/bipedal-locomotion-framework/blob/8a3c44b020142bb68236057212a05ef82144af23/src/System/include/BipedalLocomotion/System/ForwardEuler.h#L23-L31

S-Dafarra: Fine for me!

GiulioRomualdi:
@S-Dafarra let me know if you have other comments, otherwise, I will merge the PR.

PR #61, opened 2020-07-06
Implement the time-varying dcm planner

GiulioRomualdi:
This PR implements the time-varying DCM planner.

PR #65, opened 2020-07-07
Use Eigen in ConvexHullHelper class

GiulioRomualdi:
In the direction of #63

### DIFF
--- a/src/Planners/CMakeLists.txt
+++ b/src/Planners/CMakeLists.txt
@@ -14,6 +14,6 @@ if (FRAMEWORK_COMPILE_Planners)
     PRIVATE_LINK_LIBRARIES Qhull::qhullcpp Qhull::qhullstatic_r
     INSTALLATION_FOLDER    Planners
     SUBDIRECTORIES         tests
-    DEPENDS_ON_EIGEN_PRIVATE)
+    DEPENDS_ON_EIGEN_PUBLIC)
 
 endif()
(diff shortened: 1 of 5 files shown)


src/Planners/CMakeLists.txt
```diff
@@ -7,13 +7,13 @@ if (FRAMEWORK_COMPILE_Planners)
   set(H_PREFIX include/BipedalLocomotion/Planners)
 
   add_bipedal_locomotion_library(
-    NAME                   Contact
+    NAME                   Planners
```
S-Dafarra, line 10: As discussed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/61#discussion_r450776767, the definition of the Contact library was meant to provide a basic set of functions and types in order to have a common interface between planners. In principle, planners outside the framework may link this library to define footsteps or trajectories that can be used *inside* the library. For this to be effective though, it needs to be lightweight and with basically no dependencies. Why not separating ``Contact`` from ``Planners``?

GiulioRomualdi: Sorry @S-Dafarra I completely missed https://github.com/dic-iit/bipedal-locomotion-framework/pull/61#discussion_r450776767 I agree to separate `Planners` from `Contacts`

GiulioRomualdi: https://github.com/dic-iit/bipedal-locomotion-framework/pull/65/commits/37a16deb9ddafee30c47ee4b1beeaaa852c7b944 should handle the issue

Issue #56, opened 2020-06-22
Discuss the possibility to support different data types in the Integrators. 


GiulioRomualdi:
The integrator class supports only `iDynTree` objects. This limitation is due to 
https://github.com/dic-iit/bipedal-locomotion-framework/blob/8a3c44b020142bb68236057212a05ef82144af23/src/System/include/BipedalLocomotion/System/ForwardEuler.h#L48

In the next future, this may become a problem when we will extend the support the CasADi data types.

This issue is to keep track of the possible solutions.

Related issue: https://github.com/robotology/idyntree/issues/510

GiulioRomualdi:
Fixed by https://github.com/dic-iit/bipedal-locomotion-framework/pull/70

PR #70, opened 2020-07-10
Use Eigen in DynamicalSystems

GiulioRomualdi:
This goes in the direction of #56 

Unfortunately in `FloatingBaseSystemDynamics` the conversion of the Eigen vectors/matrix in iDynTree objects is done manually. I think that it can be solved with https://github.com/robotology/idyntree/issues/716

A test for the `FloatingBaseSystemDynamics` class is still missing.

### DIFF
--- a/src/System/src/FloatingBaseSystemDynamics.cpp
+++ b/src/System/src/FloatingBaseSystemDynamics.cpp
@@ -20,9 +20,9 @@ FloatingBaseDynamicalSystem::FloatingBaseDynamicalSystem()
     m_gravity(2) = -9.81;
 }
 
-void FloatingBaseDynamicalSystem::setGravityVector(const iDynTree::Vector3& gravity)
+void FloatingBaseDynamicalSystem::setGravityVector(const Eigen::Ref<const Eigen::Vector3d>& gravity)
 {
-    m_gravity = gravity;
+    iDynTree::toEigen(m_gravity) = gravity;
 }
 
 bool FloatingBaseDynamicalSystem::setKinDyn(std::shared_ptr<iDynTree::KinDynComputations> kinDyn)
@@ -48,7 +48,7 @@ bool FloatingBaseDynamicalSystem::setKinDyn(std::shared_ptr<iDynTree::KinDynComp
     return true;
 }
 
-bool FloatingBaseDynamicalSystem::setMassMatrixRegularization(const iDynTree::MatrixDynSize& matrix)
+bool FloatingBaseDynamicalSystem::setMassMatrixRegularization(const Eigen::Ref<const Eigen::MatrixXd>& matrix)
 {
 
     if (m_kinDyn == nullptr)
@@ -74,8 +74,25 @@ bool FloatingBaseDynamicalSystem::setMassMatrixRegularization(const iDynTree::Ma
     return true;
 }
 
-bool FloatingBaseDynamicalSystem::dynamics(const StateType& state,
-                                           const double& time,
+bool FloatingBaseDynamicalSystem::setState(const StateType& state)
+{
+
+    std::get<0>(m_state) = std::get<0>(state);
+    std::get<1>(m_state) = std::get<1>(state);
+    std::get<2>(m_state) = std::get<2>(state);
+    std::get<4>(m_state) = std::get<4>(state);
+
+    // project the base orientation matrix in SO3
+    // here we assume that the velocity is expressed using the mixed representation
+    const Eigen::Matrix3d& baseOrientation = std::get<3>(state);
+    Eigen::JacobiSVD<Eigen::Matrix3d> svd(baseOrientation,
+                                          Eigen::ComputeFullU | Eigen::ComputeFullV);
+    std::get<3>(m_state) = svd.matrixU() * svd.matrixV().transpose();
+
+    return true;
+}
+
+bool FloatingBaseDynamicalSystem::dynamics(const double& time,
                                            StateDerivativeType& stateDerivative)
 {
 
@@ -87,20 +104,15 @@ bool FloatingBaseDynamicalSystem::dynamics(const StateType& state,
     }
 
     // get the state
-    const iDynTree::Vector6& baseVelocity = std::get<0>(state);
-    const iDynTree::VectorDynSize& jointVelocity = std::get<1>(state);
-    const iDynTree::Position& basePosition = std::get<2>(state);
-    const iDynTree::Rotation& baseRotation = std::get<3>(state);
-    const iDynTree::VectorDynSize& jointPositions = std::get<4>(state);
+    const auto& [baseVelocity, jointVelocity, basePosition, baseOrientation, jointPositions]
+        = m_state;
 
-    // get the state derivative
-    iDynTree::Vector6& baseAcceleration = std::get<0>(stateDerivative);
-    iDynTree::VectorDynSize& jointAcceleration = std::get<1>(stateDerivative);
-    iDynTree::Vector3& baseLinearVelocity = std::get<2>(stateDerivative);
-    iDynTree::Matrix3x3& baseRotationRate = std::get<3>(stateDerivative);
-    iDynTree::VectorDynSize& jointVelocityOutput = std::get<4>(stateDerivative);
+    auto
+        & [baseAcceleration, jointAcceleration, baseLinearVelocity, baseRotationRate,
+           jointVelocityOutput]
+        = stateDerivative;
 
-    const iDynTree::VectorDynSize& jointTorques = std::get<0>(m_controlInput);
+    const Eigen::VectorXd & jointTorques = std::get<0>(m_controlInput);
     const std::vector<ContactWrench>& contactWrenches = std::get<1>(m_controlInput);
 
     // check the size of the vectors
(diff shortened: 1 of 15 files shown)


src/System/src/FloatingBaseSystemDynamics.cpp
```diff
@@ -74,8 +74,25 @@ bool FloatingBaseDynamicalSystem::setMassMatrixRegularization(const iDynTree::Ma
...
+    std::get<1>(m_state) = std::get<1>(state);
+    std::get<2>(m_state) = std::get<2>(state);
+    std::get<4>(m_state) = std::get<4>(state);
+
+    // project the base orientation matrix in SO3
+    // here we assume that the velocity is expressed using the mixed representation
+    const Eigen::Matrix3d& baseOrientation = std::get<3>(state);
+    Eigen::JacobiSVD<Eigen::Matrix3d> svd(baseOrientation,
+                                          Eigen::ComputeFullU | Eigen::ComputeFullV);
+    std::get<3>(m_state) = svd.matrixU() * svd.matrixV().transpose();
```
traversaro, line 90: This can be slow, just to understand  did you check if that is not a bit bottleneck?

GiulioRomualdi: Well @traversaro you are actually right. Half of the time is spent in the `setState()` function and the bottleneck is the `svd` calculation.

calgrind

Do you think that a [Baumgarte stabilization](https://ieeexplore.ieee.org/document/7402298?reload=true) may solve the problem?

traversaro: I don't know what the best option is at the moment, but if this PR is just about using Eigen perhaps we can leave this normalization stuff in a following PR? Or just leave the JacobiSVD  for now. 

BTW I just found https://github.com/ericjang/svd3, probably not the best option for maintainability but quite a neat idea.

S-Dafarra: Do you want to perform the integration on SO(3)? Why not passing through quaternions?

traversaro: For sure quaternion are much easier to normalize.

src/System/tests/IntegratorTest.cpp
```diff
@@ -35,60 +37,130 @@ bool areVectorsEqual(const T& vector1, const U& vector2, const double& tolerance
     return true;
 }
 
-TEST_CASE("Integrate - Linear system")
+template <typename T, typename U>
+bool areMatricesEqual(const T& matrix1, const U& matrix2, const double& tolerance = 0)
```
S-Dafarra, line 41: In Eigen you can use [``isApprox``](https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#ae8443357b808cd393be1b51974213f9c)

src/System/src/FloatingBaseSystemDynamics.cpp
S-Dafarra: Also, why you are performing this kind of computations in ``setState``? Is it to make sure that the user is actually passing a rotation matrix?

Have you considered implementing the computations for SVD manually, since the matrix is small?

src/System/src/FloatingBaseSystemDynamics.cpp
```diff
@@ -74,8 +74,25 @@ bool FloatingBaseDynamicalSystem::setMassMatrixRegularization(const iDynTree::Ma
...
 
-bool FloatingBaseDynamicalSystem::dynamics(const StateType& state,
-                                           const double& time,
+bool FloatingBaseDynamicalSystem::setState(const StateType& state)
+{
+
+    std::get<0>(m_state) = std::get<0>(state);
+    std::get<1>(m_state) = std::get<1>(state);
+    std::get<2>(m_state) = std::get<2>(state);
+    std::get<4>(m_state) = std::get<4>(state);
```
S-Dafarra, line 83: I have to say that it is a little hard to remember what is what. I suspect there could be issues where we take one element for another. After a quick googling, there are some possible implementations of "tagged tuples", where the ``get`` uses a string instead of an index. Some references:
- https://stackoverflow.com/a/46862694
- https://gist.github.com/ilsken/dd91285d50197d6345f9
- https://github.com/duckie/named_types
- https://bitbucket.org/tlutz/named-variadic-template-proposal/src/master/
- https://stackoverflow.com/questions/13065166/c11-tagged-tuple

 Clearly, this is not related to this PR and probably there is no clean solution, and I don't know if the initializer lists would work anyway, but it may be worth discussing.

src/System/src/FloatingBaseSystemDynamics.cpp
GiulioRomualdi: @S-Dafarra 
> Do you want to perform the integration on SO(3)? Why not passing through quaternions?
working on this

> Also, why you are performing this kind of computations in setState? Is it to make sure that the user is actually passing a rotation matrix?
This is the `setState()` of the `FloatingBaseSystemDynnamics` class so the user has to pass a rotation

GiulioRomualdi: Actually I can remove the svd If we switch to quaternions and I use this equation to compute the quaternion derivative
image

GiulioRomualdi: Actually using quaternion is not so trivial.

Indeed If I use `Eigen::Quaterniond` I can't do this:
```
Eigen::Quaterniond a;
a.setIdentity();
a = a + Eigen::Vector4d::Zero();
``` 
I get the following 
```
 error: no match for ‘operator+’ (operand types are ‘Eigen::Quaterniond {aka Eigen::Quaternion<double>}’ and ‘const ConstantReturnType {aka const Eigen::CwiseNullaryOp<Eigen::internal::scalar_constant_op<double>, Eigen::Matrix<double, 4, 1> >}’
```

Unfortunately, the `addArea()` is a template function and I would avoid adding template specialization there.
For the time being, I would go for the Rotation matrix + Baumgard integration.  In the future, If we want to use quaternions instead of matrices we can open a new issue and discuss it there. What do you think?

S-Dafarra: I am a little lost here. The state derivative and the state are generally supposed to be different right? In this case the state is a quaternion and the derivative a Vector4 and it makes sense for them not to be added together. Before you were dealing with the rotation matrix in a specific case. How would it work now?

src/System/tests/IntegratorTest.cpp
GiulioRomualdi: Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/70/commits/ff98d9a226054d9449e37d43705f4dfeb1347237

src/System/src/FloatingBaseSystemDynamics.cpp
GiulioRomualdi: Let me explain better. The dynamical system class contains the state, its derivative, and the input. In case of a rotation we have:
- `Eigen::Matrix3d` for the state (i.e. the rotation) 
- `Eigen::Matrix3d` for the state derivative (i.e. the rotation rate of change) 
- `Eigen::Vector3d` for the input (i.e. angular velocity)

The rate of change the rotation matrix is then computed as
<p align="center">
Ṙ = ωx R 
</p>
Ṙ is then integrated to compute R. The `ForwardEuler` integrator uses the function `addArea()` to compute the integral 
https://github.com/dic-iit/bipedal-locomotion-framework/blob/339cf76339622cf496d941258374614f43d526af/src/System/include/BipedalLocomotion/System/ForwardEuler.h#L48
This code can compile only if the `operator+()` is defined between the state and the state derivative. In this case, since the rate of change of the rotation matrix and the rotation matrix are stored in an `Eigen::Matrix3d`, the `operator+()` is well defined. 

In the case of quaternions we can decide to:
1. Store the quaternion in an `Eigen::Vector4d` and apply a similar approach. In this case, I won't have any problems;
2. Store the quaternion in an `Eigen::Quaterniond` and its rate of change in `Eigen::Vector4d`. In this case `std::get<I>(x) += std::get<I>(dx) * dT` will fail because the `operator+()` is not defined between `Eigen::quaterniond` and `Eigen::Vector4d`. 


By the way, in https://github.com/dic-iit/bipedal-locomotion-framework/pull/70/commits/9870a4f218f4f42938c66d29cb6b4bebe2a9ffa2 I implemented the Baumgarte stabilization for rotation matrix. Differently from the [original paper](https://ieeexplore.ieee.org/document/7402298?reload=true), I derived the equation in case of angular velocity expressed in the inertial frame. Thanks to this commit I was able to remove the `svd` computation from the `setState()` and consequentially I improved the overall performances of the classes.

traversaro: Great. This has also the nice property that the simulated  system remains an ODE, while with the normalization you are basically coupling an ODE with an discrete non linear instantaneous system, i.e. the block that takes in input the integrated orientation (that is  the output of the ODE) and  returns the normalized rotation instead.

S-Dafarra: That's very clear, thank you! In the hindsight, it is also pretty clean to have directly the rotation matrix derivative, nice solution.

GiulioRomualdi:
Merging

PR #73, opened 2020-07-14
Use Eigen in RecursiveLeastSquare class

GiulioRomualdi:
This PR is related to #63. Furthermore:
- `RecursiveLeastSquare` class inherits from `Advanceable` 
- `RecursiveLeastSquareTest`  does not depend on yarp anymore


GiulioRomualdi:
Actually I was thinking to change 
```cpp
void setRegressorFunction(std::function<Eigen::MatrixXd(void)> regressor);
```
into
```cpp
void setRegressor(const Eigen::Ref<const Eigen::MatrixXd>& regressor);
```

i.e. I would avoid using function pointer there. Let me know what do you think @S-Dafarra @traversaro @prashanthr05

S-Dafarra:
> Actually I was thinking to change
> 
> ```c++
> void setRegressorFunction(std::function<Eigen::MatrixXd(void)> regressor);
> ```
> 
> into
> 
> ```c++
> void setRegressor(const Eigen::Ref<const Eigen::MatrixXd>& regressor);
> ```
> 
> i.e. I would avoid using function pointer there. Let me know what do you think @S-Dafarra @traversaro @prashanthr05

Makes sense, function pointers may be slow. What about keeping them both?

GiulioRomualdi:
> Makes sense, function pointers may be slow. What about keeping them both?

I agree for keeping both methods

GiulioRomualdi:
In https://github.com/dic-iit/bipedal-locomotion-framework/pull/73/commits/2145f5c9ed6e1ea2b870f05df914619390efc9e1 I added the possibility to set the regressor matrix. If @S-Dafarra agrees on the solution I propose I will merge the PR (after the CI checks)

S-Dafarra:
> In [2145f5c](https://github.com/dic-iit/bipedal-locomotion-framework/commit/2145f5c9ed6e1ea2b870f05df914619390efc9e1) I added the possibility to set the regressor matrix. If @S-Dafarra agrees on the solution I propose I will merge the PR (after the CI checks)

Sure!

PR #77, opened 2020-07-20
Bug fix in the TimeVarying DCM planner

GiulioRomualdi:


GiulioRomualdi:
Actually the original equation was correct.

PR #78, opened 2020-07-21
Implement a Eigen-CppAD skeleton 

GiulioRomualdi:
This PR implements some utilities to use `CppAD` with `Eigen` In details:
- `Findcppad.cmake` has been designed to find `cppad` thought `pkg_config` or `cppad_DIR` Right now it works only on Linux and MacOS. 
- `AutoDiff/CppAD.h` includes and implements some utilities functions to use `CppAD` with `Eigen` (e.g. see [here](https://stackoverflow.com/questions/10088002/how-to-implement-static-cast-in-c/10088161)

**Known issue:** There seems that it is not possible to use `CppAD` inside Eigenvector fixed size [see here](https://stackoverflow.com/questions/50661842/fixed-size-eigen-matrices-in-cppad)


cmake/BipedalLocomotionFrameworkFindDependencies.cmake
```diff
@@ -144,6 +144,9 @@ checkandset_dependency(Qhull)
 find_package(casadi QUIET)
 checkandset_dependency(casadi)
 
+find_package(cppad QUIET)
```
S-Dafarra, line 147: Please add ``cppad`` as a dependency in the main CMakeLists when generating the configure file. (I know this should be automatic, but I was not able to find a way to do this cleanly).

src/AutoDiff/tests/CppADTest.cpp
```diff
@@ -0,0 +1,44 @@
...
+    // Some coefficients useful for the test
+    constexpr double a = 4;
+    constexpr double b = 1.1341;
+    constexpr double c = 2.3213;
+
+    BipedalLocomotion::AutoDiff::VectorXcppAD x(3);
+
+    // Start recording
+    CppAD::Independent(x);
+    BipedalLocomotion::AutoDiff::VectorXcppAD y = a * x.array() + c * (b * x.array()).sin();
```
S-Dafarra, line 24: To be honest ``VectorXcppAD`` seems a little ugly to me (like a fiscal code :grin:). Does it make sense to use an additional namespace, such that we have
```
BipedalLocomotion::AutoDiff::cppAD::VectorXAD
```
? Does it look any better in your opinion?

cmake/BipedalLocomotionFrameworkFindDependencies.cmake
GiulioRomualdi: Actually I didn't understand this comment.

GiulioRomualdi: Ok you're referring to this https://github.com/dic-iit/bipedal-locomotion-framework/blob/ea886f7060bcffea4e11fab645a7deec42a2476e/CMakeLists.txt#L83-L91

GiulioRomualdi: Done in be65efda73e41d3796db8d7b1682ab314ec11f6f

src/AutoDiff/tests/CppADTest.cpp
GiulioRomualdi: Done in 2814e3bb6322b95f1e05afaec30c16d6d98ad771

GiulioRomualdi:
This feature is not supported yet by Windows. As soon as https://github.com/microsoft/vcpkg/pull/12560 is merged I think we can start supporting also Windows

GiulioRomualdi:
@S-Dafarra let me know if I can merge the PR

PR #82, opened 2020-08-06
Use manif in the contacts component

GiulioRomualdi:
In this PR I removed the `iDyntree` dependency in favor of `manif`.

I also updated the CI adding `manif` dependency. Here I'm cloning manif from   https://github.com/GiulioRomualdi/manif.git because of https://github.com/artivis/manif/issues/156. I already opened a PR upstream https://github.com/artivis/manif/pull/157 as soon as it has been merged I will update this code.

I would like to wait for the review of @S-Dafarra since he is the first developer of this part of the code

### DIFF
--- a/src/Planners/src/TimeVaryingDCMPlanner.cpp
+++ b/src/Planners/src/TimeVaryingDCMPlanner.cpp
@@ -344,7 +344,7 @@ struct TimeVaryingDCMPlanner::Impl
         {
             for (const auto& footCorner : this->optiSettings.footCorners)
             {
-                point = activeContact->pose.isometry() * footCorner;
+                point = activeContact->pose.act(footCorner);
                 points.col(columnIndex) = point.head(numberOfCoordinates);
                 columnIndex++;
             }
--- a/src/Planners/src/TimeVaryingDCMPlanner.cpp
+++ b/src/Planners/src/TimeVaryingDCMPlanner.cpp
@@ -142,7 +142,7 @@ struct TimeVaryingDCMPlanner::Impl
         std::string ipoptLinearSolver{"mumps"}; /**< Linear solved used by ipopt */
 
         double plannerSamplingTime; /**< Sampling time of the planner in seconds */
-        std::vector<iDynTree::Position> footCorners; /**< Position of the corner of the foot
+        std::vector<Eigen::Vector3d> footCorners; /**< Position of the corner of the foot
                                                     expressed in a frame placed in the center of the
                                                     foot. The X axis points forward and the z
                                                     upword.*/
@@ -322,9 +322,9 @@ struct TimeVaryingDCMPlanner::Impl
         else if (contactPhase.activeContacts.size() == 2)
         {
             auto contactIt = contactPhase.activeContacts.cbegin();
-            double foot1Height = contactIt->second->pose.getPosition()[2];
+            double foot1Height = contactIt->second->pose.translation()[2];
             std::advance(contactIt, 1);
-            double foot2Height = contactIt->second->pose.getPosition()[2];
+            double foot2Height = contactIt->second->pose.translation()[2];
 
             if (foot2Height == foot1Height)
             {
@@ -339,13 +339,13 @@ struct TimeVaryingDCMPlanner::Impl
 
         // compute the convex hull only once
         int columnIndex = 0;
-        iDynTree::Position point;
+        Eigen::Vector3d point;
         for (const auto& [contactName, activeContact] : contactPhase.activeContacts)
         {
             for (const auto& footCorner : this->optiSettings.footCorners)
             {
-                point = activeContact->pose * footCorner;
-                points.col(columnIndex) = iDynTree::toEigen(point).head(numberOfCoordinates);
+                point = activeContact->pose.isometry() * footCorner;
+                points.col(columnIndex) = point.head(numberOfCoordinates);
                 columnIndex++;
             }
         }
@@ -401,7 +401,6 @@ struct TimeVaryingDCMPlanner::Impl
         casadi::DM ecmpConstraintA;
         casadi::DM ecmpConstraintB;
 
-        const auto& dcm = this->optiVariables.dcm;
         const auto& vrp = this->optiVariables.vrp;
         const auto& omega = this->optiVariables.omega;
         const auto& omegaDot = this->optiVariables.omegaDot;
@@ -432,12 +431,12 @@ struct TimeVaryingDCMPlanner::Impl
                         (contactPhaseListIt->endTime + contactPhaseListIt->beginTime) / 2);
 
                     auto contactIt = contactPhaseListIt->activeContacts.cbegin();
-                    iDynTree::Position p1 = contactIt->second->pose.getPosition();
+                    Eigen::Vector3d p1 = contactIt->second->pose.translation();
                     std::advance(contactIt, 1);
-                    iDynTree::Position p2 = contactIt->second->pose.getPosition();
+                    Eigen::Vector3d p2 = contactIt->second->pose.translation();
 
                     Eigen::Vector3d desiredDCMPosition;
-                    desiredDCMPosition = (iDynTree::toEigen(p1) + iDynTree::toEigen(p2)) / 2;
+                    desiredDCMPosition = (p1 + p2) / 2;
                     desiredDCMPosition(2) += averageDCMHeight;
 
                     dcmKnots.emplace_back(desiredDCMPosition);
@@ -450,12 +449,12 @@ struct TimeVaryingDCMPlanner::Impl
                 {
                     timeKnots.push_back(contactPhaseListIt->endTime);
                     auto contactIt = contactPhaseListIt->activeContacts.cbegin();
-                    iDynTree::Position p1 = contactIt->second->pose.getPosition();
+                    Eigen::Vector3d p1 = contactIt->second->pose.translation();
                     std::advance(contactIt, 1);
-                    iDynTree::Position p2 = contactIt->second->pose.getPosition();
+                    Eigen::Vector3d p2 = contactIt->second->pose.translation();
 
                     Eigen::Vector3d desiredDCMPosition;
-                    desiredDCMPosition = (iDynTree::toEigen(p1) + iDynTree::toEigen(p2)) / 2;
+                    desiredDCMPosition = (p1 + p2) / 2;
                     desiredDCMPosition(2) += averageDCMHeight;
 
                     dcmKnots.emplace_back(desiredDCMPosition);
@@ -476,7 +475,7 @@ struct TimeVaryingDCMPlanner::Impl
 
                 this->opti.subject_to(
                     vrp(2, k) - 9.81 / (casadi::MX::pow(omega(Sl(), k), 2) - omegaDot(Sl(), k))
-                    == contactPhaseListIt->activeContacts.begin()->second->pose.getPosition()[2]);
+                    == contactPhaseListIt->activeContacts.begin()->second->pose.translation()[2]);
             }
         }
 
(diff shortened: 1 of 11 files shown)


prashanthr05:
> In this PR I removed the iDyntree dependency in favor of manif.

Why was this necessary?

GiulioRomualdi:
https://github.com/dic-iit/bipedal-locomotion-framework/pull/82/commits/7800437e0d94b4a9dc312b5396e821c5d50be7d7 fixes a bug I introduced with this PR. 

If https://github.com/artivis/manif/issues/160 will be acepted
```cpp
leftTransform.coeffs().head<3>() << 0, -0.8, 0;
```
will become 
```cpp
leftTransform.translation() << 0, -0.8, 0;
```

GiulioRomualdi:
> Why was this necessary?

Because of #84 and for the planner of the swing foot I'm implementing right now :smile:

prashanthr05:
> > Why was this necessary?
> 
> Because of #84 and for the planner of the swing foot I'm implementing right now

Great, thanks! Could you add it in the PR description as well?  :D

src/Planners/src/TimeVaryingDCMPlanner.cpp
```diff
@@ -339,13 +339,13 @@ struct TimeVaryingDCMPlanner::Impl
...
         int columnIndex = 0;
-        iDynTree::Position point;
+        Eigen::Vector3d point;
         for (const auto& [contactName, activeContact] : contactPhase.activeContacts)
         {
             for (const auto& footCorner : this->optiSettings.footCorners)
             {
-                point = activeContact->pose * footCorner;
-                points.col(columnIndex) = iDynTree::toEigen(point).head(numberOfCoordinates);
+                point = activeContact->pose.isometry() * footCorner;
```
prashanthr05, line 347: You could also use the act() method. It must be doing the same thing.

src/Planners/tests/ContactListTest.cpp
```diff
@@ -1,28 +1,26 @@
...
-    return (c1.type == c2.type) &&
-           (c1.name == c2.name) &&
-           (iDynTree::toEigen((c1.pose * c2.pose.inverse()).asHomogeneousTransform()).isApprox(
-               iDynTree::toEigen(iDynTree::Transform::Identity().asHomogeneousTransform()))) &&
-           (iDynTree::checkDoublesAreEqual(c1.activationTime, c2.activationTime)) &&
-           (iDynTree::checkDoublesAreEqual(c1.deactivationTime, c2.deactivationTime));
+    return (c1.type == c2.type) && (c1.name == c2.name)
+           && (((c1.pose * c2.pose.inverse()).log()).isApprox(Eigen::Matrix<double, 6, 1>::Zero()))
+           && (c1.activationTime == c2.activationTime)
+           && (c1.deactivationTime == c2.deactivationTime);
```
prashanthr05, line 23: is this a strong check for doubles? Wouldn't you need a check with tolerance as it was offered by iDynTree::checkDoublesAreEqual(a, b)?

src/Planners/tests/TimeVaryingDCMPlannerTest.cpp
```diff
@@ -23,42 +21,36 @@ TEST_CASE("TimeVaryingDCMPlanner")
...
+    manif::SE3d leftTransform = manif::SE3d::Identity();
+    leftTransform.coeffs().head<3>() << 0, -0.8, 0;
     REQUIRE(contactListMap["left"].addContact(leftTransform, 0.0, 1.0));
 
-    leftPosition[0] = 0.25;
-    leftPosition[2] = 0.2;
-    leftTransform.setPosition(leftPosition);
+    // second footstep
+    leftTransform.coeffs()[0] = 0.25;
+    leftTransform.coeffs()[2] = 0.2;
```
prashanthr05, line 32: Please add a comment here as to how the coeffs() is ordered for the SE3d object. It should be 'pos+quat'. If someone modifying the tests, mistake it to be something else, then this would result in problems.

src/Planners/tests/TimeVaryingDCMPlannerTest.cpp
```diff
@@ -23,42 +21,36 @@ TEST_CASE("TimeVaryingDCMPlanner")
...
-    iDynTree::Position rightPosition;
-
-    rightPosition[0] = 0;
-    rightPosition[1] = 0.8;
-    rightPosition[2] = 0;
-    rightTransform.setPosition(rightPosition);
+    // right foot
+    // first footstep
+    manif::SE3d rightTransform = manif::SE3d::Identity();
+    leftTransform.coeffs().head<3>() << 0, 0.8, 0;
```
prashanthr05, line 38: same as above and for other coeff() related lines.

src/Planners/tests/TimeVaryingDCMPlannerTest.cpp
prashanthr05: Typical issue 7800437 fixes a bug I introduced with this PR.

src/Planners/tests/ContactListTest.cpp
S-Dafarra: Initially, this was to check if two contacts are *exactly* the same, i.e. two copies of the same object. We could try to have some coherence also with the transformations check. Either they are both checked precisely or with a tolerance.

src/Planners/tests/TimeVaryingDCMPlannerTest.cpp
S-Dafarra: Actually, I would prefer if there is an equivalent of ``setPosition``.

src/Planners/tests/TimeVaryingDCMPlannerTest.cpp
S-Dafarra: Also, shouldn't it be 
```suggestion
    rightTransform.coeffs().head<3>() << 0, 0.8, 0;
```
?

src/Planners/tests/TimeVaryingDCMPlannerTest.cpp
prashanthr05: Relevant issue https://github.com/artivis/manif/issues/160

S-Dafarra: Thanks for linking the issue, I missed it.

GiulioRomualdi:
@S-Dafarra 

> Looks good to me. The only open points would be:
> 
> * easy translation from `manif::SE3d` to `iDynTree::Transform` objects and vice-versa.

For this one I can create functions for converting the `iDynTree` object to `manif` as far as I've seen there is also the `Eigen::Map` feature in `manif`. Please see [here](https://github.com/artivis/manif/blob/devel/include/manif/impl/se3/SE3_map.h) 

> * is it possible not to use the `coeffs` method? It is really not clear what you are setting, especially for the rotation part.

You are actually right, I don't like to use `coeffs` method because is really error-prone. As @prashanthr05 already pointed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/82#discussion_r471551861, I opened an issue in the main repository https://github.com/artivis/manif/issues/160. Unfortunately, as @artivis wrote in https://github.com/artivis/manif/issues/160#issuecomment-671388708,  the solution I proposed may lead to bugs (e.g. setting a non-unitary quaternion for the rotation)

We can propose some setters like 
```cpp
void quat(const quaternion& q);
void quat(const SO3& SO3);
```

@traversaro @prashanthr05

src/Planners/src/TimeVaryingDCMPlanner.cpp
GiulioRomualdi: Done https://github.com/dic-iit/bipedal-locomotion-framework/pull/82/commits/74928a3dbf8c3bce47e90b8b87e9e4c11735d161

src/Planners/tests/TimeVaryingDCMPlannerTest.cpp
GiulioRomualdi: @S-Dafarra 
Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/82/commits/e2e3bc36514b4b761a0ba1d8734473de3694c36a

S-Dafarra:
> For this one I can create functions for converting the `iDynTree` object to `manif` as far as I've seen there is also the `Eigen::Map` feature in `manif`. Please see [here](https://github.com/artivis/manif/blob/devel/include/manif/impl/se3/SE3_map.h)

Sounds good to me.



> You are actually right, I don't like to use `coeffs` method because is really error-prone. As @prashanthr05 already pointed in [#82 (comment)](https://github.com/dic-iit/bipedal-locomotion-framework/pull/82#discussion_r471551861), I opened an issue in the main repository [artivis/manif#160](https://github.com/artivis/manif/issues/160). Unfortunately, as @artivis wrote in [artivis/manif#160 (comment)](https://github.com/artivis/manif/issues/160#issuecomment-671388708), the solution I proposed may lead to bugs (e.g. setting a non-unitary quaternion for the rotation)

I can argue that it is even easier to set non-unitary quaternion with the `coeffs` method 😁 In any case, this is just to have a "clean use-case", a sort of implicit documentation for future usages. Have you tried the SE constructors mentioned in the issue together with some copy assignment?

GiulioRomualdi:
> Have you tried the SE constructors mentioned in the issue together with some copy assignment? 

it should work but I don't like to set both a rotation and a translation if I've to change only one

GiulioRomualdi:
https://github.com/GiulioRomualdi/manif/commit/be76458b7762a8b870d2bf5ca26ec6c9a32d79ed Implements the setters proposed by @artivis

GiulioRomualdi:
In https://github.com/dic-iit/bipedal-locomotion-framework/pull/82/commits/068490e90b04caae42e2f5c2fc82c3fbe254bc33 I avoid to use the `coeffs()`  method in the tests. (it is a workaround for https://github.com/artivis/manif/issues/160) 

Regarding the `Map` is slightly more complex. `Eigen::Map<manif::SE3d>` object wants a contiguous memory 7-d vector containing the translation and the quaternion, on the other hand `iDynTree::Transform` stores the rotation and the translation in two separate objects. What I propose is to create two functions for handling the coping of a `manif` object into an `iDynTree` object and vice-versa. We can discuss this in a separate issue/PR. 

@prashanthr05 @S-Dafarra let me know if it is fine with you so I can merge the PR

S-Dafarra:
> Regarding the `Map` is slightly more complex. `Eigen::Map<manif::SE3d>` object wants a contiguous memory 7-d vector containing the translation and the quaternion, on the other hand `iDynTree::Transform` stores the rotation and the translation in two separate objects. What I propose is to create two functions for handling the coping of a `manif` object into an `iDynTree` object and vice-versa. We can discuss this in a separate issue/PR.
> 
> @prashanthr05 @S-Dafarra let me know if it is fine with you so I can merge the PR

Ok for me

S-Dafarra (approved):

src/Planners/tests/ContactListTest.cpp
GiulioRomualdi: https://github.com/dic-iit/bipedal-locomotion-framework/pull/82/commits/fba1c02cef442dee58e39f09f9eb1aa75cbe2995 goes in the direction of what @prashanthr05 suggested in https://github.com/dic-iit/bipedal-locomotion-framework/pull/82#discussion_r470568742

S-Dafarra:
Even if an approving reviewer is enough, if one of the reviewers requested changes it may be worth waiting for his approval too 😉

GiulioRomualdi:
> Even if an approving reviewer is enough, if one of the reviewers requested changes it may be worth waiting for his approval too

You're right. @prashanthr05  tell me if the change I did are fine with you, if not I open a new PR with the modifications

prashanthr05:
> You're right. @prashanthr05 tell me if the change I did are fine with you, if not I open a new PR with the modifications

:rocket: from me!!

PR #83, opened 2020-08-13
Implement the QuinticSpline class

GiulioRomualdi:
This PR introduces the `QuinticSpline` class. This class can be seen as an extension of the `iDynTree::CubicSpline` class. The main difference between this class and the `iDynTree` one is the order of the polynomials used in the spline. Using a 5th order spline guarantees to generate a trajectory that minimizes the jerk. (I will add further details for the latest statement later)

### DIFF
--- a/src/Planners/include/BipedalLocomotion/Planners/QuinticSpline.h
+++ b/src/Planners/include/BipedalLocomotion/Planners/QuinticSpline.h
@@ -1,0 +1,86 @@
+/**
+ * @file QuinticSpline.h
+ * @authors Giulio Romualdi
+ * @copyright 2020 Istituto Italiano di Tecnologia (IIT). This software may be modified and
+ * distributed under the terms of the GNU Lesser General Public License v2.1 or any later version.
+ */
+
+#ifndef BIPEDAL_LOCOMOTION_PLANNERS_QUINTIC_SPLINE_H
+#define BIPEDAL_LOCOMOTION_PLANNERS_QUINTIC_SPLINE_H
+
+#include <vector>
+#include <memory>
+
+#include <Eigen/Dense>
+
+namespace BipedalLocomotion
+{
+namespace Planners
+{
+/**
+ * Quintic spline implement a 5-th order polynomial spline in \$f\mathbb{R}^n\$f.
+ */
+class QuinticSpline
+{
+    /**
+     * Private implementation of the class
+     */
+    struct Impl;
+    std::unique_ptr<Impl> m_pimpl; /**< Private implementation */
+
+public:
+
+    /**
+     * Constructor.
+     */
+    QuinticSpline();
+
+    /**
+     * Destructor.
+     * @note It is required by the PIMPL idiom.
+     */
+    ~QuinticSpline();
+
+    /**
+     * Set the knots of the spline.
+     * @param position position of the knots in \$f\mathbb{R}^n\$f.
+     * @param time vector containing the time instant of the knots.
+     * @return True in case of success, false otherwise.
+     */
+    bool setKnots(const std::vector<Eigen::VectorXd>& position, const std::vector<double>& time);
+
+    /**
+     * Set the initial condition of the spline
+     * @param velocity initial velocity (i.e. first derivative).
+     * @param acceleration initial acceleration (i.e. second derivative).
+     * @return True in case of success, false otherwise.
+     */
+    bool setInitialConditions(Eigen::Ref<const Eigen::VectorXd> velocity,
+                              Eigen::Ref<const Eigen::VectorXd> acceleration);
+
+    /**
+     * Set the final condition of the spline
+     * @param velocity final velocity (i.e. first derivative).
+     * @param acceleration final acceleration (i.e. second derivative).
+     * @return True in case of success, false otherwise.
+     */
+    bool setFinalConditions(Eigen::Ref<const Eigen::VectorXd> velocity,
+                            Eigen::Ref<const Eigen::VectorXd> acceleration);
+
+    /**
+     * Evaluate the spline at a given point
+     * @param t instant time
+     * @param position position at time t
+     * @param velocity velocity at time t
+     * @param acceleration acceleration at time t
+     * @return True in case of success, false otherwise.
+     */
+    bool evaluatePoint(const double& t,
+                       Eigen::Ref<Eigen::VectorXd> position,
+                       Eigen::Ref<Eigen::VectorXd> velocity,
+                       Eigen::Ref<Eigen::VectorXd> acceleration);
+};
+} // namespace Planners
+} // namespace BipedalLocomotion
+
+#endif // BIPEDAL_LOCOMOTION_PLANNERS_QUINTIC_SPLINE_H
--- a/src/Planners/include/BipedalLocomotion/Planners/QuinticSpline.h
+++ b/src/Planners/include/BipedalLocomotion/Planners/QuinticSpline.h
@@ -13,14 +13,24 @@
 
 #include <Eigen/Dense>
 
+#include <BipedalLocomotion/System/Advanceable.h>
+
 namespace BipedalLocomotion
 {
 namespace Planners
 {
+
+struct QuinticSplineState
+{
+    Eigen::VectorXd position;
+    Eigen::VectorXd velocity;
+    Eigen::VectorXd acceleration;
+};
+
 /**
  * Quintic spline implement a 5-th order polynomial spline in \$f\mathbb{R}^n\$f.
  */
-class QuinticSpline
+class QuinticSpline : public System::Advanceable<QuinticSplineState>
 {
     /**
      * Private implementation of the class
@@ -29,7 +39,6 @@ class QuinticSpline
     std::unique_ptr<Impl> m_pimpl; /**< Private implementation */
 
 public:
-
     /**
      * Constructor.
      */
@@ -40,6 +49,14 @@ public:
      * @note It is required by the PIMPL idiom.
      */
     ~QuinticSpline();
+
+    /**
+     * Set the time step of the advance interface.
+     * @warning if the the time step is not set the user cannot use the advance features.
+     * @param dt the time step of the advance block.
+     * @return True in case of success, false otherwise.
+     */
+    bool setAdvanceTimeStep(const double& dt);
 
     /**
      * Set the knots of the spline.
@@ -79,6 +96,40 @@ public:
                        Eigen::Ref<Eigen::VectorXd> position,
                        Eigen::Ref<Eigen::VectorXd> velocity,
                        Eigen::Ref<Eigen::VectorXd> acceleration);
+
+    /**
+     * Evaluate the spline at a given point
+     * @param t instant time
+     * @param state of the system
+     * @return True in case of success, false otherwise.
+     */
+    bool evaluatePoint(const double& t,
+                       QuinticSplineState& state);
+
+    /**
+     * Get the state of the system.
+     * @warning if the the time step of the advance is not set the user cannot use the advance
+     * features.
+     * @return a const reference of the requested object.
+     */
+    const QuinticSplineState& get() const final;
+
+    /**
+     * Determines the validity of the object retrieved with get()
+     * @warning if the the time step of the advance is not set the user cannot use the advance
+     * features.
+     * @return True if the object is valid, false otherwise.
+     */
+    bool isValid() const final;
+
+    /**
+     * Advance the internal state. This may change the value retrievable from get().
+     * @warning if the the time step of the advance is not set the user cannot use the advance
+     * features.
+     * @return True if the advance is successfull.
+     */
+    bool advance() final;
+
 };
 } // namespace Planners
 } // namespace BipedalLocomotion
--- a/src/Planners/src/QuinticSpline.cpp
+++ b/src/Planners/src/QuinticSpline.cpp
@@ -169,7 +169,7 @@ QuinticSpline::QuinticSpline()
 QuinticSpline::~QuinticSpline() = default;
 
 bool QuinticSpline::setInitialConditions(Eigen::Ref<const Eigen::VectorXd> velocity,
-                                               Eigen::Ref<const Eigen::VectorXd> acceleration)
+                                         Eigen::Ref<const Eigen::VectorXd> acceleration)
 {
     if (velocity.size() != acceleration.size())
     {
@@ -290,7 +290,7 @@ bool QuinticSpline::Impl::computeCoefficients()
     this->polynomials.resize(knots.size() - 1);
 
     // set the velocity and acceleration boundary conditions
-    if(!this->setBoundaryVelocitiesAndAcceleration())
+    if (!this->setBoundaryVelocitiesAndAcceleration())
     {
         std::cerr << "[QuinticSpline::Impl::computeCoefficients] Unable to set the boundary "
                      "conditions related to the velocity and acceleration."
@@ -308,7 +308,7 @@ bool QuinticSpline::Impl::computeCoefficients()
     }
 
     // populate the polynomials vector with the knots
-    for(int i = 0; i < polynomials.size(); i++)
+    for (int i = 0; i < polynomials.size(); i++)
     {
         polynomials[i].initialPoint = &(knots[i]);
         polynomials[i].finalPoint = &(knots[i + 1]);
@@ -484,8 +484,8 @@ void QuinticSpline::Impl::addTripletNextKnot(const int& knotIndex,
 }
 
 void QuinticSpline::Impl::addKnownTermKnotPosition(const std::size_t& knotIndex,
-                                             const std::size_t& coordinateIndex,
-                                             Eigen::Ref<Eigen::VectorXd> b)
+                                                   const std::size_t& coordinateIndex,
+                                                   Eigen::Ref<Eigen::VectorXd> b)
 {
     const auto& poly = polynomials;
 
@@ -516,7 +516,7 @@ void QuinticSpline::Impl::addKnownTermNextKnot(const std::size_t& knotIndex,
 
     Eigen::Matrix2d tempMatrix;
     tempMatrix << -8 / std::pow(poly[i].duration, 2), 1 / poly[i].duration,
-                  14 / std::pow(poly[i].duration, 3), -2 / std::pow(poly[i].duration, 2);
+        14 / std::pow(poly[i].duration, 3), -2 / std::pow(poly[i].duration, 2);
 
     Eigen::Vector2d tempVector;
     tempVector << knots[i + 1].velocity[j], knots[i + 1].acceleration[j];
@@ -534,7 +534,7 @@ void QuinticSpline::Impl::addKnownTermPreviousKnot(const std::size_t& knotIndex,
 
     Eigen::Matrix2d tempMatrix;
     tempMatrix << 8 / std::pow(poly[i - 1].duration, 2), 1 / poly[i - 1].duration,
-                 14 / std::pow(poly[i - 1].duration, 3), 2 / std::pow(poly[i - 1].duration, 2);
+        14 / std::pow(poly[i - 1].duration, 3), 2 / std::pow(poly[i - 1].duration, 2);
 
     Eigen::Vector2d tempVector;
     tempVector << knots[i - 1].velocity[j], knots[i - 1].acceleration[j];
@@ -579,7 +579,7 @@ void QuinticSpline::Impl::computeIntermediateVelocitiesAndAcceleration()
     std::vector<Eigen::Triplet<double>> tripletsList;
 
     std::size_t numberOfExpectedTriplets = 4;
-    if(numberOfInteriorKnots  > 1)
+    if (numberOfInteriorKnots > 1)
     {
         numberOfExpectedTriplets = 2 * (6 * (numberOfInteriorKnots - 2) + 2 * 4);
     }
@@ -593,15 +593,11 @@ void QuinticSpline::Impl::computeIntermediateVelocitiesAndAcceleration()
         if (i == 0 && (i + 1) == numberOfInteriorKnots)
         {
             this->addTripletCurrentKnot(absoluteKnotIndex, 0, 0, tripletsList);
-        }
-        else if (i == 0)
+        } else if (i == 0)
         {
             this->addTripletCurrentKnot(absoluteKnotIndex, i * 2, 0, tripletsList);
             this->addTripletNextKnot(absoluteKnotIndex, i * 2, 2, tripletsList);
-
-        }
-
-        else if (i + 1 == numberOfInteriorKnots)
+        } else if (i + 1 == numberOfInteriorKnots)
         {
             this->addTripletPreviousKnot(absoluteKnotIndex, i * 2, 2 * (i - 1), tripletsList);
             this->addTripletCurrentKnot(absoluteKnotIndex, i * 2, 2 * (i - 1) + 2, tripletsList);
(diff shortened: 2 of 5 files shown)


prashanthr05:
Would you point me to a reference document I could review before going through the details of this PR?

S-Dafarra:
> Would you point me to a reference document I could review before going through the details of this PR?

It is probably already clear, but the reference issue is https://github.com/dic-iit/element_dcm-walking/issues/188

src/Planners/src/QuinticSpline.cpp
```diff
@@ -0,0 +1,787 @@
...
+    const auto& a3 = poly.a3;
+    const auto& a4 = poly.a4;
+    const auto& a5 = poly.a5;
+
+    position = a0
+             + a1 * t
+             + a2 * std::pow(t, 2)
+             + a3 * std::pow(t, 3)
+             + a4 * std::pow(t, 4)
+             + a5 * std::pow(t, 5);
```
S-Dafarra, line 693: The time is going to be elevated at the power of 5. While walking we usually don't reset the time, so this number can grow very big and cause numerical issues. Consider defining a time offset according to the first knot point.

prashanthr05:
> I could not check in detail the mathematical implementation, but I am going to trust the test sweat_smile

I can check the math details when I will do the review. It will be a good exercise for me as well.

src/Planners/src/QuinticSpline.cpp
```diff
@@ -0,0 +1,787 @@
...
+    // here we assume that at least 3 points has been defined
+    const std::size_t numberOfInteriorKnots = knots.size() - 2;
+
+    Eigen::SparseMatrix<double> A(2 * numberOfInteriorKnots, 2 * numberOfInteriorKnots);
+    std::vector<Eigen::Triplet<double>> tripletsList;
+
+    std::size_t numberOfExpectedTriplets = 4;
+    if(numberOfInteriorKnots  > 1)
+    {
+        numberOfExpectedTriplets = 2 * (6 * (numberOfInteriorKnots - 2) + 2 * 4);
```
prashanthr05, line 594: Could you add a line on how this comes up?

src/Planners/src/QuinticSpline.cpp
```diff
@@ -0,0 +1,787 @@
...
+        // in this case there is only one interior knot A will be a 2
+        if (i == 0 && (i + 1) == numberOfInteriorKnots)
+        {
+            this->addTripletCurrentKnot(absoluteKnotIndex, 0, 0, tripletsList);
+        }
+        else if (i == 0)
+        {
+            this->addTripletCurrentKnot(absoluteKnotIndex, i * 2, 0, tripletsList);
+            this->addTripletNextKnot(absoluteKnotIndex, i * 2, 2, tripletsList);
+
```
prashanthr05, line 611: this extra line could be removed.

src/Planners/src/QuinticSpline.cpp
```diff
@@ -0,0 +1,787 @@
...
+    // |    /   1      1\         / 1      1  \  |
+    // | 16 |------ + --|       3 |-- - ------|  |
+    // |    | 3        3|         | 2    2    |  |
+    // |    |T        T |         |T    T     |  |
+    // \    \ i - 1    i/         \ i    i - 1/  /
+
+    tripletList.push_back({rowOffset,
+                           columnOffset,
+                           12 * (1 / std::pow(poly[knotIndex - 1].duration, 2)
+                                 - 1 / std::pow(poly[knotIndex].duration, 2))});
```
prashanthr05, line 421:
```suggestion
    tripletList.push_back({rowOffset,
                           columnOffset,
                           12 * ( (1 / std::pow(poly[knotIndex - 1].duration, 2)
                                 - (1 / std::pow(poly[knotIndex].duration, 2)) ) });
```

I'd more careful and explicit with the use of braces here. Similarly for other triplet values.

prashanthr05: To be more readable, get a const ref on `poly[knotIndex - 1]` and `poly[knotIndex]` maybe?

src/Planners/src/QuinticSpline.cpp
```diff
@@ -0,0 +1,787 @@
...
+    // |   3            2   |
+    // |  T            T    |
+    // \   i            i   /
+
+    tripletList.push_back({rowOffset, columnOffset, -8 / std::pow(poly[knotIndex].duration, 2)});
+    tripletList.push_back({rowOffset, columnOffset + 1, 1 / poly[knotIndex].duration});
+    tripletList.push_back(
+        {rowOffset + 1, columnOffset, 14 / std::pow(poly[knotIndex].duration, 3)});
+    tripletList.push_back(
+        {rowOffset + 1, columnOffset + 1, -2 / std::pow(poly[knotIndex].duration, 2)});
```
prashanthr05, line 493: According to your derivation in https://github.com/dic-iit/element_dcm-walking/issues/188#issuecomment-673542645,  `-2 / std::pow(poly[knotIndex].duration, 2)`  must be raised to the power 3 `-2 / std::pow(poly[knotIndex].duration, 3)`. There seems to be a typo somewhere (most possibly in the issue comment). Please cross-verify.

src/Planners/src/QuinticSpline.cpp
```diff
@@ -0,0 +1,787 @@
...
+    const auto& poly = polynomials;
+
+    const auto& i = knotIndex;
+    const auto& j = coordinateIndex;
+
+    b[0] += 20
+            * (1 / std::pow(poly[i - 1].duration, 3)
+                   * (knots[i].position[j] - knots[i - 1].position[j])
+               + 1 / std::pow(poly[i].duration, 3)
+                     * (knots[i].position[j] - knots[i + 1].position[j]));
```
prashanthr05, line 509: Are you using a `+=` here assuming that the existing contents are zero?

src/Planners/include/BipedalLocomotion/Planners/QuinticSpline.h
```diff
@@ -0,0 +1,86 @@
...
+#include <Eigen/Dense>
+
+namespace BipedalLocomotion
+{
+namespace Planners
+{
+/**
+ * Quintic spline implement a 5-th order polynomial spline in \$f\mathbb{R}^n\$f.
+ */
+class QuinticSpline
```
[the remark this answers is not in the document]
GiulioRomualdi: Actually I don't know, I would like to have a similar interface to an [`iDynTree::CubicSpline`](https://robotology.github.io/idyntree/master/classiDynTree_1_1CubicSpline.html). Furthermore I was thinking to add also a function to get the entire trajectory (from the initial to the final time instant) this may help to use this class also with predictive controllers.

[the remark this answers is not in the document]
S-Dafarra: I don't think these things are mutually exclusive. Inheriting from ``Advanceable`` would imply having two additional methods, but you can still use it as you prefer.

src/Planners/src/QuinticSpline.cpp
GiulioRomualdi: Actually `getPositionAtTime` takes as input the current time minus the initial point instant. Let me know if I'm misunderstanding.


https://github.com/dic-iit/bipedal-locomotion-framework/blob/aaa044da923e6ed0dd895b5f293f36e24a5d39ab/src/Planners/src/QuinticSpline.cpp#L782-L784

src/Planners/src/QuinticSpline.cpp
```diff
@@ -0,0 +1,787 @@
...
+     * This vector is the known term used to compute the intermediate velocity and acceleration.
+     */
+    void addKnownTermPreviousKnot(const std::size_t& knotIndex,
+                                  const std::size_t& coordinateIndex,
+                                  Eigen::Ref<Eigen::VectorXd> b);
+
+    /**
+     * Compute the intermediate velocity and accelerations.
+     */
+    void computeIntermediateVelocitiesAndAccelaration();
```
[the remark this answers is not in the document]
GiulioRomualdi: Fixed in 316ad2fc96203dee9d7cc2e34a80e5ea4300b81d

src/Planners/src/QuinticSpline.cpp
```diff
@@ -0,0 +1,787 @@
...
+        // This is required or stability purposes, the matrix below may not be invertible
+        if (polynomials[i].duration == 0)
+        {
+            std::cerr << "[QuinticSpline::Impl::computePhasesDuration] Two consecutive points have "
+                         "the same time coordinate."
+                      << std::endl;
+            return false;
+        }
+
+        if (polynomials[i].duration == 0)
```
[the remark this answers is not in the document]
GiulioRomualdi: Thank you fixed in 316ad2fc96203dee9d7cc2e34a80e5ea4300b81d.

src/Planners/src/QuinticSpline.cpp
```diff
@@ -0,0 +1,787 @@
...
+    /**
+     * Compute the duration of each phase
+     */
+    bool computePhasesDuration();
+
+    /**
+     * Set tthe boundary condition. This function is called before computing all the intermediate
+     * velocities and accelerations.
+     */
+    bool setBoundaryVelocitiesAndAccelaration();
```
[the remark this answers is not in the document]
GiulioRomualdi: Fixed in 316ad2fc96203dee9d7cc2e34a80e5ea4300b81d

src/Planners/src/QuinticSpline.cpp
S-Dafarra: Ah great, I did not notice this was a method of ``Impl``. No issue then.

src/Planners/src/QuinticSpline.cpp
GiulioRomualdi: I improved the readability in 4b2c8c02d83ba54b7eb80957caa042321ba16ee2

src/Planners/include/BipedalLocomotion/Planners/QuinticSpline.h
```diff
@@ -0,0 +1,86 @@
...
+    bool setKnots(const std::vector<Eigen::VectorXd>& position, const std::vector<double>& time);
+
+    /**
+     * Set the initial condition of the spline
+     * @param velocity initial velocity (i.e. first derivative).
+     * @param acceleration initial acceleration (i.e. second derivative).
+     * @return True in case of success, false otherwise.
+     */
+    bool setInitialConditions(Eigen::Ref<const Eigen::VectorXd> velocity,
+                              Eigen::Ref<const Eigen::VectorXd> acceleration);
```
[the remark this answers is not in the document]
GiulioRomualdi: As written in [stackoverflow](https://stackoverflow.com/questions/21132538/correct-usage-of-the-eigenref-class)
> Then, there is not much difference between a `const Ref<const T>` and a `const Ref<const T>&` because it is unlikely that the function is called with an existing `Ref<const T>` object, and so a `Ref<const T>` will have to be created in most cases.

Since I don't call the functions on an existing `Ref` object. I choose:
-  `Eigen::Ref<const Eigen::VectorXd>` for readable only object
-  `Eigen::Ref<Eigen::VectorXd>` for writable object

I like the symmetry :smile: 

@traversaro @S-Dafarra what do you think about this?

src/Planners/src/QuinticSpline.cpp
```diff
@@ -0,0 +1,787 @@
...
+
+    /**
+     * addTripletCurrentKnot is an helper function to generate a triplet containing a 2x2 matrix.
```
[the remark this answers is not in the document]
GiulioRomualdi: Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/83/commits/121df8a8cef83ef427ef8d5da522f60ac2475c33

src/Planners/src/QuinticSpline.cpp
```diff
@@ -0,0 +1,787 @@
...
+    // check if the user already sets the knots
+    if (knots.empty())
+    {
+        std::cerr << "[QuinticSpline::Impl::computeCoefficients] Please set the knots before "
+                     "computing the coefficients."
+                  << std::endl;
+        return false;
+    }
+
+    // the number of polynomials is equal to the number of knots - -1
```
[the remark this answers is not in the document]
GiulioRomualdi: Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/83/commits/121df8a8cef83ef427ef8d5da522f60ac2475c33

src/Planners/src/QuinticSpline.cpp
```diff
@@ -0,0 +1,787 @@
...
+}
+
+bool QuinticSpline::Impl::computePhasesDuration()
+{
+    for (std::size_t i = 0; i < knots.size() - 1; i++)
+    {
+        polynomials[i].duration = knots[i + 1].timeInstant - knots[i].timeInstant;
+
+        // This is required or stability purposes, the matrix below may not be invertible
+        if (polynomials[i].duration == 0)
```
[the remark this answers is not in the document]
GiulioRomualdi: what do you think about this? 

```suggestion
        if (std::abs(polynomials[i].duration) <= std::numeric_limits<double>::epsilon())
```

src/Planners/include/BipedalLocomotion/Planners/QuinticSpline.h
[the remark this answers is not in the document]
S-Dafarra: I read ``Eigen::Ref`` as a fancy substitution for ``&``. In addition, it should be a very lightweight class with no dynamic memory allocation. I prefer passing them by copy, so I agree with your choice.

src/Planners/src/QuinticSpline.cpp
GiulioRomualdi: Well you are actually right the computation is a bit cryptic

given a generic internal point, the submatrix A is given by 
image 
this submatrix has 12 entries different from zero (i.e. 12 triplets)

for the first and the last points, the submatrix A contains only 8 entries. Indeed, in the case of the first internal point, the submatrix A does not contain the first two columns, and vice-versa in the case of the last internal point,  the submatrix A does not contain the last two columns.

On the light of this the `numberOfExpectedTriplets` is equal to
`numberOfExpectedTriplets = 12 * (numberOfInteriorKnots - 2) + 8 * 2`
I can update the code in this way

```suggestion
        constexpr int numberOfNonZeroExtremalPoints = 8;
        constexpr int numberOfNonZeroInternalPoints = 12;
        constexpr int extremalPoints = 2;
        numberOfExpectedTriplets = numberOfNonZeroInternalPoints * (numberOfInteriorKnots - extremalPoints)  + numberOfNonZeroExtremalPoints * extremalPoints;
```

src/Planners/src/QuinticSpline.cpp
GiulioRomualdi: Done in 97ebb0bfe8ea740c1318b2601865ae869e0ca41c

src/Planners/src/QuinticSpline.cpp
GiulioRomualdi: There was a typo in the issue comment. Fixed! Thanks for pointing out.

src/Planners/src/QuinticSpline.cpp
GiulioRomualdi: Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/83/commits/97ebb0bfe8ea740c1318b2601865ae869e0ca41c

src/Planners/src/QuinticSpline.cpp
GiulioRomualdi: Yep, `addKnownTermKnotPosition` is called only in this part of the code and here I reset to zero the vector every time 
https://github.com/dic-iit/bipedal-locomotion-framework/blob/97ebb0bfe8ea740c1318b2601865ae869e0ca41c/src/Planners/src/QuinticSpline.cpp#L618-L644

src/Planners/include/BipedalLocomotion/Planners/QuinticSpline.h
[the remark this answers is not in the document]
GiulioRomualdi: > I don't think these things are mutually exclusive. Inheriting from Advanceable would imply having two additional methods, but you can still use it as you prefer.

You are right, the only thing I don't like of advanceable in this case is the fact that you have to set the sampling time of the class (that could be not required) furthermore it is not really clear how the user should use it.
In other words:
```cpp
QuinticSpline sp;
// set the spline
sp.setSamplingTime(0.01)

auto traj = sp.get(); // <-------- Here I get the first value of the trajectory ( at t =  0 )
sp.advance(); // <----- Advace the trajectory
auto traj = sp.get(); // <-------- Here I get the second value of the trajectory ( at t = 0.01 )

sp.evaluatePoint(10, pos, vel, acc); // <-------- pos, vel and acc contains the state at t = 10

sp.advance(); // <----- Advace the trajectory
auto traj = sp.get(); // <-------- Here I get the second value of the trajectory ( at t = 0.02 )
```
I don't know if this may create problems in the future. If you think no I can add the possibility to set the sampling time. If it is set then the advance machinery works otherwise the user can get the trajectory only with `evaluatePoint()`.

src/Planners/src/QuinticSpline.cpp
```diff
@@ -0,0 +1,787 @@
...
+    }
+
+    m_pimpl->knots.resize(time.size());
+
+    const std::size_t sizeOfVectors = position[0].size();
+
+    m_pimpl->knots[0].timeInstant = time[0];
+    m_pimpl->knots[0].position = position[0];
+    m_pimpl->knots[0].velocity.resize(sizeOfVectors);
+    m_pimpl->knots[0].acceleration.resize(sizeOfVectors);
```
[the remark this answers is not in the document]
GiulioRomualdi: Done here: 53dc2b22795518d8e3f00ebe1a6ed4887fcc30ec

src/Planners/include/BipedalLocomotion/Planners/QuinticSpline.h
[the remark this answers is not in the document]
S-Dafarra: > You are right, the only thing I don't like of advanceable in this case is the fact that you have to set the sampling time of the class (that could be not required)

That is true. We can define a method like ``setAdvanceTimeStep``. In this way, it is clear that we are setting the time step for the ``advance`` method. We can discuss if this should be part of ``Advanceable`` or not. Regarding the example you showed, I don't think it is a big issue. ``Advanceable`` gives a basic interface for all the objects which output a trajectory. Then, if you have a reference to the actual class you can use its full-fledged interface.

src/Planners/src/QuinticSpline.cpp
prashanthr05: I think we can simply name them `numberOfLHSExtremumElements` and `numberOfLHSNonExtremumElements` which are 4 and 6 respectively, which gives,

```cpp
numberOfExpectedTriplets = 2*(numberOfLHSNonExtremumElements*(numberOfInteriorKnots - numberOfExtremum) + 2 * numberOfLHSExtremumElements )
```

What do you think?

src/Planners/src/QuinticSpline.cpp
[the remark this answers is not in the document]
prashanthr05: Yes. I think this should be fine.

src/Planners/src/QuinticSpline.cpp
prashanthr05: Understood.

src/Planners/include/BipedalLocomotion/Planners/QuinticSpline.h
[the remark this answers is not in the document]
GiulioRomualdi: https://github.com/dic-iit/bipedal-locomotion-framework/pull/83/commits/02557d7c8b4ccfef3bc6d147cb4aa00283110274 introduces the advance capabilities https://github.com/dic-iit/bipedal-locomotion-framework/pull/83/commits/5eb51cdf3cb1de6aed87548c5ce0224ed07b3612 is the test.

src/Planners/src/QuinticSpline.cpp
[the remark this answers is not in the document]
GiulioRomualdi: Done in 15982eaed361916a898f8884eabc138bd1b445ed

src/Planners/src/QuinticSpline.cpp
GiulioRomualdi: I don't know if `LHS` is a good acronym in this case. Is it related to the left-hand side of the aforementioned equation?

prashanthr05: > Is it related to the left-hand side of the aforementioned equation?

Yes.

src/Planners/src/QuinticSpline.cpp
```diff
@@ -0,0 +1,851 @@
...
+                                       return ((t >= p.initialPoint->timeInstant)
+                                               && (t < p.finalPoint->timeInstant));
+                                   });
+
+    if (poly == m_pimpl->polynomials.end())
+    {
+        std::cerr << "[QuinticSpline::evaluatePoint] Unable to find the sub-trajectory."
+                  << std::endl;
+        return false;
+    }
```
S-Dafarra, line 805: Notice that ``find_if`` performs a linear search, not a binary search. In any case, this is not a big deal with few knot points.

S-Dafarra (approved):

src/Planners/src/QuinticSpline.cpp
GiulioRomualdi: > > Is it related to the left-hand side of the aforementioned equation?
> 
> Yes.

Ok because in that part of the code it is not easy to understand what is the right-hand side of the equation

src/Planners/src/QuinticSpline.cpp
GiulioRomualdi: We should keep in mind this if we notice performance limitations.

src/Planners/src/QuinticSpline.cpp
prashanthr05: Then I think just the distinction between extremum and nonExtremum elements would be sufficient.

prashanthr05: Because mentioning them as,
```
constexpr int numberOfNonZeroExtremalPoints = 8;
constexpr int numberOfNonZeroInternalPoints = 12;
```

is a bit confusing as well.

GiulioRomualdi:
```cpp
constexpr int numberOfExtremumElement = 8;
constexpr int numberOfNonExtremumElement = 12;
```

what do you think?

prashanthr05: What I meant was, 
since the number of columns is 4 or 6 depending on if its an extremum or a non-extremum, we say 
```
constexpr int numberOfExtremumElements = 4;
constexpr int numberOfNonExtremumElements= 6;
```

and then we compute the `numberOfExpectedTriplets` corresponding to `x` and `xdot` as,

```
numberOfExpectedTriplets = 2*(numberOfNonExtremumElements*(numberOfInteriorKnots - numberOfExtremum) + 2 * numberOfExtremumElements )
```

Sorry for being picky here. Do you think this will be a more readable mapping of the expected matrix?

prashanthr05: Or do you think it's more meaningful to define them as in https://github.com/dic-iit/bipedal-locomotion-framework/pull/83#discussion_r482871832

GiulioRomualdi: > ```
> fNonExtremumElements*(numberOfInteriorKnots - numberOfExtremum) + 2 * numberOfExtremumElements )
> ```
> 
> 
> Sorry for being picky here. Do you think this will be a more readable m

This is fine. I'm working on this

S-Dafarra: I love the A :joy:

prashanthr05: Lovely documentation!!! Respect :rocket:

More love for the A :heart:

prashanthr05: Minor changes https://github.com/dic-iit/bipedal-locomotion-framework/commit/7d163f19c9ec8af0a3b413240f1bfa4605c66d3b#r42012518

prashanthr05 (approved):

prashanthr05:
I have given my approval for the PR.

However, please fix the typoes before merging, :)

PR #85, opened 2020-08-20
Implement Invariant EKF Floating Base Estimator

prashanthr05:
This PR,
- Implements `InvariantEKFBaseEstimator` (the **Contact Aided Invariant EKF algorithm**) by inheriting from the `FloatingBaseEstimator` class.  Reference to the original paper is [Contact-Aided Invariant Extended Kalman Filtering for Robot State Estimation](https://arxiv.org/abs/1904.09251). This implementation does not contain a full EKF SLAM algorithm using also visual factors. Instead, it's limited to estimating the foot positions of a bipedal robot only along with the base pose and IMU biases. For the original (and full) version, please see https://github.com/RossHartley/invariant-ekf.
- adds an `InvariantEKFBaseEstimatorTest` that tests the estimator algorithm for a simple static configuration of iCubGenova04 robot.

Additionally, this PR also
- adds https://github.com/artivis/manif as a dependency
- modifies the library structure so as to introduce a clean separation between regular `Estimators` algorithms and `FloatingBaseEstimators` algorithm
- modifies the `FloatingBaseEstimator` class slightly to add some required features with respect to the model computations.

Pending from the PR,
- [x] Documentation wherever required
- [x] implementation of resetEstimator() methods


prashanthr05:
Fails because CI was not updated to add `manif` as a dependency. 

To avoid any merge conflicts, maybe we can wait for this to be merged.
https://github.com/dic-iit/bipedal-locomotion-framework/pull/82

prashanthr05:
> To avoid any merge conflicts, maybe we can wait for this to be merged.

I added manif to CI in a way that merge conflicts will be avoided.

prashanthr05:
cc @S-Dafarra @GiulioRomualdi, this PR is ready to be reviewed. 
I apologize beforehand for a lengthy PR.

prashanthr05:
CI is failing due to YCM version. See https://github.com/robotology/whole-body-estimators/pull/83.

prashanthr05:
Comparison plots for a walking experiment from a Gazebo simulated dataset between a Matlab implementation, bipedal locomotion framework implementation and original implementation against the ground truth is below,

image

S-Dafarra:
Can you please try to rebase? In this way, the modification relative to the adding of ``manif`` should disappear.

prashanthr05:
> Can you please try to rebase? In this way, the modification relative to the adding of manif should disappear.

Done!! 

Please note that, this PR also contains a few changes in the parent class `FloatingBaseEstimator` that were projected upon due to the requirements in `InvariantEKFBaseEstimator`.  I also reorganized the headers to be separated from `Estimators` folder into a `FloatingBaseEstimators` folder.

prashanthr05:
The macOS CI is failing in the InvariantEKF Test, looks like the estimation results are bad (surprising). This needs to be investigated on a macOS machine. 

Instead of the windows, CI is failing due to linking the problem of YarpUtilitiesTest with `yarp sig Vector`  in the Build step.

GiulioRomualdi:
On Windows is actually failing because of 
```
2020-09-03T15:58:08.8415226Z D:\a\bipedal-locomotion-framework\bipedal-locomotion-framework\src\Estimators\tests\InvariantEKFBaseEstimatorTest.cpp(29,17): error C2065: 'M_PI': undeclared identifier
```
Here a related issue: https://github.com/robotology/QA/issues/81

This patch should solve your problem
```patch
diff --git a/cmake/AddBipedalLocomotionUnitTest.cmake b/cmake/AddBipedalLocomotionUnitTest.cmake
index 82430fd..4449f25 100644
--- a/cmake/AddBipedalLocomotionUnitTest.cmake
+++ b/cmake/AddBipedalLocomotionUnitTest.cmake
@@ -69,6 +69,7 @@ function(add_bipedal_test)
       target_link_libraries(${targetname} PRIVATE CatchTestMain ${${prefix}_LINKS})
       target_compile_definitions(${targetname} PRIVATE CATCH_CONFIG_FAST_COMPILE CATCH_CONFIG_DISABLE_MATCHERS)
       target_compile_features(${targetname} PUBLIC cxx_std_17)
+      target_compile_definitions(${targetname} PRIVATE -D_USE_MATH_DEFINES)
 
       add_test(NAME ${targetname} COMMAND ${targetname})
``` 

On my laptop, the test (in debug) is still failing. Normalizing the `simImuQuat` solves the problem.
```patch
diff --git a/src/Estimators/tests/InvariantEKFBaseEstimatorTest.cpp b/src/Estimators/tests/InvariantEKFBaseEstimatorTest.cpp
index 9b51ff0..c86684d 100644
--- a/src/Estimators/tests/InvariantEKFBaseEstimatorTest.cpp
+++ b/src/Estimators/tests/InvariantEKFBaseEstimatorTest.cpp
@@ -135,6 +135,7 @@ TEST_CASE("Invariant EKF Base Estimator")
     // ground truth
     Eigen::Vector3d simIMUPos;
     Eigen::Quaterniond simImuQuat = Eigen::Quaterniond(0.3218, -0.6304, -0.6292, 0.3212);
+    simImuQuat.normalize();
     simIMUPos << 0.0296, -0.1439,  0.4915;
 
     // IMU measures
@@ -222,4 +223,3 @@ TEST_CASE("Invariant EKF Base Estimator")
 
     REQUIRE( std::abs(newOut.stateStdDev.imuPosition(0) - resetStateStdDev.imuPosition(0)) < 2e-2);
 }
-
```

prashanthr05:
> On my laptop, the test (in debug) is still failing. Normalizing the simImuQuat solves the problem.

Yes, yet again I had failed to run the tests in Debug. I will keep this in mind. I will make relevant changes. Let's see if that fixes the macOS tests as well.

> This patch should solve your problem

I shall open a separate PR for that.

src/Estimators/CMakeLists.txt
```diff
@@ -3,19 +3,21 @@
...
+  set(H_PREFIX include/BipedalLocomotion/FloatingBaseEstimators)
   add_bipedal_locomotion_library(
     NAME                   FloatingBaseEstimators
-    SOURCES                src/FloatingBaseEstimator.cpp
-    PUBLIC_HEADERS         ${H_PREFIX}/FloatingBaseEstimatorParams.h ${H_PREFIX}/FloatingBaseEstimatorIO.h ${H_PREFIX}/FloatingBaseEstimator.h
-    PUBLIC_LINK_LIBRARIES  iDynTree::idyntree-high-level iDynTree::idyntree-core iDynTree::idyntree-model BipedalLocomotion::System Eigen3::Eigen)
-
+    SOURCES                src/FloatingBaseEstimator.cpp src/InvariantEKFBaseEstimator.cpp
+    PUBLIC_HEADERS         ${H_PREFIX}/FloatingBaseEstimatorParams.h ${H_PREFIX}/FloatingBaseEstimatorIO.h ${H_PREFIX}/FloatingBaseEstimator.h ${H_PREFIX}/InvariantEKFBaseEstimator.h
+    PUBLIC_LINK_LIBRARIES  iDynTree::idyntree-high-level iDynTree::idyntree-core iDynTree::idyntree-model BipedalLocomotion::System Eigen3::Eigen
```
S-Dafarra, line 21: ``BipedalLocomotion::ParametersHandler`` seems missing

src/Estimators/include/BipedalLocomotion/FloatingBaseEstimators/FloatingBaseEstimator.h
```diff
@@ -83,6 +83,21 @@ class FloatingBaseEstimator : public BipedalLocomotion::System::Advanceable<Floa
...
+        * @param[out] IMU_H_r_foot pose of the right foot contact frame with respect to the IMU frame
+        * @param[out] J_IMULF explicit Jacobian of left foot frame with respect to IMU frame
+        * @param[out] J_IMURF explicit Jacobian of right foot frame with respect to IMU frame
+        * @return True in case of success, false otherwise.
+        */
+        bool getIMU_H_feet(const iDynTree::JointPosDoubleArray& encoders,
+                           iDynTree::Transform& IMU_H_l_foot,
+                           iDynTree::Transform& IMU_H_r_foot,
+                           iDynTree::MatrixDynSize& J_IMULF,
+                           iDynTree::MatrixDynSize& J_IMURF);
```
S-Dafarra, line 99: Aren't we supposed to use ``Eigen`` and ``manif`` objects in the main headers?

S-Dafarra (commented):
Some initial comments

src/Estimators/include/BipedalLocomotion/FloatingBaseEstimators/FloatingBaseEstimator.h
prashanthr05: Technically, yes. But since this ModelComputations class is specific to the usage of `iDynTree Model`, I stuck to the use of iDynTree objects within this class. 

It was a design decision. Do you think it should rather use Eigen or `manif` objects?

src/Estimators/include/BipedalLocomotion/FloatingBaseEstimators/FloatingBaseEstimator.h
```diff
@@ -346,6 +381,18 @@ class FloatingBaseEstimator : public BipedalLocomotion::System::Advanceable<Floa
...
+     * Wrapper method for getting base state from internal IMU state
+     * @param[in] state internal state of the estimator
+     * @param[in] meas previous measurement to the estimator
+     * @param[in] basePose base pose as an iDynTree Transform object
+     * @param[in] baseTwist mixe- trivialized base velocity as an iDynTree Twist object
+     */
+    bool updateBaseStateFromIMUState(const FloatingBaseEstimators::InternalState& state,
+                                     const FloatingBaseEstimators::Measurements& meas,
+                                     iDynTree::Transform& basePose,
+                                     iDynTree::Twist& baseTwist);
```
prashanthr05, line 395: This should be changed to `manif` and `Eigen` objects.

prashanthr05, line 395: This should be changed to `manif` and `Eigen` objects.

src/Estimators/include/BipedalLocomotion/FloatingBaseEstimators/FloatingBaseEstimator.h
S-Dafarra: If it is a public method, then probably it is better to stick to ``Eigen``/``manif`` for consistency. If instead this is an implementation specific to iDynTree, it may make sense to remain with iDynTree.

S-Dafarra: Also, if this complicates a lot some of your experiments, we can also postpone this change.

prashanthr05: I will open an issue regarding this. So that these changes might be in place within a newer PR.

prashanthr05:
Ciao @kouroshD !! May I ask you a favor of testing this PR by running the `InvariantEKFBaseEstimatorTest` after compiling on `Debug` build on your computer?

prashanthr05:
> Ciao @kouroshD !! May I ask you a favor of testing this PR by running the `InvariantEKFBaseEstimatorTest` after compiling on `Debug` build on your computer?
> 
> You might have to install the dependency "manif" on your computer for testing this PR.

Ciao @kouroshD .. do not worry about this. We managed to procure a Mac workstation in the lab. I shall do the necessary testing myself :)

prashanthr05:
STrangely, running the `InvariantEKFBaseEstimatorTest` on macOS Catalina by compiling the repository in Debug, the test passes successfully. But with Release, it fails.

I am installing valgrind according to https://stackoverflow.com/questions/58360093/how-to-install-valgrind-on-macos-catalina-10-15-with-homebrew  by tapping into `LouisBrunner/valgrind`, since I was not able to directly install it from brew, which threw an error that it is not supported or function as expected in Catalina.

prashanthr05:
Looks like using auto on a `manif` to `Eigen` conversion was causing uninitialized value condition. This was caught by valgrind on macOS Catalina.

I changed this in https://github.com/dic-iit/bipedal-locomotion-framework/pull/85/commits/3020229aa2508e03de7c960e789bf53bed6b4e93 and the tests seem to pass both on Debug and Release.

S-Dafarra:
> Looks like using `auto` on a `manif` to `Eigen` conversion was causing uninitialized value condition. This was caught by valgrind on macOS Catalina local workstation.
> 
> I changed this in [3020229](https://github.com/dic-iit/bipedal-locomotion-framework/commit/3020229aa2508e03de7c960e789bf53bed6b4e93) and the tests seem to pass both on Debug and Release on the local workstation.

Take home message, never use ``auto`` with Eigen types https://eigen.tuxfamily.org/dox/TopicPitfalls.html :grin:

README.md
```diff
@@ -35,22 +35,27 @@ file. Please note that the indicated version is the the minimum required version
...
-- `Planner` requires:
+- `FloatingBaseEstimators` require:
+    - For using it:
+      - [`iDynTree`](https://github.com/robotology/idyntree) (version 0.11.105)
+      - [`ParametersHandler`](./src/ParametersHandler)
+      - [`manif`](https://github.com/artivis/manif)
+    - For testing:
+      - [`Catch2`](https://github.com/catchorg/Catch2)
+      - [`YARP`](https://github.com/robotology/YARP)
+- Planner` requires:
```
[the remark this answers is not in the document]
prashanthr05: Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/85/commits/b5deac5a01d8952050adae2e968e71ddb24de463

prashanthr05:
> > Looks like using `auto` on a `manif` to `Eigen` conversion was causing uninitialized value condition. This was caught by valgrind on macOS Catalina local workstation.
> > I changed this in [3020229](https://github.com/dic-iit/bipedal-locomotion-framework/commit/3020229aa2508e03de7c960e789bf53bed6b4e93) and the tests seem to pass both on Debug and Release on the local workstation.
> 
> Take home message, never use `auto` with Eigen types https://eigen.tuxfamily.org/dox/TopicPitfalls.html

Changed `auto` on Eigen types to relevant types in https://github.com/dic-iit/bipedal-locomotion-framework/pull/85/commits/b5deac5a01d8952050adae2e968e71ddb24de463

src/Estimators/CMakeLists.txt
prashanthr05: Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/85/commits/434835236824d749ac2d60e93ceb6eefbc3d96e9

src/Estimators/include/BipedalLocomotion/FloatingBaseEstimators/FloatingBaseEstimator.h
```diff
@@ -172,6 +187,26 @@ class FloatingBaseEstimator : public BipedalLocomotion::System::Advanceable<Floa
...
+    /**
+     * Reset the base pose estimate and consequently the internal state of the estimator
+     * @param[in] newBaseOrientation base link orientation as a Eigen quaternion
+     * @param[in] newBasePosition base link position
+     * @return True in case of success, false otherwise.
+     *
+     * * @note reset and advance estimator to get updated estimator output
+     */
+    virtual bool resetEstimator(const Eigen::Quaterniond& newBaseOrientation,
+                                const Eigen::Vector3d& newBasePosition) final;
```
GiulioRomualdi, line 208: here you can use:
```suggestion
    virtual bool resetEstimator(Eigen::Ref<const Eigen::Quaterniond> newBaseOrientation,
                                Eigen::Ref<const Eigen::Vector3d> newBasePosition) final;
```

src/Estimators/src/FloatingBaseEstimator.cpp
```diff
@@ -604,3 +637,73 @@ bool FloatingBaseEstimator::setupFixedVectorParamPrivate(const std::string& para
...
+    iDynTree::Twist v_IMU;
+    iDynTree::Vector3 imuLinVel, imuAngVel;
+    iDynTree::toEigen(imuLinVel) = state.imuLinearVelocity;
+    iDynTree::toEigen(imuAngVel) = state.imuOrientation.toRotationMatrix()*(meas.gyro - state.gyroscopeBias);
+    v_IMU.setLinearVec3(imuLinVel);
+    v_IMU.setAngularVec3(imuAngVel);
+
+    if (!m_modelComp.getBaseStateFromIMUState(A_H_IMU, v_IMU, basePose, baseTwist))
+    {
+        return false;
```
GiulioRomualdi, line 660: Probably you can print something here. It may help you while debugging

src/Estimators/src/FloatingBaseEstimator.cpp
```diff
@@ -604,3 +637,73 @@ bool FloatingBaseEstimator::setupFixedVectorParamPrivate(const std::string& para
...
+    }
+
+    auto A_H_IMU = A_H_B*m_modelComp.base_H_IMU();;
+    auto A_H_RF = A_H_IMU*IMU_H_RF;
+    auto A_H_LF = A_H_IMU*IMU_H_LF;
+
+    // convert to iDynTree Rotation to  Eigen angle axis and then to quaternion for consistent conversion
+    auto imuAngleAxis = Eigen::AngleAxisd(iDynTree::toEigen(A_H_IMU.getRotation()));
+    auto rfAngleAxis = Eigen::AngleAxisd(iDynTree::toEigen(A_H_RF.getRotation()));
+    auto lfAngleAxis = Eigen::AngleAxisd(iDynTree::toEigen(A_H_LF.getRotation()));
```
GiulioRomualdi, line 692: I would avoid using `auto` with `Eigen`

src/Estimators/src/InvariantEKFBaseEstimator.cpp
```diff
@@ -0,0 +1,918 @@
...
+    m_pimpl->calcFc(m_statePrev, m_pimpl->m_skew, m_options.imuBiasEstimationEnabled, Fc); // compute Fc at priori state
+    m_pimpl->calcQc(m_statePrev, m_sensorsDev, meas,  m_options.imuBiasEstimationEnabled, Qc); // compute Qc at priori state and previous measure
+    m_pimpl->calcLc(m_statePrev, m_pimpl->m_skew, m_options.imuBiasEstimationEnabled, Lc); // compute Lc at priori state
+
+    // discretize linearized dynamics and propagate covariance
+    size_t dims = Fc.rows();
+    Eigen::MatrixXd I = Eigen::MatrixXd::Identity(dims, dims);
+
+    Eigen::MatrixXd Fk = I + (Fc*dt);
+    Eigen::MatrixXd Qk = (Fk*Lc*Qc*(Lc.transpose())*(Fk.transpose()))*dt;
```
GiulioRomualdi, line 326: Do you think we may avoid continuous memory allocation if `I` ,`Fk` and `Qk` are class members?

src/Estimators/src/InvariantEKFBaseEstimator.cpp
```diff
@@ -0,0 +1,918 @@
...
+        {
+            measModelJacobian.block<6, 6>(0, m_pimpl->m_vecOffsets.gyroBias).setZero();
+        }
+
+        // prepare auxiliary gain matrix Pi
+        auxMat.resize(6, 14);
+        auxMat.block<3, 3>(0, 0) = auxMat.block<3, 3>(3, 7) = Eigen::Matrix3d::Identity();
+        auxMat.block<3, 11>(0, 3).setZero();
+        auxMat.block<3, 7>(3, 0).setZero();
+        auxMat.block<3, 4>(3, 10).setZero();
```
GiulioRomualdi, line 407: Super error-prone ;) (by the way it's fine)

src/Estimators/src/InvariantEKFBaseEstimator.cpp
```diff
@@ -0,0 +1,918 @@
...
+        estimateBias = true;
+    }
+    else
+    {
+        estimateBias = false;
+    }
+
+    Eigen::MatrixXd PHT = P*measModelJacobian.transpose();
+    Eigen::MatrixXd S = measModelJacobian*PHT + measNoiseVar;
+    Eigen::MatrixXd K = PHT*(S.inverse());
```
GiulioRomualdi, line 570: As before, probably we may avoid having this continuous memory allocation.

src/Estimators/src/InvariantEKFBaseEstimator.cpp
```diff
@@ -0,0 +1,918 @@
...
+    //        |         1   |
+    //        |            1|
+    X.resize(7, 7);
+    X.block<3, 3>(0, 0) = state.imuOrientation.toRotationMatrix();
+    X.block<3, 1>(0, 3) = state.imuLinearVelocity;
+    X.block<3, 1>(0, 4) = state.imuPosition;
+    X.block<3, 1>(0, 5) = state.rContactFramePosition;
+    X.block<3, 1>(0, 6) = state.lContactFramePosition;
+    X.bottomLeftCorner<4, 3>().setZero();
+    X.bottomRightCorner<4, 4>() = Eigen::Matrix4d::Identity();
```
GiulioRomualdi, line 682: Probably it makes sense to store X as sparse matrix

src/Estimators/include/BipedalLocomotion/FloatingBaseEstimators/FloatingBaseEstimator.h
prashanthr05: `Eigen::Ref` doesn't seem to work on Quaternions.

src/Estimators/src/InvariantEKFBaseEstimator.cpp
prashanthr05: Error-prone because of the magic numbers? or do you see some other issue? I will remove the magic numbers by proper variable assignments.

src/Estimators/src/FloatingBaseEstimator.cpp
prashanthr05: Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/85/commits/415e1729f4d047cfa520ce00633f7ede70796167

src/Estimators/src/FloatingBaseEstimator.cpp
prashanthr05: Done in https://github.com/dic-iit/bipedal-locomotion-framework/commit/415e1729f4d047cfa520ce00633f7ede70796167

src/Estimators/src/InvariantEKFBaseEstimator.cpp
prashanthr05: Magic numbers removed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/85/commits/97ce5cab64ffad935b3f2329824f4cd0d9fc482c

src/Estimators/src/InvariantEKFBaseEstimator.cpp
prashanthr05: Actually, here the catch is that these matrices do not have a fixed dimension throughout the run time. The dimensions vary if bias estimation is toggled during run time. 

In this case, storing the matrices as class members means that we need to modify the mechanics of current implementation in a way that these matrix dimensions are updated at every step. It is not difficult to implement, however, in any case, we need to be reallocating the memory during the resize. That's why I chose this way instead.

Correct me if my rationale is not good enough.

src/Estimators/src/InvariantEKFBaseEstimator.cpp
prashanthr05: Given it's only a 7x7 matrix, do you think we can achieve higher gain margins using a sparse matrix?

src/Estimators/include/BipedalLocomotion/FloatingBaseEstimators/FloatingBaseEstimator.h
prashanthr05: I opened an issue for this. https://github.com/dic-iit/bipedal-locomotion-framework/issues/107.

This will be dealt with later in a newer PR.

src/Estimators/src/InvariantEKFBaseEstimator.cpp
S-Dafarra: In our use cases, sparse matrices are useful only if need in some calculations involving inversions (e.g solving linear systems or when interfacing to solvers). Another possible application is to reduce the occupation in memory, but this is not an issue for basically all our applications. Hence, if we are not in the first case, I would avoid using sparse matrices.

src/Estimators/src/InvariantEKFBaseEstimator.cpp
S-Dafarra: I think that some resizing from time to time are better than allocating memory all the time. Can you provide some more insights on why a continuous resize may be necessary?

prashanthr05: Continuous resize won't be necessary. Just at events of bias estimation toggle. I can make the relevant changes.

Here, we do not even consider arbitrary number of contacts (so we don't need to continuously resize the matrices.)

src/Estimators/src/InvariantEKFBaseEstimator.cpp
prashanthr05: The most we do with this matrix is matrix multiplication. For inversions, we exploit with Lie group properties of direct product/semi-direct-product to only involve submatrix operations.

S-Dafarra: Ok, I would vote for keeping it dense then.

src/Estimators/src/InvariantEKFBaseEstimator.cpp
prashanthr05: I'd prefer to keep this part as it is since this also depends on the number of feet in contact with the ground which seems to be changing "almost continuously". This means anyways we might have to do an almost continuous resizing of these matrices. Instead for the above comment, I made the matrices `Fc`, `Qc` and `Lc` as class members to avoid continuous memory allocation, since their dimensions would only change at very rare events of toggling the bias estimation.

What do you think @GiulioRomualdi?

S-Dafarra: To be honest, I am with @GiulioRomualdi on this. It is always better to avoid memory allocations, especially in pieces of code that are called at every control iteration. In principle, these should have no memory allocation at all. 
It is acceptable to have some resize from time to time, also considering that the container may have some capacity (hence memory may not be allocated also in case of resizes). 
If the number of members blows up, then we can discuss alternative solutions.

Also, I have noticed that in the code below there are several temporary matrices (``X``, ``BigX``, ``innovation``, ``delta``, ``dX``, ``IminusKH``). Also this should be avoided as much as possible.

src/Estimators/src/InvariantEKFBaseEstimator.cpp
prashanthr05: Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/85/commits/6abd67857f238dc60818b25e497be228b62ab230

src/Estimators/src/InvariantEKFBaseEstimator.cpp
prashanthr05: I've partially done this in https://github.com/dic-iit/bipedal-locomotion-framework/pull/85/commits/6abd67857f238dc60818b25e497be228b62ab230. Let me know if this is in line with what is suggested. However, about the resizing of the matrices related to the measurement model/updates, it would require a bit more time to do it in a cleaner way.


I also completely agree with what is suggested. In fact, I had an earlier design in the element repo which considered these things. But not sure why I changed my mind to do things in a different way here.

S-Dafarra: ``Eigen`` automatically resizes the matrix when using the assignment operator. So you don't need to deal with the explicit resize for a first iteration I think. Then, later on, we can understand if it should be better to resize these matrices beforehand in a specific place.

prashanthr05: Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/85/commits/d311aa2da5cc8750123dce7076d0c041f93e3e08

S-Dafarra (approved):
The PR is very complex for me to delve into every detail, it is above my head :grin: But on the software side, I did not spot anything worth changing. In addition, the implementation can change easily later if needed. Hence, it looks good to me

src/Estimators/include/BipedalLocomotion/FloatingBaseEstimators/FloatingBaseEstimator.h
```diff
@@ -83,6 +83,21 @@ class FloatingBaseEstimator : public BipedalLocomotion::System::Advanceable<Floa
                            iDynTree::Transform& IMU_H_l_foot,
                            iDynTree::Transform& IMU_H_r_foot);
 
+        /**
+        * Get relative pose between IMU and the feet
+        * @param[in] encoders joint positions through encoder measurements
+        * @param[out] IMU_H_l_foot pose of the left foot contact frame with respect to the IMU frame
+        * @param[out] IMU_H_r_foot pose of the right foot contact frame with respect to the IMU frame
+        * @param[out] J_IMULF explicit Jacobian of left foot frame with respect to IMU frame
```
traversaro, line 91: What do you mean by "explicit" jacobian?

traversaro (approved):
I did not checked everithing in detail, but overall it seems fine!

src/Estimators/include/BipedalLocomotion/FloatingBaseEstimators/FloatingBaseEstimator.h
prashanthr05: I suppose I copied the actual KinDynComputation's method name where we have to "explicitly" specify the frame indices.
Usually, it's obtained wrt the base link, instead here, we have here wrt the IMU.

prashanthr05: Should I remove that?

traversaro: As you prefer, personally I don't think that the explicit meaning is clear in this context, so to avoid confusion it is better to remove it, but it is not blocking the merge of the PR for me.

prashanthr05: You are right! I updated the documentation https://github.com/dic-iit/bipedal-locomotion-framework/pull/85/commits/c5b51a1695e10c955ea54019e2d09c3ddb552274.

prashanthr05:
cc @GiulioRomualdi do you think we can merge this?

GiulioRomualdi (approved):

prashanthr05:
@GiulioRomualdi  I've squashed the commits. You could merge it once the CI passes, thank you! :)

GiulioRomualdi:
Merging

PR #87, opened 2020-08-27
Add RobotInterface library and a SensorBridge interface class

prashanthr05:
This PR,
- Adds `RobotInterface` library to the Framework
- Adds `ISensorBridge` interface class to `RobotInterface` library
~~- Adds OpenCV as a dependency for handling `cv::Mat` within the `ISensorBridge` class~~
- Adds a trivial `DummySensorBridge` test implementation to depict the use of the `ISensorBridge` interface

### DIFF
--- a/src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
+++ b/src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
@@ -1,0 +1,372 @@
+/**
+ * @file ISensorBridge.h
+ * @authors Prashanth Ramadoss
+ * @copyright 2020 Istituto Italiano di Tecnologia (IIT). This software may be modified and
+ * distributed under the terms of the GNU Lesser General Public License v2.1 or any later version.
+ */
+
+#ifndef BIPEDAL_LOCOMOTION_ROBOT_INTERFACE_ISENSOR_BRIDGE_H
+#define BIPEDAL_LOCOMOTION_ROBOT_INTERFACE_ISENSOR_BRIDGE_H
+
+#include <memory>
+#include <string>
+#include <vector>
+#include <unordered_map>
+
+#include <Eigen/Dense>
+
+#include <BipedalLocomotion/ParametersHandler/IParametersHandler.h>
+
+using namespace BipedalLocomotion::ParametersHandler;
+using Vector12d = Eigen::Matrix<double, 12, 1>;
+using Vector6d = Eigen::Matrix<double, 6, 1>;
+
+namespace BipedalLocomotion
+{
+namespace RobotInterface
+{
+
+/**
+ * Sensor bridge options
+ */
+struct SensorBridgeOptions
+{
+    bool isKinematicsEnabled{false}; /**< flag to connect kinematics measurement sources */
+    bool isIMUEnabled{false}; /**< flag to connect IMU measurement sources */
+    bool isLinearAccelerometerEnabled{false}; /**< flag to connect linear accelerometer measurement sources */
+    bool isGyroscopeEnabled{false}; /**< flag to connect gyroscope measurement sources */
+    bool isMagnetometerEnabled{false}; /**< flag to connect magnetometer measurement sources */
+    bool isSixAxisForceTorqueSensorEnabled{false}; /**< flag to connect six axis force torque measurement sources */
+    bool isThreeAxisForceTorqueSensorEnabled{false}; /**< flag to connect six axis force torque measurement sources */
+    bool isCartesianWrenchEnabled{false}; /**< flag to connect cartesian wrench measurement sources */
+    bool isCameraEnabled{false}; /**< flag to connect camera sources */
+
+    size_t nrJoints{0}; /**< number of joints available through Kinematics stream, to be configured at initialization */
+    std::unordered_map<std::string, std::pair<int, int> > imgDimensions; /**< dimensions of the camera images available through camera streams, to be configured at initialization */
+};
+
+/**
+ *  Sensor lists
+ */
+struct SensorLists
+{
+    std::vector<std::string> jointsList; /**< list of joints attached to the bridge */
+    std::vector<std::string> IMUsList;   /**< list of IMUs attached to the bridge */
+    std::vector<std::string> linearAccelerometersList; /**< list of linear accelerometers attached to the bridge */
+    std::vector<std::string> gyroscopesList; /**< list of gyroscopes attached to the bridge */
+    std::vector<std::string> orientationSensorsList; /**< list of orientation sensors attached to the bridge */
+    std::vector<std::string> magnetometersList; /**< list of magnetometers attached to the bridge */
+    std::vector<std::string> sixAxisForceTorqueSensorsList; /**< list of six axis force torque sensors attached to the bridge */
+    std::vector<std::string> threeAxisForceTorqueSensorsList; /**< list of three axis force torque sensors attached to the bridge */
+    std::vector<std::string> cartesianWrenchesList; /**< list of cartesian wrench streams attached to the bridge */
+    std::vector<std::string> rgbCamerasList; /**< list of rgb cameras attached to the bridge */
+    std::vector<std::string> depthCamerasList; /**< list of depth cameras attached to the bridge */
+};
+
+/**
+ * Sensor bridge interface.
+ */
+class ISensorBridge
+{
+public:
+    using unique_ptr = std::unique_ptr<ISensorBridge>;
+
+    using shared_ptr = std::shared_ptr<ISensorBridge>;
+
+    using weak_ptr = std::weak_ptr<ISensorBridge>;
+
+    /**
+     * Initialize estimator
+     * @param[in] handler Parameters handler
+     */
+    virtual bool initialize(std::weak_ptr<IParametersHandler> handler) = 0;
+
+    /**
+     * Get joints list
+     * @param[out] jointsList list of joints attached to the bridge
+     * @return  true/false in case of success/failure
+     */
+    virtual bool getJointsList(std::vector<std::string>& jointsList) = 0;
+
+    /**
+     * Get imu sensors
+     * @param[out] IMUsList list of IMUs attached to the bridge
+     * @return  true/false in case of success/failure
+     */
+    virtual bool getIMUsList(std::vector<std::string>& IMUsList) = 0;
+
+    /**
+     * Get linear accelerometers
+     * @param[out] linearAccelerometersList list of linear accelerometers attached to the bridge
+     * @return  true/false in case of success/failure
+     */
+    virtual bool getLinearAccelerometersList(std::vector<std::string>& linearAccelerometersList) = 0;
+
+    /**
+     * Get gyroscopes
+     * @param[out] gyroscopesList list of gyroscopes attached to the bridge
+     * @return  true/false in case of success/failure
+     */
+    virtual bool getGyroscopesList(std::vector<std::string>& gyroscopesList) = 0;
+
+    /**
+     * Get orientation sensors
+     * @param[out] orientationSensorsList list of orientation sensors attached to the bridge
+     * @return  true/false in case of success/failure
+     */
+    virtual bool getOrientationSensorsList(std::vector<std::string>& orientationSensorsList) = 0;
+
+    /**
+     * Get magnetometers sensors
+     * @param[out] magnetometersList list of magnetometers attached to the bridge
+     * @return  true/false in case of success/failure
+     */
+    virtual bool getMagnetometersList(std::vector<std::string>& magnetometersList) = 0;
+
+    /**
+     * Get 6 axis FT sensors
+     * @param[out] sixAxisForceTorqueSensorsList list of 6 axis force torque sensors attached to the bridge
+     * @return  true/false in case of success/failure
+     */
+    virtual bool getSixAxisForceTorqueSensorsList(std::vector<std::string>& sixAxisForceTorqueSensorsList) = 0;
+
+    /**
+     * Get 6 axis FT sensors
+     * @param[out] threeAxisForceTorqueSensorsList list of 3 axis force torque sensors attached to the bridge
+     * @return  true/false in case of success/failure
+     */
+    virtual bool getThreeAxisForceTorqueSensorsList(std::vector<std::string>& threeAxisForceTorqueSensorsList) = 0;
+
+   /**
+     * Get cartesian wrenches
+     * @param[out] cartesianWrenchesList list of cartesian wrenches attached to the bridge
+     * @return  true/false in case of success/failure
+     */
+    virtual bool getCartesianWrenchesList(std::vector<std::string>& cartesianWrenchesList) = 0;
+
+    /**
+     * Get rgb cameras
+     * @param[out] rgbCamerasList list of rgb cameras attached to the bridge
+     * @return  true/false in case of success/failure
+     */
+    virtual bool getRGBCamerasList(std::vector<std::string>& rgbCamerasList) = 0;
+
+    /**
+     * Get depth cameras
+     * @param[out] depthCamerasList list of depth cameras attached to the bridge
+     * @return  true/false in case of success/failure
+     */
+    virtual bool getDepthCamerasList(std::vector<std::string>& depthCamerasList) = 0;
+
+    /**
+     * Get joint position  in radians
+     * @param[in] jointName name of the joint
+     * @param[out] jointPosition joint position in radians
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     * @return true/false in case of success/failure
+     */
+    virtual bool getJointPosition(const std::string& jointName,
+                                  double& jointPosition,
+                                  double* receiveTimeInSeconds = nullptr) = 0;
+
+    /**
+     * Get all joints' positions in radians
+     * @param[out] parameter all joints' position in radians
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     *
+     * @warning the size is decided at the configuration and remains fixed,
+     * and internal checks must be done at the implementation level by the Derived class.
+     * This means that the user must pass a resized argument "jointPositions" to this method
+     *
+     * @return true/false in case of success/failure
+     */
+    virtual bool getJointPositions(Eigen::Ref<Eigen::VectorXd> jointPositions,
+                                   double* receiveTimeInSeconds = nullptr) = 0;
+
+    /**
+     * Get joint velocity in rad/s
+     * @param[in] jointName name of the joint
+     * @param[out] jointVelocity joint velocity in radians per second
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     * @return true/false in case of success/failure
+     */
+    virtual bool getJointVelocity(const std::string& jointName,
+                                  double& jointVelocity,
+                                  double* receiveTimeInSeconds = nullptr) = 0;
+
+    /**
+     * Get all joints' velocities in rad/s
+     * @param[out] parameter all joints' velocities in radians per second
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     *
+     * @warning the size is decided at the configuration and remains fixed,
+     * and internal checks must be done at the implementation level by the Derived class.
+     * This means that the user must pass a resized argument "jointVelocties" to this method
+     *
+     * @return true/false in case of success/failure
+     */
+    virtual bool getJointVelocities(Eigen::Ref<Eigen::VectorXd> jointVelocties,
+                                    double* receiveTimeInSeconds = nullptr) = 0;
+
+    /**
+     * Get IMU measurement
+     * The serialization of measurments is as follows,
+     *   (rpy acc omega mag)
+     *  - rpy in radians Roll-Pitch-Yaw Euler angles
+     *  - acc in m/s^2 linear accelerometer measurements
+     *  - omega in rad/s gyroscope measurements
+     *  - mag in tesla magnetometer measurements
+     * @param[in] imuName name of the IMU
+     * @param[out] imuMeasurement imu measurement of size 12
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     * @return true/false in case of success/failure
+     */
+    virtual bool getIMUMeasurement(const std::string& imuName,
+                                   Eigen::Ref<Vector12d> imuMeasurement,
+                                   double* receiveTimeInSeconds = nullptr) = 0;
+
+    /**
+     * Get linear accelerometer measurement in m/s^2
+     * @param[in] accName name of the linear accelerometer
+     * @param[out] accMeasurement linear accelerometer measurements of size 3
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     * @return true/false in case of success/failure
+     */
+    virtual bool getLinearAccelerometerMeasurement(const std::string& accName,
+                                                   Eigen::Ref<Eigen::Vector3d> accMeasurement,
+                                                   double* receiveTimeInSeconds = nullptr) = 0;
+
+    /**
+     * Get gyroscope measurement in rad/s
+     * @param[in] gyroName name of the gyroscope
+     * @param[out] gyroMeasurement gyroscope measurements of size 3
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     * @return true/false in case of success/failure
+     */
+    virtual bool getGyroscopeMeasure(const std::string& gyroName,
+                                     Eigen::Ref<Eigen::Vector3d> gyroMeasurement,
+                                     double* receiveTimeInSeconds = nullptr) = 0;
+
+   /**
+     * Get orientation sensor measurement in radians as roll pitch yaw Euler angles
+     * @param[in] rpyName name of the orientation sensor
+     * @param[out] rpyMeasurement rpy measurements of size 3
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     * @return true/false in case of success/failure
+     */
+    virtual bool getOrientationSensorMeasurement(const std::string& rpyName,
+                                                 Eigen::Ref<Eigen::Vector3d> rpyMeasurement,
+                                                 double* receiveTimeInSeconds = nullptr) = 0;
+
+   /**
+     * Get magentometer measurement in tesla
+     * @param[in] magName name of the magnetometer
+     * @param[out] magMeasurement magnetometer measurements of size 3
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     * @return true/false in case of success/failure
+     */
+    virtual bool getMagnetometerMeasurement(const std::string& magName,
+                                            Eigen::Ref<Eigen::Vector3d> magMeasurement,
+                                            double* receiveTimeInSeconds = nullptr) = 0;
+
+    /**
+     * Get six axis force torque measurement
+     * @param[in] ftName name of the FT sensor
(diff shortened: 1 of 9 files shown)


S-Dafarra:
> Adds OpenCV as a dependency for handling cv::Mat within the ISensorBridge class

To be honest, I think that ``OpenCV`` may be a too strong dependency for a single type. It should be possible to convert a ``cv::Mat`` type from/to an Eigen type, https://github.com/opencv/opencv/blob/e3d502310f6ae0cad79f3474b9ea2b25705985f9/modules/core/include/opencv2/core/eigen.hpp

What about using Eigen types in the interface then?

prashanthr05:
> To be honest, I think that `OpenCV` may be a too strong dependency for a single type. It should be possible to convert a `cv::Mat` type from/to an Eigen type, https://github.com/opencv/opencv/blob/e3d502310f6ae0cad79f3474b9ea2b25705985f9/modules/core/include/opencv2/core/eigen.hpp
> 
> What about using Eigen types in the interface then?


I thought about it too. If we can depend on the user to make relevant conversions from Eigen to relevant types, then we can keep the design that way.

In that case, would it also make sense to change GenericVectorContainer to Eigen?

prashanthr05:
But then, if we also want point clouds to be handled by this Bridge, then there comes a problem of adding PCL which is quite a heavy library that depends on many other libraries as a dependency. 

So we might have to think about this a bit.

S-Dafarra:
> In that case, would it also make sense to change GenericVectorContainer to Eigen?

The ``GenericContainer::Vector`` was meant to have a unique way to deal with vectors that could have also been resized. This was coming in handy when dealing with parameters. Depending on the implementation and on the needs, we may have needed a ``Yarp::Sig::Vector``, or ``iDynTree::VectorDynSize`` for example. This was before moving to Eigen in the main API. So, right now I would suggest using  ``GenericContainer::Vector`` in case you need to deal with different vector implementation at the same time, especially in case you need to resize them (something that is not possible with an ``Eigen::map``).

> But then, if we also want point clouds to be handled by this Bridge, then there comes a problem of adding PCL which is quite a heavy library that depends on many other libraries as a dependency.

Is there a specific type/class we may need to use from this library?

prashanthr05:
> The GenericContainer::Vector was meant to have a unique way to deal with multiple vector implementations that could have also been resized. This was coming in handy when dealing with parameters. Depending on the implementation and on the needs, we may have needed a Yarp::Sig::Vector, or iDynTree::VectorDynSize for example. This was before moving to Eigen in the main API. So, right now I would suggest using GenericContainer::Vector in case you need to deal with different vector implementation at the same time, especially in case you need to resize them (something that is not possible with an Eigen::map).

Agreed. 

> Is there a specific type/class we may need to use from this library?

I am just looking forward into our tasks. For example, for the staircase climbing task, if we need perception in the loop, we could begin to use PCL library extensively for plane segmentation. But from within the class, we would need pcl::PoincloudXYZRGB.

S-Dafarra:
> I am just looking forward into our tasks. For example, for the staircase climbing task, if we need perception in the loop, we could begin to use PCL library extensively for plane segmentation. But from within the class, we would need pcl::PoincloudXYZRGB.

I think there might be some conversions available. https://github.com/PointCloudLibrary/pcl/blob/2c8d58a0e6497d91cc50f61460d75b9ebf44429a/common/include/pcl/point_cloud.h#L351

Anyway, the take-home message would be to avoid heavy dependencies only for a type in an interface class. For sure, if we need some additional feature of these libraries, it makes total sense to add them. In principle, I may want to avoid having a certain dependency, paying the price of losing some features. But I am expecting the feature to be related to the dependency I am missing. In this specific case, it would be a pity not to be able to compile the ``RobotInterface`` part if OpenCV is missing. This may complicate things if we would need this to run on the robot head maybe.

prashanthr05:
> Anyway, the take-home message would be to avoid heavy dependencies only for a type in an interface class. For sure, if we need some additional feature of these libraries, it makes total sense to add them. In principle, I may want to avoid having a certain dependency, paying the price of losing some features. But I am expecting the feature to be related to the dependency I am missing. In this specific case, it would be a pity not to be able to compile the RobotInterface part if OpenCV is missing. This may complicate things if we would need this to run on the robot head maybe

I understand. In that case, we can avoid having OpenCV and PCL as a dependency for this interface class and rely on relevant conversions at a higher level. I will replace the image related methods to Eigen types. 
For the PCL, we shall deal with it in the near future.

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -0,0 +1,296 @@
...
+
+    /**
+     * Get rgb cameras list
+     */
+    virtual const std::vector<std::string> getRGBCamerasList() = 0;
+
+    /**
+     * Get depth cameras list
+     */
+    virtual const std::vector<std::string> getDepthCamerasList() = 0;
```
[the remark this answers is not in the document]
prashanthr05: Oh yes.. that was the initial intention, but I changed my mind since it would depend on whether the implementation will hold the data structure in their class. So I thought return by value would be better, but I forgot to remove the const.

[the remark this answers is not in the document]
S-Dafarra: This would be a big waste of memory. I think that, in this case, the possible choices would be:
- return by reference
- have a reference to the vector as an input to the method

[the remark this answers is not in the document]
prashanthr05: I think we can go for the latter without having to impose the maintenance of those data structures by the implementation class.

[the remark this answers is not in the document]
S-Dafarra: As you prefer!

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -0,0 +1,296 @@
...
+    /**
+     * Get joint position
+     * @param[in] jointName name of the joint
+     * @param[out] jointPosition joint position in radians
+     * @param[out] recvTimeInS receive time in seconds
+     * @return true/false in case of success/failure
+     */
+    virtual bool getJointPosition(const std::string& jointName,
+                                  double& jointPosition,
+                                  double& recvTimeInS) = 0;
```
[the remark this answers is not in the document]
prashanthr05: > Even if it is a little uglier, a possibility could be to input a double pointer instead of a reference, and default it to a nullptr. Then, inside, there should be a check on whether this pointer is null or not. What do you think?

It's not really clear to me the purpose of this suggestion? Could you explain a bit more?

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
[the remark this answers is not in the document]
prashanthr05: But actually, to be extra safe, just not return the reference of a string that "might" be local, I will use the second choice.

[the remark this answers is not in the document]
S-Dafarra: > But actually, to be extra safe, just not return the reference of a string that "might" be local, I will use the second choice.

I guess you meant ``vector`` and not ``string``, but yeah, I agree there is the risk that in a faulty implementation a reference to a local object is returned.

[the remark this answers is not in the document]
S-Dafarra: Actually, I realized later that these methods are probably called just once, so even return them by copy is not a big deal. So, feel free to return directly the vector if you believe the interface would be cleaner.

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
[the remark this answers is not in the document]
S-Dafarra: I was thinking at something like:
```c++
virtual bool getSomething(SomeType& something,
                                  double* recvTimeInS = nullptr) = 0;
```
In this way, you can call it like
```c++
SomeType output;
getSomething(output);
```
and if you are also interested in the time
```c++
double time;
SomeType output;
getSomething(output, &time);
```

Another possibility could be also to have both the methods with and without ``recvTimeInS``, where the latter simply calls the former with a dummy ``double``, i.e.
```c++
virtual bool getSomething(SomeType& something,
                                  double& recvTimeInS) = 0;

bool getSomething(SomeType& something,
                                  double& recvTimeInS) //Note, this is not virtual
{
    double dummy;
    return getSomething(something, dummy);
}
```
Does it make sense?

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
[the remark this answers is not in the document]
prashanthr05: Actually I already changed them to be passed by reference and then copied. We can go with this choice, which can also handle failure cases through the boolr eturn.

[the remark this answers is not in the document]
S-Dafarra: Sounds good!

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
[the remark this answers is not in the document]
prashanthr05: Ah, ok! I got your point now. I will make the changes corresponding to the former choice.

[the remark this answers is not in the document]
prashanthr05: Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/87/commits/3bbf33a17441c9a8ed86cf5c2528659a2cbb5f5e

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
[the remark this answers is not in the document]
prashanthr05: Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/87/commits/666670f08e102f679fff287bc896cadc7b91de14

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -0,0 +1,318 @@
...
+    bool isGyroscopeEnabled{false}; /**< flag to connect gyroscope measurement sources */
+    bool isMagnetometerEnabled{false}; /**< flag to connect magnetometer measurement sources */
+    bool isSixAxisForceTorqueSensorEnabled{false}; /**< flag to connect six axis force torque measurement sources */
+    bool isThreeAxisForceTorqueSensorEnabled{false}; /**< flag to connect six axis force torque measurement sources */
+    bool isCartesianWrenchEnabled{false}; /**< flag to connect cartesian wrench measurement sources */
+    bool isCameraEnabled{false}; /**< flag to connect camera sources */
+};
+
+/**
+ *  Senosr lists
```
S-Dafarra, line 42: Typo

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -0,0 +1,318 @@
...
+
+namespace BipedalLocomotion
+{
+namespace RobotInterface
+{
+
+/**
+ * Sensor bridge options
+ */
+struct SensorBridgeOptions
```
S-Dafarra, line 28: How are you planning to use these options?

src/RobotInterface/tests/DummyImplementation/DummySensorBridgeTest.cpp
```diff
@@ -0,0 +1,55 @@
...
+ * distributed under the terms of the GNU Lesser General Public License v2.1 or any later version.
+ */
+
+
+// Catch2
+#include <catch2/catch.hpp>
+#include <BipedalLocomotion/ParametersHandler/IParametersHandler.h>
+#include <BipedalLocomotion/ParametersHandler/StdImplementation.h>
+
+#include "./DummySensorBridge.h"
```
S-Dafarra, line 14: I guess you can remove ``./``. When using ``"``, it already looks in the current folder.

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -0,0 +1,318 @@
...
+
+    /**
+     * Get linear accelerometer measurement in m/s^2
+     * @param[in] accName name of the linear accelerometer
+     * @param[out] accMeasurement linear accelerometer measurements of size 3
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     * @return true/false in case of success/failure
+     */
+    virtual bool getLinearAccelerometerMeasurement(const std::string& accName,
+                                                   GenericContainer::Vector<double>& accMeasurement,
```
GiulioRomualdi, line 218: Is it a good idea to use the `GenericContainer::Vector` here? In all the other parts of the code, we use Eigen vectors as intefrace of the class. I would avoid having mixing different kinds of types

GiulioRomualdi: For instance here you can use an `Eigen::Ref<Eigen::Vector3d>` 
And the you can pass every different container (std vector iDynTree or yarp) through an `Eigen::Map`

prashanthr05: We had a discussion about this at https://github.com/dic-iit/bipedal-locomotion-framework/pull/87#issuecomment-682450524.

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
prashanthr05: It's just like a board of switches to turn on/off the processing during the initialization step. An instance of the SensorBridge will not consider attaching itself to sensor streams that are not enabled.

S-Dafarra: It makes sense but I did not see it used in the interface. What would be your plan?

prashanthr05: The intention was that the Derived class will hold private instances of `SensorBridgeOptions` and `SensorLists` for internal organization. I shall add a line of documentation on how it can be used.

S-Dafarra: Ok. In any case, my suggestion would be to declare what you use and leave freedom to the user for the rest. If you declare some struct in the base interface, but the base interface does not use them (neither as an attribute nor as an input parameter to a method), then there is a big chance they are never used, or used in a wrong way 😁

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
prashanthr05: The important point that was raised was,

> So, right now I would suggest using GenericContainer::Vector in case you need to deal with different vector implementation at the same time, especially in case you need to resize them (something that is not possible with an Eigen::map).

@S-Dafarra ,

> I would avoid having mixing different kinds of types

would you have an opinion on this?

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
prashanthr05: I wanted to have them within the RobotInterface namespace. Would it be better if I moved it to a new header keeping it separate from the interface?

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
S-Dafarra: > @S-Dafarra ,
> 
> > I would avoid having mixing different kinds of types
> 
> would you have an opinion on this?

I would say that it depends on whether ``accMeasurement`` has a predefined size or not. If so, I agree on using Eigen types for consistency.

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
S-Dafarra: Do they use it in the same way? My main point is that either it is clear from the code how to use these structs, or it may be better to leave the freedom to the user.

A suggestion I may give is to have a function that automatically fills this struct given a pointer to a parameter handler, for example.

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
prashanthr05: There are a few quantities that have might have a variable size like images, joint states, etc. Otherwise, the rest are all fixed size.

Anyhow, these are going to remain fixed once the user configures the number of joints or the image dimensions.

GiulioRomualdi: > There are a few quantities that have might have a variable size like images, joint states, etc. Otherwise, the rest are all fixed size.

The images I think are stored in matrices right? If yes in this case `GenericContainer::Vector` cannot be used. For the joint state back in time I simply used `iDynTree::VectorDynSize`, using Eigen the code may become
```cpp
bool SensorBridge::getJointPos(Eigen::Ref<Eigen::VectorXd> v);

....

main()
{
  SensorBridge sb;
  iDynTree::VectorDynSize q_iDynTree;
  yarp::Vector q_yarp;
  Eigen::VectorXd q_eigen;
  sb.getJointPos(iDynTree::toEigen(q_iDynTree));
  sb.getJointPos(yarp::toEigen(q_yarp));
  sb.getJointPos(yarp::toEigen(q_eigen));
}
```

GiulioRomualdi: > > There are a few quantities that have might have a variable size like images, joint states, etc. Otherwise, the rest are all fixed size.
> 
> The images I think are stored in matrices right? If yes in this case `GenericContainer::Vector` cannot be used. For the joint state, back in time I simply used `iDynTree::VectorDynSize`, using Eigen the code may
> 
> ```c++
> bool SensorBridge::getJointPos(Eigen::Ref<Eigen::VectorXd> v);
> 
> ....
> 
> main()become
> {
>   SensorBridge sb;
>   iDynTree::VectorDynSize q_iDynTree;
>   yarp::Vector q_yarp;
>   Eigen::VectorXd q_eigen;
>   sb.getJointPos(iDynTree::toEigen(q_iDynTree));
>   sb.getJointPos(yarp::toEigen(q_yarp));
>   sb.getJointPos(q_eigen);
> }
> ```

By the way this holds only if we don't want to resize the vector inside the `getJointPos()` because in this case, terrible bugs may happen 
```cpp

Eigen::VectorXd q_eigen(100);
 
sb.getJointPos(q_eigen.segment<10>(20)); //<----- If getJointPos resize the vectors terrible things may happen here 
```

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
prashanthr05: I can add a protected method in the interface as per your suggestion.

S-Dafarra: Sounds good!

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
prashanthr05: Relevant changes made in https://github.com/dic-iit/bipedal-locomotion-framework/pull/87/commits/4a4b818f9bb3ad844993a7a6596c16844e7af942

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
prashanthr05: I added optional helper methods in https://github.com/dic-iit/bipedal-locomotion-framework/pull/87/commits/4a4b818f9bb3ad844993a7a6596c16844e7af942 to avoid the structs from remaining dangling.
I added a line of documentation which will allow the user to also avoid using these structs if they decide to implement their own better ways.

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
GiulioRomualdi: Thank you for handling this :heart:

GiulioRomualdi (approved):

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -0,0 +1,349 @@
...
+protected:
+    /**
+     * Helper method to maintain SensorBridgeOptions struct by populating it from the configuration parameters
+     * @note the user may choose to use/not use this method depending on their requirements for the implementation
+     * if the user chooses to not use the method, the implementation must simply contain "return true;"
+     *
+     * @param[in] handler  Parameters handler
+     * @param[in] sensorBridgeOptions SensorBridgeOptions to hold the bridge options for streaming sensor measurements
+     */
+    virtual bool populateSensorBridgOptionsFromConfig(std::weak_ptr<IParametersHandler> handler,
```
S-Dafarra, line 328: ``Bridge``

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -0,0 +1,349 @@
...
+                                                      SensorBridgeOptions& sensorBridgeOptions) = 0;
```
S-Dafarra, line 329: If it is not compulsory, simply return true directly in the header. Otherwise, you can already provide a basic implementation that can be overridden by the user, or even not used since it is protected.

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
S-Dafarra: So, how is it handled for the elements with variable size?

prashanthr05: At the moment, we assume for joint states and images, that the size is decided at the configuration and remains fixed, and internal checks can be done at the implementation level. This means that the user must pass a resized argument to these getter methods. Correct me if I am wrong, @GiulioRomualdi!
I shall add a description in the documentation regarding this.

However, we need to understand how to handle this for the elements with variable size, for example, point clouds.

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
prashanthr05: I refrained myself from doing that to keep this class abstract. If you think it is not necessarily important to remain as a pure virtual class, I shall add the implementation that simply returns true in the header.

S-Dafarra: Yes, in C++ a class is abstract if at least a method is purely virtual, so no big deal.

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
S-Dafarra: Ok for the documentation. For point clouds, are they matrices or vector of points?

prashanthr05: They are usually handled as a struct asking for maintaining cloud width and height. Then they store a vector of XYZ points or XYZRGB points. 

I hope we will not be causing any overheads by doing multiple conversions by passing these kinds of images or point cloud measurements through this bridge.

S-Dafarra: I think may open one or more issue to discuss on this. In particular, it is not clear:
- when to use ``Eigen`` vectors vs ``GenericContainer`` vs templates
- how to deal with matrices
- how to deal with particular input classes, like point clouds

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
prashanthr05: > A class is made abstract by declaring at least one of its functions as a pure virtual function.

Ah, yes! I was thinking conversely on this.

 I will make the suggested changes then (returning true would be easy :D).

Thank you for this suggestion.

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
prashanthr05: I have added documentation for the joint states and image methods and relevant configuration parameters in the `SensorBridgeOptions` if the implementation might want to use it. https://github.com/dic-iit/bipedal-locomotion-framework/pull/87/commits/979733e1357ffed832c1533a6c6c464adcdae6b8

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
prashanthr05: Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/87/commits/979733e1357ffed832c1533a6c6c464adcdae6b8

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
prashanthr05: Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/87/commits/979733e1357ffed832c1533a6c6c464adcdae6b8

S-Dafarra (approved):

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -0,0 +1,373 @@
...
+#define BIPEDAL_LOCOMOTION_ROBOT_INTERFACE_ISENSOR_BRIDGE_H
+
+#include <memory>
+#include <string>
+#include <vector>
+#include <unordered_map>
+
+#include <Eigen/Dense>
+
+#include <BipedalLocomotion/GenericContainer/Vector.h>
```
GiulioRomualdi, line 18: This one is not required here, right?

prashanthr05: Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/87/commits/699ea3f560fc27118e23d9aa30f562487e0ef527

GiulioRomualdi:
@prashanthr05 let me know if you want to reorganize the commits otherwise I will merge the PR

prashanthr05:
> @prashanthr05 let me know if you want to reorganize the commits otherwise I will merge the PR

Ciao @GiulioRomualdi .. I have squashed and rebased all changes into a single commit. Thanks!

PR #92, opened 2020-09-03
Add the Debug compilation in GitHub Action

### LINKED ISSUE #91: Enable the compilation in Debug for the CI
Starting from #63 we started to use `Eigen` actively on the framework. As you probably know `Eigen` suppresses several checks if the consumer library/application is compiled in `Release`, this may lead to undesired behavior that are not easy to spot. See for instance https://github.com/dic-iit/bipedal-locomotion-framework/pull/85#issuecomment-686605967.

Enabling the Debug tag on CI may help the developer to spot some problems. On the other hand, it will increase the compilation time on the GitHub action.

What do you think @S-Dafarra @prashanthr05 @traversaro?

GiulioRomualdi:
This closes #91

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ env:
   YARP_TAG: 964bb26fa4791d83d72882711ea7509306248106
   iDynTree_TAG: v1.1.0
   Catch2_TAG: v2.11.3
-  Qhull_TAG: v8.0.0
+  Qhull_TAG: f7aff465101711ae4cee3f501a528d6d53e75185
   CasADi_TAG: 3.5.1
   manif_TAG: 56bb7744aea51820bd80875b4ee4369eb7d52dfb
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,7 +85,7 @@ jobs:
       with:
         path: ${{ github.workspace }}/install/deps
         # Including ${{ runner.temp }} is a workaround taken from https://github.com/robotology/whole-body-estimators/pull/62 to fix macos configuration failure on https://github.com/dic-iit/bipedal-locomotion-framework/pull/45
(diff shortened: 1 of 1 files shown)


S-Dafarra (approved):

prashanthr05 (approved):

prashanthr05:
I think the bipedal tests CMake patch as described in https://github.com/dic-iit/bipedal-locomotion-framework/pull/85#issuecomment-686605967 can be added within the scope of this PR?

I noticed that the Windows CI failure had similar errors.

prashanthr05:
> I think the bipedal tests CMake patch as described in #85 (comment) can be added within the scope of this PR?

No, let's keep it separate. (I answer myself ;D)

GiulioRomualdi:
On windows and ubuntu the Debug compilation fails because of `qhull v8.0.0`
Indeed when `qhull` is compiled in debug the a postfix `_d` is added to the library name.
```cmake
 if(CMAKE_BUILD_TYPE MATCHES "[dD]ebug")
     set(qhull_CPP qhullcpp_d)
     set(qhull_SHARED qhull_d)
     set(qhull_SHAREDP qhull_pd)
     set(qhull_SHAREDR qhull_rd)
     set(qhull_STATIC qhullstatic_d)
     set(qhull_STATICR qhullstatic_rd)
 else()
     set(qhull_CPP qhullcpp)
     set(qhull_SHARED libqhull)  # Temporarily avoid name conflict with qhull executable                                                                                                                     
     set(qhull_SHAREDP qhull_p)
     set(qhull_SHAREDR qhull_r)
     set(qhull_STATIC qhullstatic)
     set(qhull_STATICR qhullstatic_r)
 endif()
```
This creates a problem when I try to link the library here:
https://github.com/dic-iit/bipedal-locomotion-framework/blob/2d31ad55757153952e30c92328a48c681dd0642d/src/Planners/CMakeLists.txt#L22

This commits fixes the problem: https://github.com/qhull/qhull/commit/16159c648c36d79effbd537c1375b20e63398ed8#diff-af3b638bc2a3e6c650974192a53c7291

For this reason, I will update the  version used in the CI

PR #93, opened 2020-09-03
Add USE_MATH_DEFINES for tests and library compile definitions

prashanthr05:
See https://github.com/dic-iit/bipedal-locomotion-framework/pull/85#issuecomment-686605967

### DIFF
--- a/cmake/AddBipedalLocomotionLibrary.cmake
+++ b/cmake/AddBipedalLocomotionLibrary.cmake
@@ -43,7 +43,6 @@ function(add_bipedal_locomotion_library)
 
     # Add C++17 features
     target_compile_features(${name} INTERFACE cxx_std_17)
-
     target_link_libraries(${name} INTERFACE ${public_link_libraries})
 
     # Specify include directories for both compilation and installation process.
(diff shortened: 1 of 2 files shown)


S-Dafarra:
I don't remember. Was this definition useful only for executables? Does it make sense also to add this definition for the libraries?

prashanthr05:
> I don't remember. Was this definition useful only for executables? Does it make sense also to add this definition for the libraries?


I think it makes sense for libraries as well. In general, anything that would pertain the using math constants.

S-Dafarra (approved):

prashanthr05:
I made a mistake in copy-paste ;D and another mistake of not building before pushing.

Anyways, shall we also add the definitions for interface?

S-Dafarra:
> I made a mistake in copy-paste ;D and another mistake of not building before pushing.
> 
> Anyways, shall we also add the definitions for interface?

And I did not notice it :joy: In theory, interface targets are not built. So it should be handled by user classes I guess :thinking:

prashanthr05:
> And I did not notice it In theory, interface targets are not built. So it should be handled by user classes I guess 

Alright. we shall add those definitions only for libraries then.

GiulioRomualdi (approved):
Merging! Thank you @prashanthr05

PR #106, opened 2020-09-13
YarpSensorBridge Implementation

prashanthr05:
This PR adds the YarpSensorBridge implementation that refactors out our traditional way of getting sensor measurements into a library. The current implementation supports measurements read from
- a remote control board remapper implementing encoders interface
- a multiple analog sensor remapper implementing accelerometers, gyroscopes, magnetometers, orientation sensors, and force-torque sensors interface
- generic IMU sensors through a server inertial
- analog FT sensors
- cartesian wrenches opened as generic sensors considered to have dimension size 6

It needs to be configured with a PolyDriverList of the relevant devices that are available in the network, and a configuration setup that informs the SensorBridge about the devices and the underlying sensor interfaces.

Implementation roadmap
- [X] Configure step
- [X] Attach drivers step
- [x] Read sensors
- [x] Advance and Getters (missing implementation for image getters, Eigen Ref conversions of YARP Image needs to be understood).
- [x] Integrated testing with a YARP device (reading one IMU, two cartesian wrenches, and a remote control board remapper takes about an average of 50 microseconds). The device closed without any segmentation faults. On comparing IMU measurements with ports and those given by the bridge, they matched (except for those converted to radians). Should I add the test device to BLF?

- **IMPORTANT CHANGE** In `ISensorBridge` Interface class, I made the getter implementations for the sensors as optional by changing them from being pure virtual functions to virtual functions. This allows the developer to decide what they want to implement rather than polluting their API with unnecessary and unused methods.

Points to discuss,
- P.S. The current monolithic Impl source code design could be further refactored into multiple Impl classes for each SensorTypes. However, for starters, let's begin with this.

- **Caution for a possible design flaw** I Inherited the `YarpSensorBridge` from the `Advanceable` class as well. However, I have a feeling that to poll the measurements at each advance step and have a huge set of measurements stored internally would be expensive than polling for the specific measurement whenever the user calls the getXXX() method on a very specific sensor of a specific measurement type.  Also since within the advance() step all the sensor reads are done serially (no sense of parallelism in internal reads as well), this might cause a huge overhead in the pipeline. I hope this doesn't turn out to be a huge design flaw in the future.  If it does, we might have to reuse the read individual sensor methods in either a parallel way or to call the individual sensor reads while the user calls the getters. 
**(Imagine a single instance of SensorBridge configured with all sensor interfaces including cameras and multiple interfaces!)**

- Dealing with YARP images, I am using internal storage of `yarp::sig::Image` and dynamic casting it to the derived classes of `yarp::sig::ImageOf<yarp::sig::PixelRgb>`, `yarp::sig::ImageOf<yarp::sig::PixelFloat>` and `yarp::sig::FlexImage` for reading RGB Images from `IFrameGrabber` interface, depth images from `IRGBDSensor` interface and RGB images from `IRGBDSensor` interface. We need to understand if it's a good way to do so.

### DIFF
--- a/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
+++ b/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
@@ -87,7 +87,8 @@ struct YarpSensorBridge::Impl
     std::unordered_map<std::string, Vector6d> wholeBodyFTMeasures; /** < map holding six axis force torque measures */
     std::unordered_map<std::string, Eigen::Vector3d> wholeBodyInertialMeasures; /** < map holding three axis inertial sensor measures */
     std::unordered_map<std::string, Vector6d> wholeBodyCartesianWrenchMeasures; /** < map holding cartesian wrench measures */
-    std::unordered_map<std::string, Eigen::MatrixXd> wholeBodyCameraImages; /** < map holding images **/
+    std::unordered_map<std::string, Eigen::MatrixXd> wholeBodyCameraRGBImages; /** < map holding images **/
+    std::unordered_map<std::string, Eigen::MatrixXd> wholeBodyCameraDepthImages; /** < map holding images **/
 
     const int nrChannelsInYARPGenericIMUSensor{12};
     const int nrChannelsInYARPGenericCartesianWrench{6};
@@ -282,41 +283,41 @@ struct YarpSensorBridge::Impl
             {
                 std::pair<int, int> imgDimensions(rgbWidth[idx], rgbHeight[idx]);
                 auto cameraName{metaData.sensorsList.rgbCamerasList[idx]};
-                metaData.bridgeOptions.imgDimensions[cameraName] = imgDimensions;
+                metaData.bridgeOptions.rgbImgDimensions[cameraName] = imgDimensions;
             }
         }
 
-        if (ptr->getParameter("depth_cameras_list", metaData.sensorsList.depthCamerasList, VectorResizeMode::Resizable))
+        if (ptr->getParameter("rgbd_cameras_list", metaData.sensorsList.rgbdCamerasList, VectorResizeMode::Resizable))
         {
             metaData.bridgeOptions.isCameraEnabled = true;
 
-            std::vector<int> depthCamWidth, depthCamHeight;
-            if (ptr->getParameter("depth_image_width", depthCamWidth, VectorResizeMode::Resizable))
+            std::vector<int> rgbdCamWidth, rgbdCamHeight;
+            if (ptr->getParameter("rgbd_image_width", rgbdCamWidth, VectorResizeMode::Resizable))
             {
-                std::cerr << logPrefix << " Required parameter \"depth_image_width\" not available in the configuration"
+                std::cerr << logPrefix << " Required parameter \"rgbd_image_width\" not available in the configuration"
                           << std::endl;
                 return false;
             }
 
-            if (ptr->getParameter("depth_image_height", depthCamHeight, VectorResizeMode::Resizable))
+            if (ptr->getParameter("rgbd_image_height", rgbdCamHeight, VectorResizeMode::Resizable))
             {
-                std::cerr << logPrefix << " Required parameter \"depth_image_height\" not available in the configuration"
+                std::cerr << logPrefix << " Required parameter \"rgbd_image_height\" not available in the configuration"
                           << std::endl;
                 return false;
             }
 
-            if ( (depthCamWidth.size() != metaData.sensorsList.depthCamerasList.size()) ||
-                (depthCamHeight.size() != metaData.sensorsList.depthCamerasList.size()) )
+            if ( (rgbdCamWidth.size() != metaData.sensorsList.rgbdCamerasList.size()) ||
+                (rgbdCamHeight.size() != metaData.sensorsList.rgbdCamerasList.size()) )
             {
                 std::cerr << logPrefix << " Parameters list size mismatch" << std::endl;
                 return false;
             }
 
-            for (int idx = 0; idx < depthCamHeight.size(); idx++)
+            for (int idx = 0; idx < rgbdCamHeight.size(); idx++)
             {
-                std::pair<int, int> imgDimensions(depthCamWidth[idx], depthCamHeight[idx]);
-                auto cameraName{metaData.sensorsList.depthCamerasList[idx]};
-                metaData.bridgeOptions.imgDimensions[cameraName] = imgDimensions;
+                std::pair<int, int> imgDimensions(rgbdCamWidth[idx], rgbdCamHeight[idx]);
+                auto cameraName{metaData.sensorsList.rgbdCamerasList[idx]};
+                metaData.bridgeOptions.rgbdImgDimensions[cameraName] = imgDimensions;
             }
         }
 
@@ -955,21 +956,29 @@ struct YarpSensorBridge::Impl
         std::string_view interfaceType{"RGB Cameras"};
         if (!attachAllCamerasOfSpecificType(devList,
                                             metaData.sensorsList.rgbCamerasList,
-                                            metaData.bridgeOptions.imgDimensions,
+                                            metaData.bridgeOptions.rgbImgDimensions,
                                             interfaceType,
                                             wholeBodyFrameGrabberInterface,
-                                            wholeBodyCameraImages))
+                                            wholeBodyCameraRGBImages))
         {
             return false;
         }
 
-        std::string_view interfaceTypeDepth = "Depth Cameras";
+        std::string_view interfaceTypeDepth = "RGBD Cameras";
         if (!attachAllCamerasOfSpecificType(devList,
-                                            metaData.sensorsList.depthCamerasList,
-                                            metaData.bridgeOptions.imgDimensions,
+                                            metaData.sensorsList.rgbdCamerasList,
+                                            metaData.bridgeOptions.rgbdImgDimensions,
                                             interfaceTypeDepth,
                                             wholeBodyRGBDInterface,
-                                            wholeBodyCameraImages))
+                                            wholeBodyCameraDepthImages))
+        {
+            return false;
+        }
+
+        // resize also rgb images of RGBD cameras
+        if (!resizeImageBuffers(metaData.sensorsList.rgbdCamerasList,
+                                metaData.bridgeOptions.rgbdImgDimensions,
+                                wholeBodyCameraRGBImages))
         {
             return false;
         }
@@ -999,7 +1008,7 @@ struct YarpSensorBridge::Impl
 
         if (sensorMap.size() != camList.size())
         {
-            std::cout << logPrefix << " could not attach all desired rgb cameras " << interfaceType  << "." << std::endl;
+            std::cout << logPrefix << " could not attach all desired cameras of type " << interfaceType  << "." << std::endl;
             return false;
         }
 
--- a/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
+++ b/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
@@ -1,0 +1,1042 @@
+/**
+ * @file YarpSensorBridgeImpl.h
+ * @authors Prashanth Ramadoss
+ * @copyright 2020 Istituto Italiano di Tecnologia (IIT). This software may be modified and
+ * distributed under the terms of the GNU Lesser General Public License v2.1 or any later version.
+ */
+
+#ifndef BIPEDAL_LOCOMOTION_ROBOT_INTERFACE_YARP_SENSOR_BRIDGE_IMPL_H
+#define BIPEDAL_LOCOMOTION_ROBOT_INTERFACE_YARP_SENSOR_BRIDGE_IMPL_H
+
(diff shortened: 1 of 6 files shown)


src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
```diff
@@ -0,0 +1,1695 @@
...
+
+    const int nrChannelsInYARPGenericIMUSensor{12};
+    const int nrChannelsInYARPGenericCartesianWrench{6};
+    const int nrChannelsInYARPAnalogSixAxisFTSensor{6};
+
+    std::vector<std::string> failedSensorReads;
+    SensorBridgeMetaData metaData; /**< struct holding meta data **/
+    bool bridgeInitialized{false}; /**< flag set to true if the bridge is successfully initialized */
+    bool driversAttached{false}; /**< flag set to true if the bridge is successfully attached to required device drivers */
+    bool sensorDryRunEnabled{false}; /**< flag to enable running a test stream of sensor interfaces after attaching to the device drivers */
```
prashanthr05, line 121: This flag is unused. I am not running a dry run read sensor anywhere!

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
```diff
@@ -0,0 +1,1695 @@
...
+#include <yarp/dev/IEncodersTimed.h>
+#include <yarp/dev/IAxisInfo.h>
+
+// std
+#include <cmath>
+#include <set>
+#include <algorithm>
+
+using namespace BipedalLocomotion::RobotInterface;
+using namespace BipedalLocomotion::GenericContainer;
```
GiulioRomualdi, line 40: You should not use `using namespace` in the header file. Here the explanation: https://stackoverflow.com/questions/5849457/using-namespace-in-c-headers

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
```diff
@@ -0,0 +1,1695 @@
...
+    using StampedYARPImage = std::pair<yarp::sig::Image, double>;
+
+    /**
+     * Struct holding remapped remote control board interfaces
+     */
+    struct
+    {
+        yarp::dev::IEncodersTimed* encoders;
+        yarp::dev::IAxisInfo* axis;
+    } controlBoardRemapperInterfaces;
```
GiulioRomualdi, line 59: This is the c style for defining a struct you can write  
```suggestion
    struct ControlBoardRemapperInterfaces
    {
        yarp::dev::IEncodersTimed* encoders;
        yarp::dev::IAxisInfo* axis;
    };
```

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
```diff
@@ -0,0 +1,1695 @@
...
+    /**
+     * Struct holding remapped MAS interfaces -inertial sensors related
+     */
+    struct
+    {
+        yarp::dev::IThreeAxisLinearAccelerometers* accelerometers;
+        yarp::dev::IThreeAxisGyroscopes* gyroscopes;
+        yarp::dev::IThreeAxisMagnetometers* magnetometers;
+        yarp::dev::IOrientationSensors* orientationSensors;
+    } wholeBodyMASInertialsInterface;
```
GiulioRomualdi, line 70: Same as above

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
```diff
@@ -0,0 +1,1695 @@
...
+
+    /**
+     * Configure cartesian wrenches meta data
+     */
+    bool configureCartesianWrenches(std::weak_ptr<IParametersHandler> handler,
+                                    SensorBridgeMetaData& metaData)
+    {
+        constexpr std::string_view logPrefix = "[YarpSensorBridge::Impl::configureCartesianWrenchSensors] ";
+        auto ptr = handler.lock();
+        if (ptr == nullptr) { return false; }
```
GiulioRomualdi, line 272:
```suggestion
        if (ptr == nullptr) 
        {
             return false;
        }
```

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
```diff
@@ -0,0 +1,1695 @@
...
+
+    /**
+     * Configure cameras meta data
+     */
+    bool configureCameras(std::weak_ptr<IParametersHandler> handler,
+                          SensorBridgeMetaData& metaData)
+    {
+        constexpr std::string_view logPrefix = "[YarpSensorBridge::Impl::configureCameras] ";
+        auto ptr = handler.lock();
+        if (ptr == nullptr) { return false; }
```
GiulioRomualdi, line 292:
```suggestion
        if (ptr == nullptr) 
        {
             return false;
        }
```

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
```diff
@@ -0,0 +1,1695 @@
...
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
+     * Attach to cartesian wrench interface
+     */
+    bool attachCartesianWrenchInterface(const yarp::dev::PolyDriverList& devList)
```
GiulioRomualdi, line 1007: :heart:

src/RobotInterface/YarpImplementation/src/YarpSensorBridge.cpp
```diff
@@ -0,0 +1,440 @@
...
+    {
+        return false;
+    }
+    rgbdCamerasList = m_pimpl->metaData.sensorsList.rgbdCamerasList;
+    return true;
+}
+
+bool YarpSensorBridge::getJointPosition(const std::string& jointName,
+                                        double& jointPosition,
+                                        double* receiveTimeInSeconds)
```
GiulioRomualdi, line 223: I'm missing something for sure, but is there a reason why you chose to use the pointer instead of a reference? (The same comment is valid for the others methods)

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -35,14 +35,16 @@ struct SensorBridgeOptions
...
     bool isMagnetometerEnabled{false}; /**< flag to connect magnetometer measurement sources */
     bool isSixAxisForceTorqueSensorEnabled{false}; /**< flag to connect six axis force torque measurement sources */
     bool isThreeAxisForceTorqueSensorEnabled{false}; /**< flag to connect six axis force torque measurement sources */
     bool isCartesianWrenchEnabled{false}; /**< flag to connect cartesian wrench measurement sources */
     bool isCameraEnabled{false}; /**< flag to connect camera sources */
 
     size_t nrJoints{0}; /**< number of joints available through Kinematics stream, to be configured at initialization */
-    std::unordered_map<std::string, std::pair<int, int> > imgDimensions; /**< dimensions of the camera images available through camera streams, to be configured at initialization */
+    std::unordered_map<std::string, std::pair<int, int> > rgbImgDimensions; /**< dimensions of the images available through rgb camera streams, to be configured at initialization */
+    std::unordered_map<std::string, std::pair<int, int> > rgbdImgDimensions; /**< dimensions of the depth images available through rgbd camera streams, to be configured at initialization */
```
GiulioRomualdi, line 47: since the dimension is always a positive number probably you can use `std::size_t`

GiulioRomualdi (commented):
Some minor comments except https://github.com/dic-iit/bipedal-locomotion-framework/pull/106#discussion_r489437205

src/RobotInterface/YarpImplementation/src/YarpSensorBridge.cpp
prashanthr05: To make it optional.. in the headers it's defined with a default `nullptr` assignment.
This was suggested by @S-Dafarra because some times, we are not interested in timestamps, so we can just directly get the measurement.

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
prashanthr05: Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/106/commits/a50d061e44dd2acc504bcd578df9f24901a0073f

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
prashanthr05: Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/106/commits/a50d061e44dd2acc504bcd578df9f24901a0073f

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
prashanthr05: Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/106/commits/0593bcfbf32c11c49d2c9504bbc0aa130d441148

prashanthr05:
The configure methods must be modified in view with https://github.com/dic-iit/bipedal-locomotion-framework/pull/103/files.

This means, no need to specify `BipedalLocomotion::GenericContainer::VectorResizeMode` while getting parameters from the parameter handler. 

This is why the CI is failing.

prashanthr05:
> The configure methods must be modified in view with https://github.com/dic-iit/bipedal-locomotion-framework/pull/103/files.
> 
> This means, no need to specify `BipedalLocomotion::GenericContainer::VectorResizeMode` while getting parameters from the parameter handler.
> 
> This is why the CI is failing.

Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/106/commits/4f9b30cc918792b1bd53acf43d7297eeed5ece0a

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
prashanthr05: Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/106/commits/8ac1b32c085dd24fce06f38bb718677c1b4df29c

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
prashanthr05: Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/106/commits/8ac1b32c085dd24fce06f38bb718677c1b4df29c

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
prashanthr05: Removed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/106/commits/9e6f3399f5c778a35bf36045fe8fa4cdc21b8d36

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
prashanthr05: Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/106/commits/4882805dd460666ec63b6d8d437f1d956c216a21

prashanthr05:
@GiulioRomualdi I've addressed your comments. 

I have also opened relevant issues for addressing important points left behind in this PR.

GiulioRomualdi:
Thank you @prashanthr05. If you agree I can merge the PR as it is. Let me know if you want to reorganize the commits

prashanthr05:
> Thank you @prashanthr05. If you agree I can merge the PR as it is. Let me know if you want to reorganize the commits

I've squashed some minor commits and left some other commits which modify changes from other PRs to keep the history of changes informative.

Please feel free to merge the PR once the CI passes.

GiulioRomualdi:
Merging!

prashanthr05:
cc @GiulioRomualdi an example usage of this with respect to the YARP device implementation is available in 
https://github.com/prashanthr05/bipedal-locomotion-framework/commit/f310861b9d6931ae3fdf11f354c436b508d907d9

The only difference with respect to the usage in RFModule would be the opening of the relevant devices and populating a PolyDriverList, which with a YARP device is possible to do within the configuration files and use `yarprobotinterface` attach to get a PolyDriverList.

This will be useful for your controller implementation.

PR #117, opened 2020-09-24
Fix some bugs in the python script for generating urdf files

GiulioRomualdi:
This PR fixes some bugs in `generate_stairs_urdf.py` and `GenerateAndAddUrdfModel.cmake`


Python/stairs/generate_stairs_urdf.py
```diff
@@ -16,9 +16,9 @@
 class Step:
     def __init__(self, step_index, size):
         self.name = 's_' + str(step_index)
-        self.height = size[0]
-        self.length = size[1]
-        self.width = size[2]
+        self.length = size[0]
+        self.width = size[1]
+        self.height = size[2]
```
S-Dafarra, line 21: To avoid the same problem again, how do you determined this relation? Any piece of documentation worth to be linked?

GiulioRomualdi: > To avoid the same problem again, how do you determined this relation? Any piece of documentation worth to be linked?

Actually my first idea was having a vector containing `[x y z]` dimensions that I called `[length width height]` then by mistake I confused the index.
I tried to replicate the order in the `urdf` file. I can add some documentation there

GiulioRomualdi: Here I added some documentation 016b8cacba6e301f2c0c532957d7c2a7756209c7, let me know if it's fine

S-Dafarra (approved):

PR #118, opened 2020-09-24
Bugfix in Yarp iRobotControl implementation

### LINKED ISSUE #115: Understand why the getControlModes sometimes fails
I built the first application using the `IRobotInterface` when I run the application on my laptop the application sometimes fails here:
https://github.com/dic-iit/bipedal-locomotion-framework/blob/c055c48edb86aa9f23fb7cc572a4ea005f1c626b/src/RobotInterface/YarpImplementation/src/YarpRobotControl.cpp#L131-L135

Back in time we solved the problem using a timer and checking several times if the yarp interface were able to collect the data. [See here](https://github.com/robotology/walking-controllers/blob/ba4034b65804eb311a33a96d5987ba25873f3a46/src/RobotInterface/src/Helper.cpp#L62-L115).

I would avoid implementing the same trick also here. (Probably we have the same problem with the `ISensorBridge` implemented by @prashanthr05)
 
Furthermore the fact that this behavior is completely random bothers me a lot. 

Is there someone able to explain to me why this is happening?
@traversaro @S-Dafarra @prashanthr05

GiulioRomualdi:
This PR fixes two bugs in the `yarp` implementation of the `iRobotControl` class.
- https://github.com/dic-iit/bipedal-locomotion-framework/pull/118/commits/8fc2bef8a813d078c64852ec13ca00d3a6e099c3 fixes #115 I followed what @diegoferigo implemented in [wb-toolbox](https://github.com/robotology/wb-toolbox/blob/4278d56dc16e4a76e0b4ef6b872bc3e26c43e451/toolbox/base/src/RobotInterface.cpp#L83).
- https://github.com/dic-iit/bipedal-locomotion-framework/pull/118/commits/8c17556b6a2b8bbe8a5372c0d6060860c3c7e540 resets the vectors containing the desired joint values and the associated control modes when the interface is initialized.

### DIFF
--- a/src/RobotInterface/YarpImplementation/src/YarpRobotControl.cpp
+++ b/src/RobotInterface/YarpImplementation/src/YarpRobotControl.cpp
@@ -6,6 +6,7 @@
  */
 
 #include <cmath>
+#include <thread>
 #include <unordered_map>
 
 #include <yarp/dev/IAxisInfo.h>
@@ -128,10 +129,20 @@ struct YarpRobotControl::Impl
             return false;
         }
 
-        if (!this->controlModeInterface->getControlModes(this->controlModesYarp.data()))
+        // try to read the control mode
+        unsigned counter = 0;
+        constexpr unsigned maxIter = 100;
+        constexpr unsigned timeout = 500;
+        while (!this->controlModeInterface->getControlModes(this->controlModesYarp.data()))
         {
-            std::cerr << errorPrefix << "Error reading the control mode." << std::endl;
-            return false;
+            if (++counter == maxIter)
+            {
+                std::cerr << errorPrefix << "Error while reading the control mode." << std::endl;
+                return false;
+            }
+
+            // Sleep for some while
+            std::this_thread::sleep_for(std::chrono::microseconds(timeout));
         }
 
         for (std::size_t i = 0; i < this->actuatedDOFs; i++)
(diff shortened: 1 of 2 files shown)


src/RobotInterface/YarpImplementation/src/YarpRobotControl.cpp
```diff
@@ -128,10 +129,20 @@ struct YarpRobotControl::Impl
             return false;
         }
 
-        if (!this->controlModeInterface->getControlModes(this->controlModesYarp.data()))
+        // try to read the control mode
+        unsigned counter = 0;
+        constexpr unsigned maxIter = 100;
+        constexpr unsigned timeout = 500;
```
S-Dafarra, line 135: Is it worth putting them as part of the initialization parameters?

GiulioRomualdi: We can have default values and the user can change them. Do you agree?

S-Dafarra: Absolutely

prashanthr05 (approved):
Approved considering the changes suggested by @S-Dafarra !

src/RobotInterface/YarpImplementation/src/YarpRobotControl.cpp
GiulioRomualdi: Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/118/commits/78738d2c5d8b12dd4d27669a07074a17e02073de

GiulioRomualdi:
Waiting for @S-Dafarra approval before merging

src/RobotInterface/YarpImplementation/src/YarpRobotControl.cpp
```diff
@@ -458,10 +504,41 @@ bool YarpRobotControl::initialize(std::weak_ptr<ParametersHandler::IParametersHa
         return false;
     }
 
-    bool ok = ptr->getParameter("positioning_duration", m_pimpl->positioningDuration);
-    ok = ok && ptr->getParameter("positioning_tolerance", m_pimpl->positioningTolerance);
+    // optional parameters
+    int temp = 0;
+    if (ptr->getParameter("reading_timeout", temp))
```
S-Dafarra, line 509: The problem with this is that if the ``reading_timeout`` field is missing, it will always throw an error. We may need to add a "quiet" mode for optional parameters. ANyways, this goes outside the scope of this PR.

S-Dafarra (approved):

src/RobotInterface/YarpImplementation/src/YarpRobotControl.cpp
GiulioRomualdi: > The problem with this is that if the reading_timeout field is missing, it will always throw an error. We may need to add a "quiet" mode for optional parameters. Anyways, this goes outside the scope of this PR.

Yes, this is really annoying. I opened an issue for that: https://github.com/dic-iit/bipedal-locomotion-framework/issues/120

PR #119, opened 2020-09-24
Improve performances TimeVaryingDCMPlanner

GiulioRomualdi:
This PR attempts to reduce the memory allocation (and in general improve the performances) in the `TimeVaryingDCMPlanner`, in details:
- https://github.com/dic-iit/bipedal-locomotion-framework/commit/a333d3fb631b91510d0e22ba1b21819320f6e28e Introduces the `QuinticSpline` instead of the `iDynTree::CubicSpline`, here thanks to the usage of the `Eigen::Ref<>` + `Eigen::Map<>` it was possible to directly use `casadi::DM` objects as input of the `QuinticSpline::evaluatePoint()` method. 
- https://github.com/dic-iit/bipedal-locomotion-framework/commit/0681b2d52dcc3698407ba0127be967cdc7def7be add `omega dot` as a state of the Planner.
- I noticed that building the constraints in a `for loop`  is time-consuming (for casadi), in https://github.com/dic-iit/bipedal-locomotion-framework/commit/5a4218a120f300091a900b5158a9e224e1e552bc the initialization of the constraints for `omega` and `omega dot` is now added outside a for loop.
- https://github.com/dic-iit/bipedal-locomotion-framework/commit/7540dcdd03e3c624f3095961fac6292b27cb80d9 I also add an initial guess for the `vrp`. 

I'm pretty sure we can improve the optimization of this part of the code, for instance by using the `casadi mapaccum` feature but I didn't find any documentation on how to use this feature in `c++`

CC casadi users @S-Dafarra @Giulero

### DIFF
--- a/src/Planners/src/TimeVaryingDCMPlanner.cpp
+++ b/src/Planners/src/TimeVaryingDCMPlanner.cpp
@@ -512,6 +512,7 @@ struct TimeVaryingDCMPlanner::Impl
         this->opti.set_value(this->optiParameters.dcmRefererenceTraj, initialValueDCM);
 
         this->opti.set_initial(this->optiVariables.dcm, initialValueDCM);
+        this->opti.set_initial(this->optiVariables.vrp, initialValueDCM(Sl(), Sl(0, -1)));
         this->opti.set_initial(this->optiVariables.omega, initialValueOmega);
         this->opti.set_initial(this->optiVariables.omegaDot,
                                casadi::DM::zeros(1, omegaDot.columns()));
--- a/src/Planners/src/TimeVaryingDCMPlanner.cpp
+++ b/src/Planners/src/TimeVaryingDCMPlanner.cpp
@@ -7,11 +7,9 @@
 
 #include <casadi/casadi.hpp>
 
-#include <iDynTree/Core/EigenHelpers.h>
-#include <iDynTree/Core/CubicSpline.h>
-
 #include <BipedalLocomotion/Planners/TimeVaryingDCMPlanner.h>
 #include <BipedalLocomotion/Planners/ConvexHullHelper.h>
+#include <BipedalLocomotion/Planners/QuinticSpline.h>
 #include <BipedalLocomotion/System/DynamicalSystem.h>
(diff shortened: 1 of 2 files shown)


src/Planners/src/TimeVaryingDCMPlanner.cpp
```diff
@@ -431,12 +429,11 @@ struct TimeVaryingDCMPlanner::Impl
...
                     auto contactIt = contactPhaseListIt->activeContacts.cbegin();
-                    Eigen::Vector3d p1 = contactIt->second->pose.translation();
+                    const Eigen::Vector3d p1 = contactIt->second->pose.translation();
                     std::advance(contactIt, 1);
-                    Eigen::Vector3d p2 = contactIt->second->pose.translation();
+                    const Eigen::Vector3d p2 = contactIt->second->pose.translation();
 
-                    Eigen::Vector3d desiredDCMPosition;
-                    desiredDCMPosition = (p1 + p2) / 2;
+                    Eigen::Vector3d desiredDCMPosition = (p1 + p2) / 2;
```
S-Dafarra, line 436:
```suggestion
                    Eigen::Vector3d desiredDCMPosition = (p1 + p2) / 2.0;
```

src/Planners/src/TimeVaryingDCMPlanner.cpp
```diff
@@ -449,12 +446,11 @@ struct TimeVaryingDCMPlanner::Impl
...
```
S-Dafarra, line 453:
```suggestion
                    Eigen::Vector3d desiredDCMPosition = (p1 + p2) / 2.0;
```

S-Dafarra (approved):

S-Dafarra:
> I'm pretty sure we can improve the optimization of this part of the code, for instance by using the `casadi mapaccum` feature but I didn't find any documentation on how to use this feature in `c++`

Notice anyway that ``mapaccum`` may be interesting if you want to run that constraint in multi-thread. I have never used it in C++, either, but we can discuss.

Giulero (approved):

PR #126, opened 2020-09-29
Use Casadi::Map function to compute the system dynamics in the Time varying DCM planner

GiulioRomualdi:
`Casadi::Map` is a feature implemented in casadi that allows improving the performances of the code while computing the system dynamics. 

Probably we can also increase the performances by calling the map as 
```cpp
auto trajectory = this->optiFunctions.dcmDynamics.map(horizonSampling, "thread", 8);
```
instead of
https://github.com/dic-iit/bipedal-locomotion-framework/blob/3b99825b4fdb4c9d352b1d2d27e35ccee2a59465/src/Planners/src/TimeVaryingDCMPlanner.cpp#L231

### DIFF
--- a/src/Planners/src/TimeVaryingDCMPlanner.cpp
+++ b/src/Planners/src/TimeVaryingDCMPlanner.cpp
@@ -11,40 +11,8 @@
 #include <BipedalLocomotion/Planners/ConvexHullHelper.h>
 #include <BipedalLocomotion/Planners/QuinticSpline.h>
 #include <BipedalLocomotion/System/DynamicalSystem.h>
-#include <BipedalLocomotion/System/ForwardEuler.h>
 
 using namespace BipedalLocomotion::Planners;
-using namespace BipedalLocomotion::System;
-
-class TimeVaryingDCMPlannerDynamics : public DynamicalSystem<std::tuple<casadi::MX, casadi::MX>,
-                                                             std::tuple<casadi::MX, casadi::MX>,
-                                                             std::tuple<casadi::MX, casadi::MX>>
-{
-public:
-    /**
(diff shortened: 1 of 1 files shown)


src/Planners/src/TimeVaryingDCMPlanner.cpp
```diff
@@ -230,43 +211,31 @@ struct TimeVaryingDCMPlanner::Impl
...
 
-        const auto& [initialDcmVelocity, initialOmegaVelocity] = stateDerivative;
         this->opti.subject_to(casadi::MX::vertcat({initialDcmVelocity, initialOmegaVelocity})
                               == casadi::MX::vertcat({this->optiParameters.dcmInitialVelocity, 0}));
 
         // final values
         this->opti.subject_to(omega(Sl(), -1) == this->optiParameters.omegaFinalValue);
         this->opti.subject_to(dcm(Sl(), -1) == this->optiParameters.dcmFinalPosition);
 
+        auto finalDcmVelocity = (omega(-1) - omegaDot(-1) / omega(-1)) * (dcm(Sl(), -1) - vrp(Sl(), -1));
```
S-Dafarra, line 225: Just to understand, what is the difference between ``Sl()`` and ``-1``?

GiulioRomualdi: `Sl()` takes the entire slice `-1` the last element. For instance 
`vrp(Sl(), -1))` can be seen as `vrp(:, end)`

S-Dafarra: Ah I see, thanks

S-Dafarra (approved):

PR #131, opened 2020-10-22
Propose a first implementation of the Python bindings

GiulioRomualdi:
This PR proposes a first implementation of the python bindings using pybind11. 

Among all the components of BipedalLocomotionFramework only the QuinticSpline binding has been tested. 

The main skeleton of this PR is taken from https://github.com/robotology/idyntree/pull/741.

The mapping between `eigen` and `numpy` is automatically done by `pybind11`.

This is an example of python code that can be used to compute the `QuinticSpline`
```python
import BipedalLocomotion.pybind as bl
import numpy as np

init_time = 0.32
final_time = 2.64

time = [init_time, 0.45, 1.0, 1.5, final_time]
knots = [np.random.rand(4,1) for i in range(5)]

init_velocity = np.zeros(5)
init_acceleration = np.zeros(5)

final_velocity = np.zeros(5)
final_acceleration = np.zeros(5)

spline = bl.QuinticSpline()

spline.set_knots(knots, time)
spline.set_initial_conditions(init_velocity, init_acceleration)
spline.set_final_conditions(final_velocity, final_acceleration))

position = np.ndarray(shape =(4))
velocity = np.ndarray(shape =(4))
acceleration = np.ndarray(shape =(4))

t = 0.86
spline.evaluate_point(t, position, velocity, acceleration)
```

cc @diegoferigo @traversaro @S-Dafarra @DanielePucci

### DIFF
--- a/bindings/CMakeLists.txt
+++ b/bindings/CMakeLists.txt
@@ -1,0 +1,49 @@
+# Copyright (C) 2020 Istituto Italiano di Tecnologia (IIT). All rights reserved.
+# This software may be modified and distributed under the terms of the
+# GNU Lesser General Public License v2.1 or any later version.
+
+option(FRAMEWORK_USES_PYTHON "Do you want to create the Python bindings" FALSE)
+
+if(FRAMEWORK_USES_PYTHON)
+    find_package(Python3 COMPONENTS Interpreter QUIET)
+    if (Python3_FOUND)
+        execute_process(COMMAND ${Python3_EXECUTABLE}
+            -c "from distutils import sysconfig; print(sysconfig.get_python_lib(1,0,prefix=''))"
+            OUTPUT_VARIABLE _PYTHON_INSTDIR)
+        string(STRIP ${_PYTHON_INSTDIR} FRAMEWORK_PYTHON_INSTALL_DIR)
+        # # Bindings are installed in `BipedalLocomotion` subdirectory.
+        # list(APPEND FRAMEWORK_BINARY_DIRS "${FRAMEWORK_PYTHON_INSTALL_DIR}/BipedalLocomotion")
+    endif()
+endif()
+
+# Setup installation directory for Python bindings.
+if (FRAMEWORK_USES_PYTHON)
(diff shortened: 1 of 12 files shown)


bindings/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+if(FRAMEWORK_USES_PYTHON)
+    find_package(Python3 COMPONENTS Interpreter QUIET)
+    if (Python3_FOUND)
+        execute_process(COMMAND ${Python3_EXECUTABLE}
+            -c "from distutils import sysconfig; print(sysconfig.get_python_lib(1,0,prefix=''))"
+            OUTPUT_VARIABLE _PYTHON_INSTDIR)
+        string(STRIP ${_PYTHON_INSTDIR} FRAMEWORK_PYTHON_INSTALL_DIR)
+        # # Bindings are installed in `BipedalLocomotion` subdirectory.
+        # list(APPEND FRAMEWORK_BINARY_DIRS "${FRAMEWORK_PYTHON_INSTALL_DIR}/BipedalLocomotion")
+    endif()
```
diegoferigo, line 16: Just checking. I doubt you have a virtualenv enabled in your system. Does this installs the files into `/usr` or `/usr/local`?

diegoferigo (commented):
Awesome! I would have suggested using `pytest` instead of `unittest` but this is a great start.

diegoferigo:
Nit: [PEP8](https://www.python.org/dev/peps/pep-0008/#package-and-module-names) recommends using lowercase module names, e.g. `bipedal_locomotion`. If you don't think to add any alternative bindings framework (iDynTree currently has both pybind and SWIG), I suggest to modify the structure to use `import bipedal_locomotion.bindings`.

GiulioRomualdi:
Related discussion regarding a possible solution to automatize the generation of the bindings with `pybind11`: https://github.com/RobotLocomotion/drake/issues/7889

GiulioRomualdi:
We can close this PR since @diegoferigo proposed a more complete implementation in #134

PR #155, opened 2020-11-27
devices: Allow dynamic installation of plugin ini files

prashanthr05:
Replace manually written ini files with YARP dynamic plugin ini installation feature.

PR #156, opened 2020-11-27
Implement a YARP based ROS publisher class in YarpUtilities library.

prashanthr05:
This PR adds a class for publishing wrenches, joint states and transforms over ROS topics using YARP only software.
Inspiration was taken from [human-dynamics-estimation/publishers](https://github.com/robotology/human-dynamics-estimation/tree/master/publishers).

The class internally contains a YARP based ROS node and a set of publishers which can be configured for publishing over certain topics and configuring connections to a YARP transform server in order to publish transforms. 

Although the class might be ROS independent, in order to run the code, ROS is required and usual YARP-ROS connections need to be made.

A test device has also been implemented, I can include it in this PR if required.

This PR depends on  https://github.com/dic-iit/bipedal-locomotion-framework/pull/155 if the test device needs to be added.

### DIFF
--- a/src/YarpUtilities/include/BipedalLocomotion/YarpUtilities/RosPublisher.h
+++ b/src/YarpUtilities/include/BipedalLocomotion/YarpUtilities/RosPublisher.h
@@ -54,8 +54,8 @@ public:
      * @param[in] jointList list of joints
      * @param[in] jointPositions list of joint positions in m or rad with the same size as the joints list
      */
-    bool publishJointStates(const BipedalLocomotion::GenericContainer::Vector<std::string>& jointList,
-                            const BipedalLocomotion::GenericContainer::Vector<double>& jointPositions);
+    bool publishJointStates(const BipedalLocomotion::GenericContainer::Vector<const std::string>::Ref jointList,
+                            const BipedalLocomotion::GenericContainer::Vector<const double>::Ref jointPositions);
     
     /**
      * Publish the joint states over the configured joint states topic 
@@ -64,9 +64,9 @@ public:
      * @param[in] jointPositions list of joint positions in m or rad with the same size as the joints list
      * @param[in] jointVelocities list of joint velocities in m/s or rad/s with the same size as the joints list
      */
-    bool publishJointStates(const BipedalLocomotion::GenericContainer::Vector<std::string>& jointList,
-                            const BipedalLocomotion::GenericContainer::Vector<double>& jointPositions,
-                            const BipedalLocomotion::GenericContainer::Vector<double>& jointVelocities);
+    bool publishJointStates(const BipedalLocomotion::GenericContainer::Vector<const std::string>::Ref jointList,
+                            const BipedalLocomotion::GenericContainer::Vector<const double>::Ref jointPositions,
+                            const BipedalLocomotion::GenericContainer::Vector<const double>::Ref jointVelocities);
     
     /**
      * Publish the joint states over the configured joint states topic 
@@ -76,10 +76,10 @@ public:
      * @param[in] jointVelocities list of joint velocities in m/s or rad/s with the same size as the joints list
      * @param[in] jointPositions list of joint torques/forces in Nm or N with the same size as the joints list
      */
-    bool publishJointStates(const BipedalLocomotion::GenericContainer::Vector<std::string>& jointList,
-                            const BipedalLocomotion::GenericContainer::Vector<double>& jointPositions,
-                            const BipedalLocomotion::GenericContainer::Vector<double>& jointVelocities,
-                            const BipedalLocomotion::GenericContainer::Vector<double>& jointEfforts);
+    bool publishJointStates(const BipedalLocomotion::GenericContainer::Vector<const std::string>::Ref jointList,
+                            const BipedalLocomotion::GenericContainer::Vector<const double>::Ref jointPositions,
+                            const BipedalLocomotion::GenericContainer::Vector<const double>::Ref jointVelocities,
+                            const BipedalLocomotion::GenericContainer::Vector<const double>::Ref jointEfforts);
     
     /**
      * Publish the wrenches over the configured wrench topic associated to the frame
@@ -87,7 +87,7 @@ public:
      * @param[in] wrench6d 6d wrench as force-torque in N-Nm
      */
     bool publishWrench(const std::string& frameName, 
-                       BipedalLocomotion::GenericContainer::Vector<double>& wrench6d);
+                       BipedalLocomotion::GenericContainer::Vector<const double>::Ref wrench6d);
     
     /**
      * Publish transforms to the transform server
@@ -96,7 +96,7 @@ public:
      * @param[in] transformAsVector16d 4x4 transform as a vector data
      */
     bool publishTransform(const std::string& target, const std::string& source, 
-                          const BipedalLocomotion::GenericContainer::Vector<double>& transformAsVector16d);
+                          const BipedalLocomotion::GenericContainer::Vector<const double>::Ref transformAsVector16d);
     
     /**
      * configure which transform server to connect to
--- a/src/YarpUtilities/include/BipedalLocomotion/YarpUtilities/RosPublisher.h
+++ b/src/YarpUtilities/include/BipedalLocomotion/YarpUtilities/RosPublisher.h
@@ -22,6 +22,14 @@ namespace BipedalLocomotion
 namespace YarpUtilities
 {
 
+/**
+ * The class internally contains a YARP based ROS node and a set of publishers.
+ * Current implementation consists of 
+ *  - Joint states publisher
+ *  - Wrenches publisher
+ *  - Transform broadcaster
+ * Although the class might be ROS independent, in order to run the code, ROS is required and usual YARP-ROS connections need to be made.
+ */
 class RosPublisher 
 {
 public:
@@ -38,6 +46,12 @@ public:
     
     /**
      * Configures the publishers and ROS topics for publishing messages
+     * The parameters used to configure the RosPublisher,
+     * - "joint_states_topic" name of the topic over which the joint states need to be published
+     * - "transform_server_port" name of the already running transform server's remote port
+     * - "WrenchPublishers" is group with the following parameters,
+     *      - "frame_names" a list containing the frames at which the published wrenches should be expressed
+     *      - "topics" a list containing the topics over which the wrenches need to be published. Must be the same size and order as frame_names.
      * @param[in] handler Parameter handler
      * @note this method needs to be called after construction and before publishing  
      */
--- a/src/YarpUtilities/include/BipedalLocomotion/YarpUtilities/RosPublisher.h
+++ b/src/YarpUtilities/include/BipedalLocomotion/YarpUtilities/RosPublisher.h
@@ -1,0 +1,140 @@
+/**
+ * @file RosPublisher.h
+ * @authors Prashanth Ramadoss
+ * @copyright 2019 Istituto Italiano di Tecnologia (IIT). This software may be modified and
+ * distributed under the terms of the GNU Lesser General Public License v2.1 or any later version.
+ */
+
+#ifndef BIPEDAL_LOCOMOTION_YARP_UTILITIES_ROS_PUBLISHER_H
+#define BIPEDAL_LOCOMOTION_YARP_UTILITIES_ROS_PUBLISHER_H
+
+#include <BipedalLocomotion/ParametersHandler/IParametersHandler.h>
+#include <BipedalLocomotion/GenericContainer/Vector.h>
+#include <memory>
+#include <iostream>
+#include <string>
+namespace BipedalLocomotion
+{
+
+/**
+ * Helper for YARP library.
+ */
+namespace YarpUtilities
+{
+
+class RosPublisher 
+{
+public:
+    /**
+     * Constructor
+     * @param[in] nodeName Name of the ROS publisher node 
+     */
+    RosPublisher(const std::string& nodeName);
+    
+    /**
+     * Destructor     
+     */
+    ~RosPublisher();
+    
+    /**
+     * Configures the publishers and ROS topics for publishing messages
+     * @param[in] handler Parameter handler
+     * @note this method needs to be called after construction and before publishing  
+     */
+    bool initialize(std::weak_ptr<BipedalLocomotion::ParametersHandler::IParametersHandler> handler);
+    
+    /**
+     * Clear internal configuration and close the node
+     */
+    void stop();
+        
+    /**
+     * Publish the joint states over the configured joint states topic 
+     * The joint velocities and joint efforts are set to zero.
+     * @param[in] jointList list of joints
+     * @param[in] jointPositions list of joint positions in m or rad with the same size as the joints list
+     */
+    bool publishJointStates(const BipedalLocomotion::GenericContainer::Vector<std::string>& jointList,
+                            const BipedalLocomotion::GenericContainer::Vector<double>& jointPositions);
+    
+    /**
+     * Publish the joint states over the configured joint states topic 
+     * The joint velocities and joint efforts are set to zero.
+     * @param[in] jointList list of joints
+     * @param[in] jointPositions list of joint positions in m or rad with the same size as the joints list
+     * @param[in] jointVelocities list of joint velocities in m/s or rad/s with the same size as the joints list
+     */
+    bool publishJointStates(const BipedalLocomotion::GenericContainer::Vector<std::string>& jointList,
+                            const BipedalLocomotion::GenericContainer::Vector<double>& jointPositions,
+                            const BipedalLocomotion::GenericContainer::Vector<double>& jointVelocities);
+    
+    /**
+     * Publish the joint states over the configured joint states topic 
+     * The joint velocities and joint efforts are set to zero.
+     * @param[in] jointList list of joints
+     * @param[in] jointPositions list of joint positions in m or rad with the same size as the joints list
+     * @param[in] jointVelocities list of joint velocities in m/s or rad/s with the same size as the joints list
+     * @param[in] jointPositions list of joint torques/forces in Nm or N with the same size as the joints list
+     */
+    bool publishJointStates(const BipedalLocomotion::GenericContainer::Vector<std::string>& jointList,
+                            const BipedalLocomotion::GenericContainer::Vector<double>& jointPositions,
+                            const BipedalLocomotion::GenericContainer::Vector<double>& jointVelocities,
+                            const BipedalLocomotion::GenericContainer::Vector<double>& jointEfforts);
+    
+    /**
+     * Publish the wrenches over the configured wrench topic associated to the frame
+     * @param[in] frameName frame at which the wrench will be expressed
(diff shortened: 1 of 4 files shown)


prashanthr05:
cc @Yeshasvitvs

prashanthr05:
An example usage of the configuration for the RosPublisher class in YARP device format is as follows,
This type of initialization was chosen thanks to @Yeshasvitvs advice, to be extensible to most ROS topic scenarios.

``` cpp
<?xml version="1.0" encoding="UTF-8" ?>
<device  xmlns:xi="http://www.w3.org/2001/XInclude" name="ros-publisher-test" type="ROSPublisherTestDevice">
    <param name="joint_states_topic">/joint_states</param>
    <param name="transform_server_port">/tfServer</param>
     <group name="WrenchPublishers">
        <param name="frame_names">("right", "left")</param> 
        <param name="topics">("/wrench/right", "/wrench/left")</param>
    </group>
</device>
```

src/YarpUtilities/include/BipedalLocomotion/YarpUtilities/RosPublisher.h
```diff
@@ -0,0 +1,140 @@
...
+     */
+    void stop();
+        
+    /**
+     * Publish the joint states over the configured joint states topic 
+     * The joint velocities and joint efforts are set to zero.
+     * @param[in] jointList list of joints
+     * @param[in] jointPositions list of joint positions in m or rad with the same size as the joints list
+     */
+    bool publishJointStates(const BipedalLocomotion::GenericContainer::Vector<std::string>& jointList,
```
S-Dafarra, line 57: Can you try to substitute 
```
const BipedalLocomotion::GenericContainer::Vector<something>&
```
with 
```
const BipedalLocomotion::GenericContainer::Vector<something>::Ref
```
?
This would simplify the usage, as you don't need to use ``make_vector`` before, and you can directly input your original vector. Check https://github.com/dic-iit/bipedal-locomotion-framework/pull/102 and https://github.com/dic-iit/bipedal-locomotion-framework/pull/103.

This applies also everywhere else.

prashanthr05: Thanks for the suggestion. I knew something was missing there but couldn't point to this.

S-Dafarra: I just updated the comment above for a clarification about ``const``.

prashanthr05: Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/156/commits/0df2ac4e136da996196f4bf12446664b4c0dd9b7

DanielePucci:
CC @dic-iit/dynamic-interaction-control

prashanthr05:
> but the transform servers don't seem to be working as expected. I am debugging. (The code seems to be correct, though).

Seems to be working fine now. Looks like it was a local YARP configuration problem.

Screenshot from 2020-12-01 15-16-21

src/YarpUtilities/include/BipedalLocomotion/YarpUtilities/RosPublisher.h
```diff
@@ -0,0 +1,140 @@
...
+     * @param[in] wrench6d 6d wrench as force-torque in N-Nm
+     */
+    bool publishWrench(const std::string& frameName, 
+                       BipedalLocomotion::GenericContainer::Vector<const double>::Ref wrench6d);
+    
+    /**
+     * Publish transforms to the transform server
+     * @param[in] target  target frame for the transform
+     * @param[in] source source frame of the transform
+     * @param[in] transformAsVector16d 4x4 transform as a vector data
```
S-Dafarra, line 96: Row major or column major?

src/YarpUtilities/src/RosPublisher.cpp
```diff
@@ -0,0 +1,482 @@
...
+    bool setupParamV(std::weak_ptr<BipedalLocomotion::ParametersHandler::IParametersHandler> handler,
+                     const std::string& param, std::vector<Scalar>& vec, const std::string& prefix)
+    {
+        auto handle = handler.lock();
+        if (handle == nullptr)
+        {
+            return false;
+        }
+
+        if (!handle->getParameter(param, GenericContainer::make_vector(vec, GenericContainer::VectorResizeMode::Resizable)))
```
S-Dafarra, line 80: By default,
```
getParameter(param, vec)
```
should consider ``vec`` as resizable. The syntax you used is useful when you don't want the vector to be considered resizable.

src/YarpUtilities/include/BipedalLocomotion/YarpUtilities/RosPublisher.h
```diff
@@ -0,0 +1,140 @@
...
+     * @param[in] frameName frame name
+     * @param[in] topicName topic name      
+     */
+    bool configureWrenchPublisher(const std::string& frameName, 
+                                  const std::string& topicName);
+    
+    /**
+     * remove wrench publisher
+     * @param[in] frameName frame name 
+     */
```
S-Dafarra, line 124: It is really just my OCD speaking, but please use the first capital letter in the documentation.

src/YarpUtilities/src/RosPublisher.cpp
```diff
@@ -0,0 +1,482 @@
...
+}
+
+bool RosPublisher::configureJointStatePublisher(const std::string& topicName)
+{
+    if (m_pimpl->jointStatePublisher.ptr == nullptr)
+    {
+        m_pimpl->jointStatePublisher.ptr = std::make_unique< yarp::os::Publisher<JointStateMsg> >();
+    }
+    
+    m_pimpl->jointStatePublisher.ptr->close();
```
S-Dafarra, line 195: I have noticed you used this logic a few times. Consider using proper constructors/destructors in order to properly allocate and close these objects automatically.

src/YarpUtilities/src/RosPublisher.cpp
```diff
@@ -0,0 +1,482 @@
...
+    }
+
+    return true;
+}
+
+bool RosPublisher::publishJointStates(const GenericContainer::Vector<const std::string>::Ref jointList,
+                                      const GenericContainer::Vector<const double>::Ref jointPositions)
+{        
+    std::vector<double> jointVelocities(jointList.size(), 0.0);
+    std::vector<double> jointEfforts(jointList.size(), 0.0);
```
S-Dafarra, line 323: Actually, here you don't need two separate vectors. In addition, consider having a dummy buffer of the correct size as class method in order to avoid unnecessary memory allocation.

src/YarpUtilities/src/RosPublisher.cpp
```diff
@@ -0,0 +1,482 @@
...
+    pub.msg.header.seq++;
+    pub.msg.header.frame_id = frame;
+    pub.msg.header.stamp = m_pimpl->getTimeStampFromYarp();
+    pub.msg.wrench.force.x = wrench6d(0);
+    pub.msg.wrench.force.y = wrench6d(1);
+    pub.msg.wrench.force.z = wrench6d(2);
+    
+    pub.msg.wrench.torque.x = wrench6d(3);
+    pub.msg.wrench.torque.y = wrench6d(4);
+    pub.msg.wrench.torque.z = wrench6d(5);
```
S-Dafarra, line 424: Maybe it is worth writing somewhere this expected serialization.

src/YarpUtilities/src/RosPublisher.cpp
```diff
@@ -0,0 +1,482 @@
...
+
+yarp::rosmsg::TickTime RosPublisher::Impl::getTimeStampFromYarp()
+{
+    std::string_view printPrefix = "[RosPublisher::Impl::getTimeStampFromYarp] ";
+    yarp::rosmsg::TickTime rosTickTime;
+    double yarpTimeStamp = yarp::os::Time::now();
+
+    uint64_t time = static_cast<uint64_t>(yarpTimeStamp * 1000000000UL);
+    uint64_t nsec_part = time % 1000000000UL;
+    uint64_t sec_part = time / 1000000000UL;
```
S-Dafarra, line 438: Can 't you cast directly ``yarpTimeStamp`` instead of multiplying it first times 1000000000UL?

src/YarpUtilities/src/RosPublisher.cpp
prashanthr05: I tried it actually. Don't remember why now, the issue I faced with that. I think I will give it one more try.

src/YarpUtilities/src/RosPublisher.cpp
prashanthr05: Actually this function is a copy-paste from `hde`. :D  
Although I remember some issues in the past with ROS timestamps specially while handling `/tf`, if we are not careful enough it throws a timestamp extrapolation error.

Maybe @Yeshasvitvs @lrapetti  could add some light on this?

Yeshasvitvs: > Actually this function is a copy-paste from

@prashanthr05 I guess you got it from https://github.com/robotology/human-dynamics-estimation/blame/master/publishers/HumanStatePublisher/HumanStatePublisher.cpp#L517-L536

This part was added by @diegoferigo at the initial stage of HDE v2, may be he can add something to the conversation. However, I notice that it is similar to the idl autogenerated code for `TickTime.h` https://github.com/robotology/yarp/blob/514b62f0f3575437cd86001d113773cb16372727/src/libYARP_rosmsg/src/idl_generated_code/yarp/rosmsg/TickTime.h#L45-L53

CC @lrapetti

diegoferigo: I have no memory of the logic of this code after 2y but I guess that the conversion has to deal with the fact the yarp time is a double and ros time is a pair of unsigned (sec, nsec), and the conversion should maximize the range while preventing overflow.

I think that today you can do much better using chrono and converting first the double to ns and then from it extracting (sec, nsec).

S-Dafarra: Thanks @Yeshasvitvs, @diegoferigo, after reading it a couple of times I realized why those operations were necessary. It is also true that if it required several people to guess what is going on, maybe it is worth a bit of refractory. Indeed, the hint of ``std::chrono`` may be a good move, https://en.cppreference.com/w/cpp/chrono/duration

prashanthr05: Changed to use `chrono` in https://github.com/dic-iit/bipedal-locomotion-framework/pull/156/commits/d57a286f4780f87c335ff0b3b91ca01443ffa069

src/YarpUtilities/src/RosPublisher.cpp
prashanthr05: Addressed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/156/commits/d57a286f4780f87c335ff0b3b91ca01443ffa069

src/YarpUtilities/src/RosPublisher.cpp
prashanthr05: Addressed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/156/commits/d57a286f4780f87c335ff0b3b91ca01443ffa069

src/YarpUtilities/src/RosPublisher.cpp
prashanthr05: Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/156/commits/d57a286f4780f87c335ff0b3b91ca01443ffa069

src/YarpUtilities/include/BipedalLocomotion/YarpUtilities/RosPublisher.h
prashanthr05: Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/156/commits/d57a286f4780f87c335ff0b3b91ca01443ffa069

src/YarpUtilities/src/RosPublisher.cpp
prashanthr05: Addressed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/156/commits/d57a286f4780f87c335ff0b3b91ca01443ffa069

src/YarpUtilities/include/BipedalLocomotion/YarpUtilities/RosPublisher.h
prashanthr05: Row major, added documentation in https://github.com/dic-iit/bipedal-locomotion-framework/pull/156/commits/d57a286f4780f87c335ff0b3b91ca01443ffa069

src/YarpUtilities/src/RosPublisher.cpp
```diff
@@ -96,13 +111,21 @@ class RosPublisher::Impl
     
     bool publishTF{false}; /**< flag to enable publishing transforms to YARP transform server */    
     bool initialized{false}; /**< flag to chekc if the publisher was initialized */
+    
+    // placeholder variables
+    yarp::sig::Matrix pose; /**< placeholder variable for publishing transform data*/
+    std::vector<double> jointStateZero; /**< placeholder variable for zero joint state data*/
+    
+    std::chrono::nanoseconds timeInNanoSeconds; /**< chrono nanoseconds conversion from YARP to ROS timestamp*/
```
prashanthr05, line 119: This is unused, I will remove it.

prashanthr05: Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/156/commits/1f927f2d77a8f4cb629753f29fe92cc4861af96d

src/YarpUtilities/src/RosPublisher.cpp
```diff
@@ -0,0 +1,496 @@
...
+}
+
+yarp::rosmsg::TickTime RosPublisher::Impl::getTimeStampFromYarp()
+{
+    std::string_view printPrefix = "[RosPublisher::Impl::getTimeStampFromYarp] ";
+    yarp::rosmsg::TickTime rosTickTime;
+    
+    std::chrono::duration<double> timeStamp(yarp::os::Time::now());
+    uint64_t nsec_part = std::chrono::duration_cast<std::chrono::nanoseconds>(timeStamp).count() % 1000000000UL;
+    uint64_t sec_part = std::chrono::duration_cast<std::chrono::seconds>(timeStamp).count();
```
S-Dafarra, line 452: If I understood well, the actual time should be 
```
timeStamp = sec_part + nsec_part/1e-9
```
so why not exploiting chrono also for that? You could subtract the number of seconds from the initial time and then convert it to nanoseconds
```    
uint64_t sec_part  = std::chrono::duration_cast<std::chrono::seconds>(timeStamp).count();
auto nanosec_chrono =   std::chrono::duration_cast<std::chrono::seconds>(timeStamp - sec_part);
uint64_t nsec_part = std::chrono::duration_cast<std::chrono::nanoseconds>(nanosec_chrono).count();
```
I am not sure if this works, neither if it is correct, but I would like to avoid the appearance of weird 1000000000UL that are super error-prone.

prashanthr05: > but I would like to avoid the appearance of weird 1000000000UL that are super error-prone.

Agreed. I have fixed the timestamp processing in https://github.com/dic-iit/bipedal-locomotion-framework/pull/156/commits/5a4c6b5e8badb124eedeb987bad1a5202a0a3e97 to avoid magic numbers.

S-Dafarra:
Just to understand, do you have any test we can keep?

prashanthr05:
> Just to understand, do you have any test we can keep?

I have a test device in https://github.com/prashanthr05/bipedal-locomotion-framework/tree/device/tes-ros-publisher/devices/ROSPublisherTestDevice that I used to verify the working of this class. 

I used that approach because it was on top of my mind at the time of implementation. But I suppose, it can be converted to a standard C++ application as well.

S-Dafarra (approved):

S-Dafarra:
> > Just to understand, do you have any test we can keep?
> 
> I have a test device in https://github.com/prashanthr05/bipedal-locomotion-framework/tree/device/tes-ros-publisher/devices/ROSPublisherTestDevice that I used to verify the working of this class.
> 
> I used that approach because it was on top of my mind at the time of implementation. But I suppose, it can be converted to a standard C++ application as well.

Yeah, it would good to have some sort of testing, eventually using also bash scripts, but this can also go in a separate PR.

prashanthr05:
> > > Just to understand, do you have any test we can keep?
> > 
> > 
> > I have a test device in https://github.com/prashanthr05/bipedal-locomotion-framework/tree/device/tes-ros-publisher/devices/ROSPublisherTestDevice that I used to verify the working of this class.
> > I used that approach because it was on top of my mind at the time of implementation. But I suppose, it can be converted to a standard C++ application as well.
> 
> Yeah, it would good to have some sort of testing, eventually using also bash scripts, but this can also go in a separate PR.

I can open a new PR with the device, runner application(for running the prerequisites), and the commands to check the expected outputs (for the moment)

S-Dafarra:
> > > > Just to understand, do you have any test we can keep?
> > > 
> > > 
> > > I have a test device in https://github.com/prashanthr05/bipedal-locomotion-framework/tree/device/tes-ros-publisher/devices/ROSPublisherTestDevice that I used to verify the working of this class.
> > > I used that approach because it was on top of my mind at the time of implementation. But I suppose, it can be converted to a standard C++ application as well.
> > 
> > 
> > Yeah, it would good to have some sort of testing, eventually using also bash scripts, but this can also go in a separate PR.
> 
> I can open a new PR with the device, runner yarpmanager application(for running the prerequisites), and the commands to check the expected outputs. Eventually we can convert it to a test format we like.

Sure! As @traversaro mentioned, "if you like a feature, put a test on it" :grin:

prashanthr05:
> > > > > Just to understand, do you have any test we can keep?
> > > > 
> > > > 
> > > > I have a test device in https://github.com/prashanthr05/bipedal-locomotion-framework/tree/device/tes-ros-publisher/devices/ROSPublisherTestDevice that I used to verify the working of this class.
> > > > I used that approach because it was on top of my mind at the time of implementation. But I suppose, it can be converted to a standard C++ application as well.
> > > 
> > > 
> > > Yeah, it would good to have some sort of testing, eventually using also bash scripts, but this can also go in a separate PR.
> > 
> > 
> > I can open a new PR with the device, runner yarpmanager application(for running the prerequisites), and the commands to check the expected outputs. Eventually we can convert it to a test format we like.
> 
> Sure! As @traversaro mentioned, "if you like a feature, put a test on it"

A test device has been added in https://github.com/dic-iit/bipedal-locomotion-framework/pull/160.

prashanthr05:
@GiulioRomualdi if you have any comments for this PR, let me know. 
@S-Dafarra otherwise, are we good for the merge?

S-Dafarra:
> @GiulioRomualdi if you have any comments for this PR, let me know.
> @S-Dafarra otherwise, are we good for the merge?

Ok for me

GiulioRomualdi (approved):

GiulioRomualdi:
Merging! Thank you @prashanthr05

PR #157, opened 2020-11-27
Implement Continuous algebraic Riccati equation function

GiulioRomualdi:
This PR implements an algorithm to solve the Continuous algebraic Riccati equation.

The main idea has been taken from [`drake`](https://github.com/RobotLocomotion/drake/blob/master/math/continuous_algebraic_riccati_equation.cc).

cc @Giulero


src/Math/include/BipedalLocomotion/Math/AlgebraicRiccatiEquation.h
```diff
@@ -0,0 +1,50 @@
...
+ * @param tolerance max admissible error
+ * @param maxIterations max number of iterations
+ * @param verbosity level
+ * @return a pair. If the Function is able to find a solution the first term is true and the second
+ * one contains the solution. Otherwise the first element of the pair is false.
+ * @note This function implements the algorithm presented in the paper: Solving the algebraic
+ * Riccati equation with the matrix sign function
+ * https://www.sciencedirect.com/science/article/pii/0024379587902229
+ */
+std::pair<bool, Eigen::MatrixXd>
```
S-Dafarra, line 38: It is a pity to return a matrix, especially considering that this method can be called in controllers that need to run fast. I would suggest making it an input and return a simple boolean.

src/Math/src/AlgebraicRiccatiEquation.cpp
```diff
@@ -0,0 +1,129 @@
...
+        result.first = false;
+        return result;
+    }
+
+    Eigen::MatrixXd H(2 * n, 2 * n);
+    H << A, B * R_cholesky.solve(B.transpose()),
+         Q, -A.transpose();
+
+    Eigen::MatrixXd Z = H;
+    Eigen::MatrixXd Z_old;
```
S-Dafarra, line 80: I would prefer solutions that minimize the chance of memory allocation. Consider turning this as parameters of a class. In case you want to keep the function-like approach, you can still define a static function, but at the point is the user that decides to not care about memory allocation.

src/Math/src/AlgebraicRiccatiEquation.cpp
```diff
@@ -0,0 +1,129 @@
...
+        return result;
+    }
+
+    constexpr double symmetricTolerance = 1e-8;
+    if (!Q.isApprox(Q.transpose(), symmetricTolerance))
+    {
+        std::cerr << errorPrefix << "The matrix Q must be symmetric." << std::endl;
+        result.first = false;
+        return result;
+    }
```
prashanthr05, line 73: This could possibly be an utility function which could be used in other instances for debugging (for example also in estimators library). What do you think?

prashanthr05: `symmetricTolerance` parameter could be a generic defnition across the framework.

S-Dafarra: Also, it may be time-consuming. Consider having a way to skip that, like putting it into an assert.

src/Math/src/AlgebraicRiccatiEquation.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+    Eigen::MatrixXd H(2 * n, 2 * n);
+    H << A, B * R_cholesky.solve(B.transpose()),
+         Q, -A.transpose();
```
prashanthr05, line 77: Wikipedia uses $-Q$ instead of $Q$ and $-B R^{-1} B^T$ instead of positive  https://en.wikipedia.org/wiki/Algebraic_Riccati_equation

Could you confirm that with a clarification with the original paper?

src/Math/src/AlgebraicRiccatiEquation.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+    for (; iteration < maxIterations && relativeNorm >= tolerance;
+         iteration++)
+    {
+        Z_old = Z;
+        // R. Byers. Solving the algebraic Riccati equation with the matrix sign
+        // function. Linear Algebra Appl., 85:267–279, 1987
+        // Added determinant scaling to improve convergence (converges in rough half
+        // the iterations with this)
+        const double ck = std::pow(std::abs(Z.determinant()), -1.0 / p);
```
prashanthr05, line 94:
```suggestion
        const double ck = std::pow(std::abs(Z.determinant()), -1.0 / (2*p));
```

Looks like that in the original paper?

prashanthr05: I think the above comment can be safely ignored. For the implemented one is the good, see Eqn 2. of the paper.

src/Math/src/AlgebraicRiccatiEquation.cpp
prashanthr05: Also, could you please write this in matrix form as a comment? just to understand which submatrix goes where?

src/Math/include/BipedalLocomotion/Math/AlgebraicRiccatiEquation.h
GiulioRomualdi: I would keep both the implementation. For instance, in an LQR this matrix is computed only once

S-Dafarra: Well, in an infinite-time LQR is computed only once :wink: In any case, I agree, it is possible to have both versions. As mentioned [below](https://github.com/dic-iit/bipedal-locomotion-framework/pull/157#discussion_r531680512) you could define a static method and/or exploit overloading

GiulioRomualdi:
hI, @S-Dafarra  I was trying to transform the function into a class but I've several doubts. I was thinking to organize it as 
```cpp
bool initialize(std::weak_ptr<ParametersHandler::IParametersHandler> handler);

bool setMatrices(Eigen::Ref<const Eigen::MatrixXd> A,
                     Eigen::Ref<const Eigen::MatrixXd> B,
                     Eigen::Ref<const Eigen::MatrixXd> Q,
                     Eigen::Ref<const Eigen::MatrixXd> R);

Solution getSolution();

bool getSolution(Eigen::Ref<Eigen::MatrixXd> S);
```

however, I don't know how to split the code. For instance, in `setMatrices` I can just copy the matrices `A`, `B` `Q` and `R` and I can also create the hamiltonian matrix `H`. 
On the other hand in `getSolution()` I can run the algorithm to compute `S`. However, given a set of `A` `B` `Q` and `R` only one solution exists so it's not efficient to compute the solution in `getSolution()`

For instance this code will run the algorithm 4 times. 
```cpp
CARE care();
care.setMatrices(A,B,Q,R);

CARE::Solution sol = care.getSolution();
sol = care.getSolution();
sol = care.getSolution();
sol = care.getSolution();
sol = care.getSolution();
```

So the question is: shall we compute the solution while setting `A`, `B` `Q` and `R` and store it as a parameter of the class? In this case, shall we add the following matrices as a parameter of the class (notice that the size of  `H` is known only after `setMatrices` is called)?
```cpp
Eigen::MatrixXd Z = H;
Eigen::MatrixXd Z_old;
```
Indeed every time we set new matrices `H` has to be reset

S-Dafarra:
I think you can add a ``solve`` method that does the computation. This may also just output a boolean in case of success/failure. Then, the entire pipeline ``init`` -> ``set`` -> ``solve`` -> ``get`` could be handled by a single static function, in case you don't care about memory allocation.

GiulioRomualdi:
I like it

GiulioRomualdi:
Hi @S-Dafarra and @prashanthr05 I completely restructure the code. Let me know what you think

src/Math/src/CARE.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+    void resizeInteralMatrices()
+    {
+        const std::size_t n = B.rows();
+
+        // resize the Hamiltonian matrix
+        Z.resize(2 * n, 2 * n);
+        ZOld.resize(2 * n, 2 * n);
+
+        identity = Eigen::MatrixXd::Identity(n, n);
```
S-Dafarra, line 42: Here I would use a ``resize`` and a ``setIdentity``. In this way, if the size is already the correct one, it avoids memory allocations.

src/Math/src/CARE.cpp
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    // Z represents the Hamiltonian matrix.
+    //       _                _
+    //      | A   B inv(R) B'  |
+    // Z  = |                  |
+    //      |_Q       -  A'   _|
+
+    m_pimpl->Z << m_pimpl->A, m_pimpl->B * R_cholesky.solve(m_pimpl->B.transpose()),
+                  m_pimpl->Q, -m_pimpl->A.transpose();
```
S-Dafarra, line 136: I would suggest the use of intermediate variables to avoid the generation of temporary objects, especially for 
```
m_pimpl->B * R_cholesky.solve(m_pimpl->B.transpose()),
```
Also, for readability and for catching early mistakes, I would prefer assigning the blocks.

src/Math/src/CARE.cpp
```diff
@@ -0,0 +1,191 @@
...
+    {
+        m_pimpl->ZOld = m_pimpl->Z;
+
+        // R. Byers. Solving the algebraic Riccati equation with the matrix sign
+        // function. Linear Algebra Appl., 85:267–279, 1987
+        // Added determinant scaling to improve convergence (converges in rough half
+        // the iterations with this)
+        const double ck = std::pow(std::abs(m_pimpl->Z.determinant()), -1.0 / p);
+        m_pimpl->Z *= ck;
+        m_pimpl->Z = m_pimpl->Z - 0.5 * (m_pimpl->Z - m_pimpl->Z.inverse());
```
S-Dafarra, line 153: The same as above here. Almost surely, ``- m_pimpl->Z.inverse()`` generates a temporary object.

prashanthr05 (approved):
I also clarified the previous doubt of the Hamiltonian matrix's subblock sign differences from Wikipedia. Turns out the choice used here is also a Hamiltonian matrix because it respects the identity $J Z = (J Z)^T$ where $J = \begin{bmatrix} 0_n & I_n \\\ - I_n & 0_n \end{bmatrix}$.
(the requirement of off-diagonal elements being symmetric and the sum of diagonal elements equal to zero is the property).



For me the changes look neat. However, I will also incorporate @S-Dafarra suggestions.

src/Math/include/BipedalLocomotion/Math/CARE.h
```diff
@@ -0,0 +1,94 @@
...
+     */
+    struct Impl;
+    std::unique_ptr<Impl> m_pimpl;
+
+public:
+
+    /**
+     * Initialize the continuous algebraic riccati equation solver.
+     * @param handler pointer to the parameter handler.
+     * @note The following parameters are required by the class:
```
prashanthr05, line 44: this note needs to be changed since the parameters are optional, rather, the method itself is optional.

src/Math/src/CARE.cpp
GiulioRomualdi: It should be done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/157/commits/b3209960b6d9c6109d2ee92b2dad4638f74e5ae5

src/Math/src/CARE.cpp
GiulioRomualdi: It should be done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/157/commits/b3209960b6d9c6109d2ee92b2dad4638f74e5ae5

src/Math/src/CARE.cpp
GiulioRomualdi: For the  `<<` please check https://github.com/dic-iit/bipedal-locomotion-framework/pull/157/commits/40b7b4cf3d0aca65b62f67c138ec3578617758d5

src/Math/src/CARE.cpp
GiulioRomualdi: It should be done in b320996

src/Math/include/BipedalLocomotion/Math/CARE.h
GiulioRomualdi: Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/commit/3c6176db6cd43d8d0b98d3c0c2deec7c928e3393

GiulioRomualdi:
@S-Dafarra let me know if you're fine with the modifications

S-Dafarra (approved):

S-Dafarra:
> @S-Dafarra let me know if you're fine with the modifications

Yes, feel free to merge as soon as the CI passess.

PR #204, opened 2021-02-23
 C++ refactoring of Contact

### LINKED ISSUE #162: Migrate Contact library outside Planners folder
In https://github.com/dic-iit/bipedal-locomotion-framework/pull/142, we identified the `Contact` class to have a more generic usage extending beyond just planning. 
So we abstracted the current Contact class into a `ContactBase` class and implemented derived classes `PlannedContact` and `EstimatedContact` classes.

In order to not break the current workflows and python bindings related merges, I created a new library `Contacts` and added only the above mentioned classes (See https://github.com/dic-iit/bipedal-locomotion-framework/pull/142#discussion_r522830098). 
However, it would make sense to eventually move other contacts-related components from the `Contact` library that are currently existing in the `Planners` folder as well into the new `Contacts` folder.

Consequently, we might also have to make changes in the python bindings generation? (do we? please correct me if I am wrong.)
Specifically, the changes will be
- renaming of `Contact` as `PlannedContact`
- renaming `ContactList`, `ContactPhase`, `ContactPhaseList` to corresponding `PlannedContact` equivalents
- replacing using namespace Planners` to `using namespace `Contacts`

cc @GiulioRomualdi @S-Dafarra  @diegoferigo

GiulioRomualdi:
This is the  C++ refactoring of `Contact` class. I'm not really sure about this solution for the following reasons:
1.  Right now we are exposing `Contacts` and `Contact`. The former is in https://github.com/dic-iit/bipedal-locomotion-framework/tree/master/src/Contacts while the latter is in https://github.com/dic-iit/bipedal-locomotion-framework/tree/master/src/Planners. This is confusing :sob:
    image

2. Shall we move also the `ContactLists` etc in `Contacts`?
3. I do not know how to handle the transition in the bindings for the following reasons
    1. the old `Contact` is just a typedef of `PlannedContact`,
    2. The new `Contact` class is a base class that we should wrap in python so this will destroy the retro compatibility
    3. I do not know how to handle the deprecation with pybind

cc @prashanthr05 @traversaro @S-Dafarra and @diegoferigo 

This will close #162


traversaro:
> The new Contact class is a base class that we should wrap in python so this will destroy the retro compatibility

I think it is safe to do so at this point? We have a big warning in the README exactly for that.

GiulioRomualdi:
If @paolo-viceconte and @diegoferigo agrees we can proceed in this way
1. in all your python code you should just substitute `Contact` with `PlannedContact`
2. If you want to run code that that requires the old implementation you can use the release 0.1.0 https://github.com/dic-iit/bipedal-locomotion-framework/releases/tag/v0.1.0

This will unblock @prashanthr05 from porting the estimator in python

diegoferigo:
Yes you can proceed in this way. No problem if some code here and there breaks, it seems a trivial change.

paolo-viceconte:
I agree as well.

prashanthr05:
Related issue: https://github.com/dic-iit/bipedal-locomotion-framework/issues/162

prashanthr05:
> 2.Shall we move also the ContactLists etc in Contacts?

I suppose we could move the data structures 
- `ContactList`, 
- `ContactPhase`, and 
- `ContactPhaseList` 

to `Contacts`. If we break, let's break it altogether in an uniform way. But, should we also rename them?

GiulioRomualdi:
> Yes you can proceed in this way. No problem if some code here and there breaks, it seems a trivial change.


This is the only  modification required in python
image

GiulioRomualdi:
I also created the bindings for the `EstimatedContacts` so @prashanthr05 can start using it

prashanthr05 (approved):
Looks good! Thanks @GiulioRomualdi !!

PR #207, opened 2021-02-24
LeggedOdometry: add changeFixedFrame overload with frame pose

prashanthr05:


src/Estimators/src/LeggedOdometry.cpp
```diff
@@ -905,3 +905,26 @@ bool LeggedOdometry::Impl::changeFixedFrame(const iDynTree::FrameIndex& newIdx,
 }
 
 
+bool LeggedOdometry::changeFixedFrame(const std::ptrdiff_t& newIdx,
+                                      const Eigen::Quaterniond& frameOrientationInWorld,
+                                      const Eigen::Vector3d& framePositionInWorld)
```
GiulioRomualdi, line 910: Why not eigen::ref?

prashanthr05: I thought about it, but here I followed an existing signature in the FloatingBaseEstimators.

GiulioRomualdi: ok

GiulioRomualdi (approved):

PR #220, opened 2021-03-02
[TimeVaryingDCMPlanner] restructure of setDCMReference() and setContactPhaseList()

GiulioRomualdi:
#208 implements   `setDCMReference()` however the user had to call the function after `setContactPhaseList()`.  Thanks to this PR, `setDCMReference()` does not depends on `setContactPhaseList()` and the user can call it whenever their want. 

cc @paolo-viceconte

### DIFF
--- a/src/Planners/src/TimeVaryingDCMPlanner.cpp
+++ b/src/Planners/src/TimeVaryingDCMPlanner.cpp
@@ -31,7 +31,10 @@ struct TimeVaryingDCMPlanner::Impl
                                           constraints related to the position of the Zero Moment
                                           Point (ZMP) */
 
-    BipedalLocomotion::Planners::QuinticSpline dcmRef;
+    BipedalLocomotion::Planners::QuinticSpline dcmRef;/**< Spline used to compute the internal DCM. This is used only if
+                                                         useExternalDCMReference is false. */
+    Eigen::MatrixXd externalDCMTrajectory; /**< External DCM trajectory. This is used only if
+                                              useExternalDCMReference is true. */
 
     casadi::Opti opti; /**< CasADi opti stack */
     casadi::MX costFunction; /**< Cost function of the optimization problem */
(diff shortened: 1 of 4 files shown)


prashanthr05 (approved):
Ok for me! But I would wait for @S-Dafarra approval as well.

src/Planners/src/TimeVaryingDCMPlanner.cpp
```diff
@@ -491,6 +494,8 @@ struct TimeVaryingDCMPlanner::Impl
     bool setupOptimizationProblem(const ContactPhaseList& contactPhaseList,
                                   const DCMPlannerState& initialState)
     {
+        constexpr auto errorPrefix = "[TimeVaryingDCMPlanner::Impl::setupOptimizationProblem] ";
```
S-Dafarra, line 497: Maybe we can add some tags to distinguish between errors and info messages.

S-Dafarra (approved):
Just a simple suggestion. So basically, you can set the initial dcm condition without changing the foot list, right?

GiulioRomualdi:
>  So basically, you can set the initial dcm condition without changing the foot list, right?

Right

GiulioRomualdi:
Comment https://github.com/dic-iit/bipedal-locomotion-framework/pull/220#issuecomment-789555938 updated. @S-Dafarra

src/Planners/src/TimeVaryingDCMPlanner.cpp
GiulioRomualdi: I would do it in a separate PR for all the error/warning/info in the repo. Indeed now all the errors/warnings/info are printed with `cerr` and there is not a common function to call.  https://github.com/dic-iit/bipedal-locomotion-framework/issues/137#issuecomment-736655938

GiulioRomualdi: For instance, we can use something like this: https://github.com/robotology/idyntree/blob/2e5cf7180a517bf91777d48761c4c12c94e7e404/src/core/include/iDynTree/Core/Utils.h#L66-L93

S-Dafarra: I was not referring to ``cerr`` but rather to the message itself. It is not clear whether the message is an error or not. I was suggesting adding something like ``[ERROR]`` since it is going to be much easier to find them in the middle of tons of other messages.

GiulioRomualdi: On that one I agree done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/220/commits/7999c2f0010291c9bda996d3a3b0eb218ce7c988

PR #226, opened 2021-03-04
Use std::optional instead of raw pointer in ISensorBridge

GiulioRomualdi:
**THIS PR BREAKS THE API**

This PR attempts to remove the usage of raw pointers in ISensorBridge. This can be done using `std::optional<std::reference_wrapper<double>>`. 
You can find further information in https://stackoverflow.com/questions/62454934/optionalreference-wrappert-vs-optionalt-practical-examples

cc @isorrentino @traversaro @S-Dafarra

### DIFF
--- a/src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
+++ b/src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
@@ -8,10 +8,12 @@
 #ifndef BIPEDAL_LOCOMOTION_ROBOT_INTERFACE_ISENSOR_BRIDGE_H
 #define BIPEDAL_LOCOMOTION_ROBOT_INTERFACE_ISENSOR_BRIDGE_H
 
+#include <functional>
 #include <memory>
+#include <optional>
 #include <string>
-#include <vector>
 #include <unordered_map>
+#include <vector>
 
 #include <Eigen/Dense>
 
--- a/src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
+++ b/src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
@@ -83,6 +83,8 @@ public:
 
     using weak_ptr = std::weak_ptr<ISensorBridge>;
 
+    using OptionalDoubleRef = std::optional<std::reference_wrapper<double>>;
+
     /**
      * Initialize estimator
      * @param[in] handler Parameters handler
@@ -161,7 +163,7 @@ public:
      */
     virtual bool getJointPosition(const std::string& jointName,
                                   double& jointPosition,
-                                  double* receiveTimeInSeconds = nullptr) { return false; };
+                                  OptionalDoubleRef receiveTimeInSeconds = {}) { return false; };
 
     /**
      * Get all joints' positions in radians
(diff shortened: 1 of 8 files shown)


src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -75,13 +72,19 @@ struct SensorBridgeMetaData
...
+    using Vector6d = Eigen::Matrix<double, 6, 1>;
+
 public:
     using unique_ptr = std::unique_ptr<ISensorBridge>;
 
     using shared_ptr = std::shared_ptr<ISensorBridge>;
 
     using weak_ptr = std::weak_ptr<ISensorBridge>;
 
+    using OptionalDoubleRef = std::optional<std::reference_wrapper<double>>;
```
S-Dafarra, line 86: Some good old IWYU problems :grin:

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -75,13 +72,19 @@ struct SensorBridgeMetaData
  */
 class ISensorBridge
 {
+protected:
+    using Vector12d = Eigen::Matrix<double, 12, 1>;
+    using Vector6d = Eigen::Matrix<double, 6, 1>;
```
S-Dafarra, line 77: Here you changed the accessibility of these aliases. Before they were public, now they are not accessible by the user. if these are used only in the implementation, I would move them in a cpp file.

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
GiulioRomualdi: Ops I forgot to add the includes

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
GiulioRomualdi: I put it there because before where not in a namespace so everyone including the files had Vector aliases.  However it’s nice to have it in the h because they simplify the declaration of the methods.

a possible solution is to include them in math component so everyone can use them

S-Dafarra: As you prefer. Just to express my point a little better, either set them as ``public`` or hide them completely. It would be weird to declare public methods using a private alias.

GiulioRomualdi: Ok I put them public but inside ISensorBridge class :smile_cat:

GiulioRomualdi: Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/226/commits/7ad35bcf0091a6a3271fe7d5a218400d39582c53

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
GiulioRomualdi: 3311ef2 cedd192

S-Dafarra (approved):

PR #229, opened 2021-03-08
Implement the Inverse kinematics in the repository

GiulioRomualdi:
This PR implements the InverseKinematics component in the repository. 

In details, it introduces the class `task` and the interface `inverse kinematics`, 
The `task` describes a generic task that can be handled in inverse kinematics. I also implemented 4 concrete classes that describe different tasks Namelly:
-  `SE3Task`: a task that ensures the tracking of a cartesian trajectory
- `SO3Task`:  it ensures the tracking of the desired orientation
- `JointsTrackingTask`: it ensures the tracking of a desired joint configuration. (You can use it as a regularization factor) 
- `CoMTask` : it ensures the tracking of the CoM 

classBipedalLocomotion_1_1IK_1_1Task__inherit__graph

Once the user has a list of tasks the user can build the inverse kinematics with the `InverseKinematics` interface.  

I also implemented a `QPInverseKinematics` class. This is a concrete implementation of an integration base IK. In this specific implementation, I used `osqp-eigen` to solve the `ik` problem.

classBipedalLocomotion_1_1IK_1_1QPInverseKinematics__inherit__graph

I also implemented some tests. 

The PR is ready to be reviewed @traversaro @S-Dafarra.


DanielePucci:
IB-IK in BLF @lrapetti @paolo-viceconte @Yeshasvitvs @claudia-lat

src/IK/include/BipedalLocomotion/IK/QPInverseKinematics.h
```diff
@@ -0,0 +1,117 @@
...
+namespace BipedalLocomotion
+{
+
+namespace IK
+{
+
+/**
+ * DCMPlanner defines a trajectory generator for the variable height Divergent
+ * component of motion (DCM). Please inherit publicly from this class in order to define your
+ * planner.
```
traversaro, line 32: Leftover?

GiulioRomualdi:
Implementation finished! The PR is ready to be reviewed

.github/workflows/ci.yml
```diff
@@ -163,12 +165,38 @@ jobs:
...
+        cmake --build . --config ${{ matrix.build_type }} --target install
+
+
         # CasADi
         # We override the casadi installation structure to be compliant with the folder structure used for
         # the other dependencies. Please check:
         # https://github.com/casadi/casadi/blob/6f122ca22e2a869903628c5738f154c8ac0f7455/CMakeLists.txt#L317
         cd ${GITHUB_WORKSPACE}
-        git clone https://github.com/casadi/casadi.git -b ${CasADi_TAG} casadi
+        git clone https://github.com/GiulioRomualdi/casadi.git -b ${CasADi_TAG} casadi
```
S-Dafarra, line 199: Maybe we should move that to ``dic`` to reduce the bus factor.

.github/workflows/ci.yml
```diff
@@ -257,13 +285,35 @@ jobs:
...
+        mkdir -p build
+        cd build
+        cmake -DCMAKE_PREFIX_PATH=${GITHUB_WORKSPACE}/install/deps -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} \
+              -DCMAKE_INSTALL_PREFIX=${GITHUB_WORKSPACE}/install/deps ..
+        cmake --build . --config ${{ matrix.build_type }} --target install
+
         # CasADi
         cd ${GITHUB_WORKSPACE}
-        git clone https://github.com/casadi/casadi.git -b ${CasADi_TAG} casadi
+        git clone https://github.com/GiulioRomualdi/casadi.git -b ${CasADi_TAG} casadi
```
S-Dafarra, line 311: Same as above

src/IK/include/BipedalLocomotion/IK/Task.h
```diff
@@ -0,0 +1,117 @@
...
+{
+
+/**
+ * Task describes a IK task element. The element is described by a matrix
+ * \f$A\f$ and a vector \f$b\f$. This class describes both a linear equality constraint and a linear
+ * inequality constraint. In case of equality constraint \f$ A \f$ and \f$ b \f$ represents: \f$ Ax
+ * = b\f$ In case of inequality constraint \f$ A \f$ and \f$ b \f$ represents: \f$ Ax \le b \f$
+ * @note Please inherit this class if you want to build your own optimal control problem.
+ */
+class Task
```
S-Dafarra, line 33: As discussed F2F, the name ``Task`` is very generic, but the interface is suitable only for linear tasks.

src/IK/include/BipedalLocomotion/IK/InverseKinematics.h
```diff
@@ -0,0 +1,96 @@
...
+     * @param priority is the priority of the task. The lower the number the higher the priority.
+     * @param weight weight associated to the task.
+     * @param isInequality true if the task should be treated as inequality constraint
+     * @return true if the task has been added to the inverse kinematics.
+     */
+    virtual bool addTask(std::shared_ptr<Task> task,
+                         const std::string& taskName,
+                         std::size_t priority,
+                         std::optional<Eigen::Ref<const Eigen::VectorXd>> weight = {},
+                         bool isInequality = false) = 0;
```
S-Dafarra, line 64: Ideally, the information on whether the task is an equality or not belongs to the task itself. I would consider moving this to the task class itself. In addition, for readability purposes, I would avoid using booleans when their meaning is not evident https://abseil.io/tips/94

src/IK/include/BipedalLocomotion/IK/InverseKinematics.h
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Get a vector containing the name of the tasks.
+     * @return an std::vector containing all the names associated to the tasks
+     */
+    virtual std::vector<std::string> getTasksName() const = 0;
```
S-Dafarra, line 70: Would it be more grammatically correct ``getTaskNames``?

S-Dafarra (commented):
Some initial comments

src/IK/include/BipedalLocomotion/IK/JointsTrackingTask.h
```diff
@@ -0,0 +1,97 @@
...
+
+    bool m_isInitialized{false};
+public:
+    /**
+     * Initialize the planner.
+     * @param paramHandler pointer to the parameters handler.
+     * @note the following parameters are required by the class
+     * |           Parameter Name           |   Type   |                                       Description                                      | Mandatory |
+     * |:----------------------------------:|:--------:|:--------------------------------------------------------------------------------------:|:---------:|
+     * |    `robot_velocity_variable_name`  | `string` |   Name of the variable contained in `VariablesHandler` describing the robot velocity   |    Yes    |
```
S-Dafarra, line 51: "robot_velocity" may not be immediately clear. Do you mean the vectorization of joint velocities and base velocity?

src/IK/include/BipedalLocomotion/IK/JointsTrackingTask.h
```diff
@@ -0,0 +1,97 @@
...
+     */
+    bool setSetpoint(Eigen::Ref<const Eigen::VectorXd> jointPosition);
+
+    /**
+     * Set the desired setpoint.
+     * @param jointPosition vector containing the desired joint position in radians.
+     * @param jointVelocity vector containing the desired joint velocity in radians per second.
+     * @return True in case of success, false otherwise.
+     */
+    bool setSetpoint(Eigen::Ref<const Eigen::VectorXd> jointPosition,
```
S-Dafarra, line 84: You used a different capitalization in ``CoMTask``

src/IK/include/BipedalLocomotion/IK/JointsTrackingTask.h
```diff
@@ -0,0 +1,97 @@
...
+ * want to control the desired joint position of the robot.
+ * The task represents the following equation
+ * \f[
+ * \begin{bmatrix} 0_6 & I_n \end{bmatrix} \nu = \dot{s} ^ *
+ * \f]
+ * where \f$0_6\f$ and \f$I_n\f$ are the zero and the identity matrix.
+ * The desired joint velocity is chosen such that the joint will converge to the desired
+ * trajectory and it is computed with a standard standard PD controller in \f$\mathbb{R}^n\f$.
+ */
+class JointsTrackingTask : public Task
```
S-Dafarra, line 33: Probably ``JointTrackingTask`` would be more grammatically correct.

src/IK/include/BipedalLocomotion/IK/InverseKinematics.h
S-Dafarra: Also, how are inequalities handled if they are not constraints?

src/IK/include/BipedalLocomotion/IK/QPInverseKinematics.h
```diff
@@ -0,0 +1,130 @@
...
+    ~QPInverseKinematics();
+
+    /**
+     * Add a task in the inverse kinematics
+     * @param task pointer to a given task
+     * @param priority is the priority of the task. The lower the number the higher the priority.
+     * @param weight weight associated to the task.
+     * @param isInequality true if the task should be treated as inequality constraint
+     * @note currently we support only task with priority 0 or 1. If the priority is set to 0 the
+     * task will be considered as equality constraint. This the weight is not required.
```
S-Dafarra, line 68: So, if ``priority`` is 0, the task will be considered an equality independently of ``isInequality``? 
Also
>```
>This the weight is not required.
>```
seems to have a typo.

src/IK/include/BipedalLocomotion/IK/QPInverseKinematics.h
```diff
@@ -0,0 +1,130 @@
...
+    bool finalize(const System::VariablesHandler& handler) override;
+
+
+    /**
+     * Initialize the planner.
+     * @param paramHandler pointer to the parameters handler.
+     * @note the following parameters are required by the class
+     * |           Parameter Name           |   Type   |                                       Description                                      | Mandatory |
+     * |:----------------------------------:|:--------:|:--------------------------------------------------------------------------------------:|:---------:|
+     * |    `robot_velocity_variable_name`  | `string` |   Name of the variable contained in `VariablesHandler` describing the robot velocity   |    Yes    |
```
S-Dafarra, line 92: Is this a copy-paste?

src/IK/include/BipedalLocomotion/IK/QPInverseKinematics.h
```diff
@@ -0,0 +1,130 @@
...
+    bool initialize(std::weak_ptr<ParametersHandler::IParametersHandler> handler) override;
+
+    /**
+     * Get a vector containing the name of the tasks.
+     * @return an std::vector containing all the names associated to the tasks
+     */
+    std::vector<std::string> getTasksName() const override;
+
+    /**
+     * Return if the content of get is valid.
```
S-Dafarra, line 105: I guess that a ``true`` is missing.

S-Dafarra (commented):
Some other comments

src/IK/include/BipedalLocomotion/IK/CoMTask.h
```diff
@@ -0,0 +1,105 @@
...
+     * |           Parameter Name           |   Type   |                                       Description                                      | Mandatory |
+     * |:----------------------------------:|:--------:|:--------------------------------------------------------------------------------------:|:---------:|
+     * |   `robot_velocity_variable_name`   | `string` |   Name of the variable contained in `VariablesHandler` describing the robot velocity   |    Yes    |
+     * |             `kp_linear`            | `double` |                             Gain of the position controller                            |    Yes    |
+     * @return True in case of success, false otherwise.
+     */
+    bool initialize(std::weak_ptr<ParametersHandler::IParametersHandler> paramHandler) override;
+
+    /**
+     * Set the variable required by the task
```
S-Dafarra, line 75: Variable handler?

This applies also to the other tasks.

src/IK/include/BipedalLocomotion/IK/Task.h
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Get the vector b.
+     * @return a const reference to the vector b.
+     */
+    Eigen::Ref<const Eigen::VectorXd> getB() const;
+
+    /**
+     * Get the description of the element.
+     * @return a string containing the description of the element.
```
S-Dafarra, line 103: ``element`` -> ``task``

src/IK/include/BipedalLocomotion/IK/CoMTask.h
```diff
@@ -0,0 +1,105 @@
...
+class CoMTask : public Task
+{
+    LieGroupControllers::ProportionalControllerR3d m_R3Controller; /**< P Controller in R3 */
+
+    System::VariablesHandler::VariableDescription m_robotVelocityVariable; /**< Variable
+                                                                                  describing the
+                                                                                  robot velocity
+                                                                                  (base + joint) */
+
+    static constexpr std::size_t m_linearVelocitySize{3}; /**< Size of the spatial velocity vector. */
```
S-Dafarra, line 55: ``spatial``->``linear``

src/IK/src/CoMTask.cpp
```diff
@@ -0,0 +1,153 @@
...
+                     "different from the one expected. Expected size: {}. Given size: {}.",
+                     m_kinDyn->getNrOfDegreesOfFreedom() + m_spatialVelocitySize,
+                     m_robotVelocityVariable.size);
+        return false;
+    }
+
+    // resize the matrices
+    m_A.resize(m_linearVelocitySize, variablesHandler.getNumberOfVariables());
+    m_A.setZero();
+    m_b.resize(m_linearVelocitySize);
```
S-Dafarra, line 50: Does it make sense to initialize also ``m_b`` to zero?

src/IK/src/CoMTask.cpp
```diff
@@ -0,0 +1,153 @@
...
+    // set the state
+    m_R3Controller.setState(toEigen(m_kinDyn->getCenterOfMassPosition()));
+
+    // update the controller
+    m_R3Controller.computeControlLaw();
+
+    m_b = m_R3Controller.getControl().coeffs();
+
+    // Workaround because matrix view is not compatible with Eigen::Ref
+    // https://github.com/robotology/idyntree/issues/797
```
S-Dafarra, line 131: I think this comment is no more necessary, right?

src/IK/src/JointsTrackingTask.cpp
```diff
@@ -0,0 +1,159 @@
...
+}
+
+std::size_t JointsTrackingTask::size() const
+{
+    if (m_kinDyn == nullptr)
+    {
+        log()->warn("[JointsTrackingTask::size] Please call setKinDyn method before. A size equal "
+                    "to zero will be returned.");
+        return 0;
+    }
```
S-Dafarra, line 157: This might be a good place for an ``assert``, in case the warning message gets unnoticed.

S-Dafarra (commented):
Still reviewing

src/IK/include/BipedalLocomotion/IK/InverseKinematics.h
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * InverseKinematics implements the interface for the inverse kinematics. Please inherits this class
+ * if you want to implement your custom IK.
+ */
+class InverseKinematics : public BipedalLocomotion::System::Advanceable<IKState>
+{
+
+public:
+    /**
+     * Initialize the planner.
```
S-Dafarra, line 46: Copy paste

src/IK/src/QPInverseKinematics.cpp
```diff
@@ -0,0 +1,405 @@
...
+    const bool taskExist = (m_pimpl->tasks.find(taskName) != m_pimpl->tasks.end());
+    if (taskExist)
+    {
+        log()->error("{} The task named {} already exist.", logPrefix, taskName);
+        return false;
+    }
+
+    if (priority != 0 && priority != 1)
+    {
+        log()->error("{} At the time being we support only priority equal to 0 or 1.", logPrefix);
```
S-Dafarra, line 148:
```suggestion
        log()->error("{} For the time being we support only priority equal to 0 or 1.", logPrefix);
```

S-Dafarra (commented):
WIP

src/IK/include/BipedalLocomotion/IK/InverseKinematics.h
```diff
@@ -0,0 +1,96 @@
...
+{
+    Eigen::VectorXd jointVelocity; /**< Joints velocity in rad per seconds */
+    manif::SE3d::Tangent baseVelocity; /**< Mixed spatial velocity of the base */
+};
+
+/**
+ * InverseKinematics implements the interface for the inverse kinematics. Please inherits this class
+ * if you want to implement your custom IK.
+ */
+class InverseKinematics : public BipedalLocomotion::System::Advanceable<IKState>
```
traversaro, line 41: This seems something that gives in output the robot velocity, so could it make sense to call it `VelocityInverseKinematics` , to distinguish it from the position-based kinematics?

src/IK/tests/QPInverseKinematicsTest.cpp
```diff
@@ -0,0 +1,193 @@
...
+            // check that everything went fine
+            REQUIRE(kinDyn->setRobotState(worldBasePos.asHomogeneousTransform(),
+                                          jointsPos,
+                                          ikAndTasks.ik->get().baseVelocity.coeffs(),
+                                          ikAndTasks.ik->get().jointVelocity,
+                                          gravity));
+
+            REQUIRE(ikAndTasks.ik->get().jointVelocity.isApprox(iDynTree::toEigen(jointsVel), tolerance));
+            REQUIRE(desiredCoMVelocity.isApprox(iDynTree::toEigen(kinDyn->getCenterOfMassVelocity()), tolerance));
+            REQUIRE(desiredVelocity.isApprox(iDynTree::toEigen(kinDyn->getFrameVel(controlledFrame)), tolerance));
```
S-Dafarra, line 190: To be honest, I am not sure if this is checking if the QPIK is working. You are initializing it already in the solution when setting the robot state in line 149, right?

S-Dafarra (commented):
Last comments

.github/workflows/ci.yml
GiulioRomualdi: c6a135a13f17c52155951cf0a032b01d3ed545c9

.github/workflows/ci.yml
GiulioRomualdi: c6a135a13f17c52155951cf0a032b01d3ed545c9

src/IK/include/BipedalLocomotion/IK/Task.h
GiulioRomualdi: What do you think of `LinearTask`?

S-Dafarra: Sounds good!

src/IK/include/BipedalLocomotion/IK/InverseKinematics.h
GiulioRomualdi: > Ideally, the information on whether the task is an equality or not belongs to the task itself. I would consider moving this to the task class itself.

I agree 


> I would avoid using booleans when their meaning is not evident https://abseil.io/tips/94

I can create a struct that contains the priority and the weight and pass it to the ik every time a new task is handled 

> Also, how are inequalities handled if they are not constraints?

In the specific QP implementation, this is not supported. I can return an error if the user wants to do that

S-Dafarra: >Also, how are inequalities handled if they are not constraints?

More relevant thread: Also, how are inequalities handled if they are not constraints?

src/IK/include/BipedalLocomotion/IK/JointsTrackingTask.h
GiulioRomualdi: > Do you mean the vectorization of joint velocities and base velocity?

Yes I can improve the description

src/IK/include/BipedalLocomotion/IK/InverseKinematics.h
S-Dafarra: > I can create a struct that contains the priority and the weight and pass it to the ik every time a new task is handled

If you move it inside the ``Task`` class, I think that is enough. For what concerns the priority, I like how it is done right now. It makes a lot of sense to set the priority only when adding the task, as it is more related to own the task is considered together with the others rather than the task itself.

src/IK/include/BipedalLocomotion/IK/InverseKinematics.h
GiulioRomualdi: Done in b7a79cab9bac0e934e9807bbdfa2e00a098f7574 and f05ee3afdcaa596e27f258a17ca05827b1d03a6c

src/IK/include/BipedalLocomotion/IK/JointsTrackingTask.h
GiulioRomualdi: Done in 915687caf100a7f5a548cb265b6f89bc326134ca and tests updated in 1660da99d070e25467569a0a842221241dcf5119 and 922836ac0b506cb19a28d4b8f855cf8fc24167aa

src/IK/src/JointsTrackingTask.cpp
GiulioRomualdi: Done in dd0018fdec02bc1cf173cbd88b2c581f9335de6e

src/IK/src/QPInverseKinematics.cpp
```diff
@@ -0,0 +1,405 @@
...
+    // Compute the gradient and the hessian
+    m_pimpl->hessian.setZero();
+    m_pimpl->gradient.setZero();
+    for (const auto& cost : m_pimpl->costs)
+    {
+        Eigen::Ref<const Eigen::MatrixXd> A = cost.get().task->getA();
+        Eigen::Ref<const Eigen::VectorXd> b = cost.get().task->getB();
+
+        m_pimpl->hessian += A.transpose() * cost.get().weight.asDiagonal() * A;
+        m_pimpl->gradient -= A.transpose() * cost.get().weight.asDiagonal() * b;
```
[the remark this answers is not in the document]
GiulioRomualdi: I think we can avoid to add a buffer following what is written here http://eigen.tuxfamily.org/dox/TopicWritingEfficientProductExpression.html
```suggestion
        m_pimpl->hessian.noalias() += A.transpose() * cost.get().weight.asDiagonal() * A;
        m_pimpl->gradient.noalias() -= A.transpose() * cost.get().weight.asDiagonal() * b;
```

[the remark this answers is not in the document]
S-Dafarra: I think that works only if ``weight`` is a scalar, i.e.
```
m_pimpl->hessian.noalias() += cost.get().weight * (A.transpose() * A);
```
so that we fall in the second row of the table in http://eigen.tuxfamily.org/dox/TopicWritingEfficientProductExpression.html. Otherwise, it is a three matrix multiplication, and I think that still requires a temporary object.

traversaro:
To be honest I found a bit difficult to understand what the InverseKinematics interface was just from the docs, so I tried to sketch a diagram (always remembering that every system can be thought in terms of inputs/outpus/state/parameters) of a **possible** use of this class (possible because the use can be different in other contexts, for example I guess @lrapetti use case is different):

ExplanationWalkingIK


(raw diagram: 
[ExplanationWalkingIK.zip](https://github.com/dic-iit/bipedal-locomotion-framework/files/6115222/ExplanationWalkingIK.zip)
 )


In this diagram, the logic that is exposed by the `BipedalLocomotion::IK::InverseKinematics`  interface and the  `BipedalLocomotion::IK::QPInverseKinematics` is the one contained in the yellow block. 

My doubts are the following:
* The name `InverseKinematics` for something that takes in input task positions (and optionally task velocities) and outputs desired robot state may be misleading, especially as we also used quite a lot the Ipopt-based inverse kinematics in which the input are also task positions, and the outputs are instead robot state positions (something that can be obtained also in this case by attaching to the proposed system an integrator). Even `VelocityInverseKinematics` may be misleading. To be honest I can't think about a good name, so feel free to proceed as otherwise we risk to just loose time bikeshedding, but I just to point out this. Probably it is just sufficient to briefly describe what this interface is (i.e. what are the inputs and what are the outputs) in the interface descriptions.
* What I indicate in the schema as "Velocity Inverse Kinematics" as far as I understand can't be used decouple from the tasks, i.e. injecting directly the velocity that you want. That make sense if it simplifies implementation and directly setting the task velocity is not needed, but I wanted to double check this two hypothesis with you.
* The `setKinDyn` method used by Tasks and by the IK in general is used for two different goals: one is set the "Kinematic Model" **parameter** of the system, and the other is to pass the $s^m$ **input** to the inverse kinematics system. If this is convenient for the implementation it make sense, but I would perhaps mention it in the setKinDyn docs as it may be confusing for users (for example, one user may want to use as feedback in the taks controllers the output of the integrator $s^d$ instead of the measured position $s^m$ or the other way around, and how this is done is a bit criptic.
* Not relevant for the PR, but just for me to understand: in the diagram I indicated a block as `???` as I don't know how you go from $\nu^d$ to $\dot{s}^d$, which logic is there?

src/IK/include/BipedalLocomotion/IK/InverseKinematics.h
GiulioRomualdi: Done in 56b575c33a0152d1774a91b545415252cfd036a3 421043c08b12d00e53bcbfa73e1a1d40158cc5c0 da394d9231983063376294518ff6cdb602867bb1 e049f4e899bea387162621ea8380da52f84b09f3 2453f9e199d2f8866d637864ef63a5c1293c293d f0cee977e288508bc666a9180b4816f78db2356d

GiulioRomualdi:
Hi @traversaro 

> 1. The name InverseKinematics for something that takes in input task positions (and optionally task velocities) and outputs desired robot state may be misleading, especially as we also used quite a lot the Ipopt-based inverse kinematics in which the input are also task positions, and the outputs are instead robot state positions (something that can be obtained also in this case by attaching to the proposed system an integrator). Even VelocityInverseKinematics may be misleading. To be honest I can't think about a good name, so feel free to proceed as otherwise we risk to just loose time bikeshedding, but I just to point out this. Probably it is just sufficient to briefly describe what this interface is (i.e. what are the inputs and what are the outputs) in the interface descriptions.

We can also rename it in `IntegrationBaseInverseKinematics`, what do you think?


> 2. What I indicate in the schema as "Velocity Inverse Kinematics" as far as I understand can't be used decouple from the tasks, i.e. injecting directly the velocity that you want. That make sense if it simplifies implementation and directly setting the task velocity is not needed, but I wanted to double check this two hypothesis with you.

The inverse kinematics is strictly related to `task`.  If we need a pure velocity task we can easily implement it. 

> 3. The setKinDyn method used by Tasks and by the IK in general is used for two different goals: one is set the "Kinematic Model" parameter of the system, and the other is to pass the $s^m$ input to the inverse kinematics system. If this is convenient for the implementation it make sense, but I would perhaps mention it in the setKinDyn docs as it may be confusing for users (for example, one user may want to use as feedback in the taks controllers the output of the integrator $s^d$ instead of the measured position $s^m$ or the other way around, and how this is done is a bit cryptic.

The `kinDyn` object is used in the tasks just as a container of information (e.g. joint position or frame pose). The user should feed the `kindyn` object from the outside. So in the case of IK, the `kindyn` should contain desired values evaluated at the previous step (and integrated) otherwise if the user wants to use the IK class as a velocity controller `kindyn` should contain measured values. 

> 4. Not relevant for the PR, but just for me to understand: in the diagram I indicated a block as ??? as I don't know how you go from $\nu^d$ to $\dot{s}^d$, which logic is there?

In the `QPInverseKinematics` $\nu = [\dot{p}; \omega \dot{s}]$

traversaro:
> > 1. Not relevant for the PR, but just for me to understand: in the diagram I indicated a block as ??? as I don't know how you go from νd\nu^d to s˙d\dot{s}^d, which logic is there?
> 
> In the `QPInverseKinematics` ν=[p˙;ωs˙]

As per t2t discussion, tipically only the joint part of the vector is taken and the base is ignored, because the tasks of the QP are such that the velocities are compatible with the active contact constraints. 

I updated the diagram accordingly:
ExplanationWalkingIK

[ExplanationWalkingIK.zip](https://github.com/dic-iit/bipedal-locomotion-framework/files/6115634/ExplanationWalkingIK.zip)

src/IK/src/QPInverseKinematics.cpp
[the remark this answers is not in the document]
GiulioRomualdi: I investigated a bit in the solution you propose. I took the approach from https://stackoverflow.com/questions/52918480/identifying-temporary-object-creation-in-eigen

The following code prints a string every time a dynamic allocation is done. 

```cpp
#include <cassert>
#include <iostream>

static long int nb_temporaries;
inline void on_temporary_creation(long int size) {
  // here's a great place to set a breakpoint when debugging failures in this test!
  if(size!=0) nb_temporaries++;

  std::cerr << "Number of allocation " << nb_temporaries << std::endl ;
}

#define EIGEN_DENSE_STORAGE_CTOR_PLUGIN { on_temporary_creation(size); }

#include <Eigen/Dense>

int main()
{
    Eigen::MatrixXd a(10,10);
    Eigen::MatrixXd c(10,10);
    Eigen::VectorXd weight(10);

    c.setIdentity();
    weight.setRandom();
    a.setRandom();


    // option 1
    Eigen::MatrixXd temp(10,10);
    temp.noalias() = a.transpose() * weight.asDiagonal() * a;
    c.noalias() += temp;

    // option 2
    // c.noalias() += a.transpose() * weight.asDiagonal() * a;

    return 0;
}
```

If I run the code with `option 1` enabled I got
```
Number of allocation 1
Number of allocation 2
Number of allocation 3
Number of allocation 4
Number of allocation 5
```

Allocation 1 is for `a`, allocation 2 is for `c`, allocation 3 if for `weight` and allocation 4 is for `temp`. (this dynamic allocation are done only once in our case). Allocation 5 is a temporary object created in `temp.noalias() = a.transpose() * weight.asDiagonal() * a;` Thus every cicle eigen allocate a matrix.

If I run the code with `option 2` enabled I got
```
Number of allocation 1
Number of allocation 2
Number of allocation 3
Number of allocation 4
```
Allocation 1 is for `a`, allocation 2 is for `c`, allocation 3 if for `weight`. allocation 4 is a temporary object created in
`c.noalias() += a.transpose() * weight.asDiagonal() * a;`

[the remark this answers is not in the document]
GiulioRomualdi: This is a possible solution 
```
    Eigen::MatrixXd temp(10,10);                                                                     
    temp.noalias() = a.transpose() * weight.asDiagonal();
    c.noalias() += temp * a;
```

In this case we have 
```
Number of allocation 1
Number of allocation 2
Number of allocation 3
Number of allocation 4
```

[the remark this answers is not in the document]
GiulioRomualdi: This patch goes in the direction you propose. However it breaks the symmetry between cost and constraints (`std::vector<std::reference_wrapper<const TaskWithPriority>> costs` -> `std::vector<std::reference_wrapper<TaskWithPriority>> costs`) (i.e. the `const` disappear)


```patch
diff --git a/src/IK/src/QPInverseKinematics.cpp b/src/IK/src/QPInverseKinematics.cpp
index f85b8d8f..6845d00d 100644
--- a/src/IK/src/QPInverseKinematics.cpp
+++ b/src/IK/src/QPInverseKinematics.cpp
@@ -21,6 +21,8 @@ struct QPInverseKinematics::Impl
         std::shared_ptr<Task> task;
         std::size_t priority;
         Eigen::VectorXd weight;
+
+        Eigen::MatrixXd tmp;
     };
 
     IKState solution;
@@ -32,7 +34,7 @@ struct QPInverseKinematics::Impl
     std::unordered_map<std::string, TaskWithPriority> tasks;
 
     std::vector<std::reference_wrapper<const TaskWithPriority>> constraints;
-    std::vector<std::reference_wrapper<const TaskWithPriority>> costs;
+    std::vector<std::reference_wrapper<TaskWithPriority>> costs;
 
     Eigen::MatrixXd hessian;
     Eigen::VectorXd gradient;
@@ -207,6 +209,12 @@ bool QPInverseKinematics::finalize(const System::VariablesHandler& handler)
         }
     }
 
+    // resize the  temporary matrix usefull to reduce dynamics allocation while advance is called
+    for (auto& cost : m_pimpl->costs)
+    {
+        cost.get().tmp.resize(handler.getNumberOfVariables(), cost.get().weight.size());
+    }
+
     m_pimpl->solver.data()->setNumberOfVariables(handler.getNumberOfVariables());
     m_pimpl->solver.data()->setNumberOfConstraints(numberOfConstraints);
 
@@ -262,13 +270,15 @@ bool QPInverseKinematics::advance()
     // Compute the gradient and the hessian
     m_pimpl->hessian.setZero();
     m_pimpl->gradient.setZero();
-    for (const auto& cost : m_pimpl->costs)
+    for (auto& cost : m_pimpl->costs)
     {
         Eigen::Ref<const Eigen::MatrixXd> A = cost.get().task->getA();
         Eigen::Ref<const Eigen::VectorXd> b = cost.get().task->getB();
 
-        m_pimpl->hessian += A.transpose() * cost.get().weight.asDiagonal() * A;
-        m_pimpl->gradient -= A.transpose() * cost.get().weight.asDiagonal() * b;
+        cost.get().tmp.noalias() = A.transpose() * cost.get().weight.asDiagonal();
+
+        m_pimpl->hessian.noalias() += cost.get().tmp * A;
+        m_pimpl->gradient.noalias() -= cost.get().tmp * b;
     }
 
     // compute the constraints
```

src/IK/include/BipedalLocomotion/IK/QPInverseKinematics.h
GiulioRomualdi: There is also `verbosity`. In bc04b65075df62ec82a7abab3184d38761c8dcd0 I improved the documentation

src/IK/include/BipedalLocomotion/IK/QPInverseKinematics.h
GiulioRomualdi: Fixed in 5896d41fc6da6f533fccdcdb22a82c4d852d68ce

src/IK/include/BipedalLocomotion/IK/SE3Task.h
```diff
@@ -0,0 +1,109 @@
...
+    iDynTree::FrameIndex m_frameIndex; /**< Frame controlled by the OptimalControlElement */
+
+    static constexpr std::size_t m_spatialVelocitySize{6}; /**< Size of the spatial velocity vector. */
+
+    bool m_isInitialized{false}; /**< True if the task has been initialized. */
+
+public:
+
+    /**
+     * Initialize the planner.
```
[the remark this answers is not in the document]
GiulioRomualdi: Fixed 9c65b9fa4c53161183995c8fa6b5e6b864f4278e thanks

src/IK/include/BipedalLocomotion/IK/JointsTrackingTask.h
GiulioRomualdi: Done!

src/IK/include/BipedalLocomotion/IK/Task.h
GiulioRomualdi: 2e990780cd55007b27e35cf3021e18032d4eb46a

src/IK/src/QPInverseKinematics.cpp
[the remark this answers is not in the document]
GiulioRomualdi: Done here 852c792b33903e8fa49f024420fb8e50c28847da

src/IK/src/CoMTask.cpp
GiulioRomualdi: Done in e60156cfd83505d4493b73eab031819a08169c15

src/IK/include/BipedalLocomotion/IK/InverseKinematics.h
GiulioRomualdi: Done in 19b7f38f8f6711d7c8ec18d4ee4a0bb73808f42f

src/IK/src/QPInverseKinematics.cpp
GiulioRomualdi: Fixed here: dfca80a61830a2b8514031544154a8235931e201

src/IK/src/QPInverseKinematics.cpp
```diff
@@ -0,0 +1,405 @@
...
+        return false;
+    }
+
+    if (priority == 1 && !weight)
+    {
+        log()->error("{} In case of priority equal to 1 the weight is mandatory.", logPrefix);
+        return false;
+    }
+
+    if (weight)
```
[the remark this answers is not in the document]
GiulioRomualdi: Done in dfca80a61830a2b8514031544154a8235931e201

src/IK/src/QPInverseKinematics.cpp
```diff
@@ -0,0 +1,405 @@
...
+    if (priority == 0)
+    {
+        if (isInequality)
+            m_pimpl->inequalityConstraints.push_back(m_pimpl->tasks[taskName]);
+        else
+            m_pimpl->equalityConstraints.push_back(m_pimpl->tasks[taskName]);
+    } else
+    {
+        m_pimpl->costs.push_back(m_pimpl->tasks[taskName]);
+    }
```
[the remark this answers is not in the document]
GiulioRomualdi: For the time being, this feature is not supported. I just added a check in  dfca80a61830a2b8514031544154a8235931e201. If the user sets `priority = 1` and the task is `inequality` an error is returned.

src/IK/src/QPInverseKinematics.cpp
```diff
@@ -0,0 +1,405 @@
...
+
+    // update of all the tasks
+    for (auto& [name, task] : m_pimpl->tasks)
+    {
+        task.task->update();
+    }
+
+    // Compute the gradient and the hessian
+    m_pimpl->hessian.setZero();
+    m_pimpl->gradient.setZero();
```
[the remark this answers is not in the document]
GiulioRomualdi: The hessian is given by the sum of the hessian (same for the gradient).

I think this is related to https://github.com/dic-iit/bipedal-locomotion-framework/pull/229#discussion_r590588096

src/IK/src/SE3Task.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+    // update the controller
+    m_SO3Controller.computeControlLaw();
+    m_R3Controller.computeControlLaw();
+
+    m_b.head<3>() = m_R3Controller.getControl().coeffs();
+    m_b.tail<3>() = m_SO3Controller.getControl().coeffs();
+
+    // Workaround because matrix view is not compatible with Eigen::Ref
+    // https://github.com/robotology/idyntree/issues/797
```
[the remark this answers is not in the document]
GiulioRomualdi: Done 14de72d982eb15c2e796a252f925735ac4702827

src/IK/src/QPInverseKinematics.cpp
```diff
@@ -0,0 +1,405 @@
...
+
+    bool isFirstIteration{true};
+    bool isValid{false};
+    bool isInitialized{false};
+
+    bool initializeSolver()
+    {
+        constexpr auto logPrefix = "[QPInversekinematics::Impl::initializeSolver]";
+        // Hessian matrix
+        Eigen::SparseMatrix<double> hessianSparse = this->hessian.sparseView();
```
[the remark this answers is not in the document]
GiulioRomualdi: I completely agree with this comment.

As you know osqp-eigen wants sparse matrices, however, it is not possible to write in a block of a sparse matrix so this makes more complex the usage in our scenario.

The task may return the sparsity pattern of A (b is always dense). However, I don't know how to build the sparsity pattern of  `A.transpose() * A`

Once we have the sparsity pattern we may think to copy the elements of the dense matrix into a sparse matrix. We may use a double for loop but I hope there will be something more efficient. 

I think that this comment requires a separate issue and investigation. For the time being, we can keep this approach (that is already used in `WalkingControllers`) and then find a possible solution later on. 

I'll open the issue as soon as the PR gets merged

src/IK/src/QPInverseKinematics.cpp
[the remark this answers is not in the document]
S-Dafarra: So if the entirety of the hessian and the gradient are recomputed every time, there is no need to initialize them to zero in the update method. I agree on initializing them to zero once, but then doing that at every control loop seems not necessary.

src/IK/include/BipedalLocomotion/IK/Task.h
GiulioRomualdi: Here the commits 7975cd4e22e520aa719241b8bee4e55b8b4acd94 8d17d8c2aa86762f8dc8f4af3233fbba44cd0af7 91bea8d9233cb835fd296731156de5aa87009364 eebe87c2d2352d10121c4bdca4da810cb728acc5 8943124cb5a410c3998782790bc3ca5c030886c8 cc617c066233c0f549dc886d2e898e65cc80d081 51f396a1048beccd68713051cc22eea48f8cdfa9 d57d384474d63dc70535882d2ba72490c4831a54

src/IK/tests/QPInverseKinematicsTest.cpp
S-Dafarra: It really depends on what you want to test :wink:

src/IK/src/QPInverseKinematics.cpp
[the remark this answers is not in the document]
S-Dafarra: Ah no, I see it now, sorry for the delay :grin: Since you are adding things, you need to initialize to something. You could have initialized using the first hessian, but in this way, it is cleaner. 

Ignore this comment

src/IK/tests/QPInverseKinematicsTest.cpp
GiulioRomualdi: As discussed F2F I reimplemented the test in https://github.com/dic-iit/bipedal-locomotion-framework/pull/229/commits/df1e11f3405405211b6a04ccbc9466c297d445f3

Here I run the IK sequentially and I check that the solution converged to the desired one. The tests run in three different models automatically generated by iDynTree.

While I was writing the test I noticed that `osqp-eigen` does not support unconstrained optimization problems. This is actually a limitation since we may want to consider all the tasks in the cost function (If I'm not mistaken the AnDy ik works in this way @lrapetti )

S-Dafarra: Is it a limitation of ``osqp`` or ``osqp-eigen``? It may be worked around with some box constraints on the control variables, eventually with large (but not infinite) bounds.

GiulioRomualdi:
For the sake of completeness, I add the two schemes that can be used as a while using the IK component.

The IK component can actually be used as Velocity controller or real IK. Indeed it is important to notice that `IntegrationBaseIKState` is a struct containing the joints velocity. When a robot velocity controller is available, one can set these joint velocities to the low-level robot controller. In this case, the $i^d$ quantities in the following figures  can be evaluated by using robot sensor feedback, and the robot is said to be velocity controlled. On the other hand, if the robot velocity control
is not available, one may integrate the outcome of IntegrationBasedIK to obtain the desired joint position to be set to a low-level robot position controller. In this case, the $i ^d$ quantities can be evaluated by using the desired integrated quantities instead of
sensor feedback, and the block behaves as an inverse kinematics module, and the robot is said to be position controlled.

Here you can find an example of the IK component used as a velocity controller 
Velocity controller

Here you can find an example of the IK component used as inverse kinematics. In this case, the `System::FloatingBaseSystemDynamics` and `System::Integrator` classes can be used to integrate the output of the IK taking into account the geometrical structure of the configuration space ($ R^3 \times SO(3) \times R^n$)
IK


Here the updated blocks [ExplanationWalkingIK.zip](https://github.com/dic-iit/bipedal-locomotion-framework/files/6119036/ExplanationWalkingIK.zip)


cc @traversaro

src/IK/include/BipedalLocomotion/IK/SE3Task.h
```diff
@@ -0,0 +1,124 @@
...
+     * System::VariablesHandler.
+     * @param variablesHandler reference to a variables handler.
+     * @note The handler must contain a variable named as the parameter
+     * `robot_velocity_variable_name` stored in the parameter handler. The variable represents the
+     * generalized velocity of the robot. Where the generalized robot velocity is a vector
+     * containing the base spatial-velocity (expressed in mixed representation) and the joints
+     * velocity.
+     * @return True in case of success, false otherwise.
+     */
+    bool setVariableHandler(const System::VariablesHandler& variablesHandler) override;
```
S-Dafarra, line 92: I just noticed that these methods are called ``setVariableHandler`` while the object is called ``VariablesHandler`` (note the ``s``). Consider uniforming them

src/IK/include/BipedalLocomotion/IK/SO3Task.h
```diff
@@ -0,0 +1,119 @@
...
+namespace BipedalLocomotion
+{
+namespace IK
+{
+
+/**
+ * SO3Task is a concrete implementation of the IK::Task. Please use this element if you
+ * want to control the position and the orientation of a given frame rigidly attached to the robot.
+ * The task assumes perfect control of the robot velocity \f$\dot{\nu}\f$ that contains the base
+ * linear and angular velocity expressed in mixed representation and the joints velocity.
```
S-Dafarra, line 29: I think there is a typo: "joints velocity" -> "joint velocities". I think this affects the documentation of the other tasks as well

src/IK/src/QPInverseKinematics.cpp
```diff
@@ -0,0 +1,417 @@
...
+    constexpr auto logPrefix = "[QPInverseKinematics::advance]";
+
+    // when advance is called the previous solution is no more valid
+    m_pimpl->isValid = false;
+
+    if (!m_pimpl->isInitialized)
+    {
+        log()->error("{} Please call initialize() before advance().", logPrefix);
+        return false;
+    }
```
S-Dafarra, line 273: I think that it is worth checking that ``finalize`` has been called too.

src/IK/tests/QPInverseKinematicsTest.cpp
```diff
@@ -0,0 +1,286 @@
...
+            auto ikAndTasks = createIK(parameterHandler, kinDyn, variablesHandler);
+
+            REQUIRE(ikAndTasks.se3Task->setSetPoint(desiredSetPoints.endEffectorPose,
+                                                    manif::SE3d::Tangent::Zero()));
+            REQUIRE(ikAndTasks.comTask->setSetPoint(desiredSetPoints.CoMPosition,
+                                                    Eigen::Vector3d::Zero()));
+            REQUIRE(ikAndTasks.regularizationTask->setSetPoint(desiredSetPoints.joints));
+
+            // propagate the inverse kinematics for
+            constexpr std::size_t iterations = 1200;
```
S-Dafarra, line 237: This test required 20min to valgrind to be finished. Please, start from a point that would require less iterations to converge (like 10-15 maybe)

src/IK/src/QPInverseKinematics.cpp
```diff
@@ -0,0 +1,405 @@
...
+    for (const auto& constraint : m_pimpl->equalityConstraints)
+    {
+        Eigen::Ref<const Eigen::MatrixXd> A = constraint.get().task->getA();
+        Eigen::Ref<const Eigen::VectorXd> b = constraint.get().task->getB();
+
+        m_pimpl->lowerBound.segment(index, constraint.get().task->size()) = b;
+        m_pimpl->upperBound.segment(index, constraint.get().task->size()) = b;
+        m_pimpl->constraintMatrix.middleRows(index, constraint.get().task->size()) = A;
+
+        index += constraint.get().task->size();
```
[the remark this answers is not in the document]
S-Dafarra: Note that it would be enough to check if at the end ``index`` is equal to the total number of constraints. You cannot know which constraint increased in dimension, but you know that at least one did.

traversaro:
> For the sake of completeness, I add the two schemes that explain how the class can be used.

Great! For the sake of completeness, at least on iCub and R1 robots as of 2021.02 distro version the low level joint velocity controller is actually just integrating the desired velocity and passing it to the position controller (see https://github.com/vvv-school/vvv18/blob/cf1ed79686bb14ba083d95109f81ac356388fe9a/material/yarp/vvv18-yarp-motor-control.pdf), so in the iCub case the two scheme are more similar then it may seems, even if they are different due to the nature and the delay of the feedback used in the "task proportional controllers".

traversaro (approved):

src/IK/tests/QPInverseKinematicsTest.cpp
lrapetti: > This is actually a limitation since we may want to consider all the tasks in the cost function (If I'm not mistaken the AnDy ik works in this way @lrapetti )

Yes, I confirm all the task are in the const function. However, if the solver is set to QP we use `osqp-eigen` as well, and we use joint velocity constraints (not sure if you are forced to set them, @kouroshD worked on that part and might know better).

kouroshD: > While I was writing the test I noticed that osqp-eigen does not support unconstrained optimization problems. This is actually a limitation since we may want to consider all the tasks in the cost function (If I'm not mistaken the AnDy ik works in this way @lrapetti )


I am not sure why you are making this decision to avoid using constraint in IK. 
Since a constraint is ensured to be respected (with a good sampling time), while a soft constraint added to the task can be violated. Anyhow, as far as I remember and logically, you should be able to solve the optimization problem without constraints as well.

S-Dafarra: That's just a possible implementation. It has pros and cons. For example, you may decide to use no constraints so that you are sure you always get a solution, even if you have two completely contradictory tasks.

kouroshD: Note that in a manifold,  if you have a set of constraints such that the possible space of your solution is not empty, it should retrieve a _best solution_ as a result, and it is the task of the designer to do not fix such contradictory constraints!

S-Dafarra: > and it is the task of the designer to do not fix such contradictory constraints!

You should not trust anyone :grin: 

Jokes apart, one thing is what you should do or not, another thing is a limitation introduced by the software. You should not justify a flaw in the design with the excuse that the corresponding use cases are not good enough.

src/IK/include/BipedalLocomotion/IK/QPInverseKinematics.h
```diff
@@ -0,0 +1,134 @@
...
+     * @note currently we support only task with priority 0 or 1. If the priority is set to 0 the
+     * task will be considered as a constraint. In this case the weight is not required.
+     * @warning The QPInverseKinamatics cannot handle  handle inequality tasks (please check
+     * Task::Type) with priority equal to 1.
+     * @return true if the task has been added to the inverse kinematics.
+     */
+    bool addTask(std::shared_ptr<LinearTask> task,
+                 const std::string& taskName,
+                 std::size_t priority,
+                 std::optional<Eigen::Ref<const Eigen::VectorXd>> weight = {}) override;
```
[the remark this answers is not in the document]
GiulioRomualdi: 6c08903c3935d0f3633cf945fa08c014049aa3e1 and edaa27c502fd5860709a6e62d4c1b7e84c1e2b5d

src/IK/include/BipedalLocomotion/IK/SE3Task.h
GiulioRomualdi: Here a20a0bb6049b610ec387893b128160628b29bf5f 1e984f31ca741bc2f7c6b6ad6353b1810c5284af df7db576f70c848476d873b3182db329669cbac0 e79ad62722ea71ad14a242385140c01db78205ca b4b26e7800bf4295cdffcd8b1c024e90b1f6b891 54f28336d9eb9c5ac9233597e150f741836a4752 4c80a83e04e79fe67c381c711f6d3507054fc904 44cb5d7347bb275393c69e1f56c2d8c5a4907ba2 ccf53b79a0993be9eb3ca22a82acd228c8a371bd e3fec1aef61ab2091b6d8e0eee5dc5e235c79ea2

src/IK/tests/QPInverseKinematicsTest.cpp
GiulioRomualdi: Done fd6618167931b8f3ced0f1b2a0ec72fdb52c687a

src/IK/include/BipedalLocomotion/IK/SO3Task.h
GiulioRomualdi: Here e7259e9b2dd9ea3b190127aa6bb37aee85839e5d ac2c75c243fbd825e44f6128bd2f18b3520c837a b062a812f4053688913d2ef34a96812c2d03a902 e83a331f3f2b6c6388046ba76522c8e542d436ac 73412f709b8bf6a9aa1a89bf7e07a10fcf6b3d3f 9b2e205084e86398b5a3f66c311a7dd10cdc285f

src/IK/src/QPInverseKinematics.cpp
GiulioRomualdi: Done here a8329da9

GiulioRomualdi:
@S-Dafarra I apply all your comment let me know if you are fine with the current implementation. 
After your approval, I will reorganize the commits before the merge

.github/workflows/conda-forge-ci.yml
```diff
@@ -30,19 +30,19 @@ jobs:
     - name: Dependencies
       shell: bash -l {0}
       run: |
-        # Compilation related dependencies 
+        # Compilation related dependencies
         mamba install cmake compilers make ninja pkg-config
         # Actual dependencies
-        mamba install -c robotology-staging idyntree yarp matio-cpp lie-group-controllers eigen qhull casadi cppad manif spdlog catch2
-        
+        mamba install -c robotology-staging idyntree yarp matio-cpp lie-group-controllers eigen qhull casadi cppad manif spdlog catch2 osqp osqp-eigen
```
traversaro, line 36:
```suggestion
        mamba install -c robotology-staging idyntree yarp matio-cpp lie-group-controllers eigen qhull casadi cppad manif spdlog catch2 libosqp osqp-eigen
```

traversaro:
The conda build is failing as the change in osqp-eigen still needs to be released. Probably we can disable that in conda until it is released.

src/IK/tests/QPInverseKinematicsTest.cpp
GiulioRomualdi: https://github.com/robotology/osqp-eigen/issues/84

S-Dafarra (approved):

GiulioRomualdi:
Here two updated images:

image

image

[ExplanationWalkingIK.zip](https://github.com/ami-iit/bipedal-locomotion-framework/files/7558088/ExplanationWalkingIK.zip)

PR #243, opened 2021-03-22
Avoid the "Generating the Urdf Model from"  message to appear when doing ccmake

S-Dafarra:
With STATUS, Cmake automatically adds the leading "--" and a message avoids to popup when using ccmake.


GiulioRomualdi (approved):

GiulioRomualdi:
Thank you @S-Dafarra

PR #249, opened 2021-03-24
Fix - update of m_isValid variable added in JointTrackingTask

isorrentino:


GiulioRomualdi (commented):
Thank you @isorrentino 
Could you please add also the check in the test?
For instance you may add 
```cpp
REQUIRE(task.isValid());
```

 after 

https://github.com/dic-iit/bipedal-locomotion-framework/blob/fe9c3d2237a51347db1c7d641cbfa350538cece9/src/TSID/tests/JointTrackingTaskTest.cpp#L93-L94

isorrentino:
> Could you please add also the check in the test?

@GiulioRomualdi done. I added it to all the tests, both in IK and TSID.

GiulioRomualdi:
Thank you @isorrentino. Let's wait for the CI

GiulioRomualdi (approved):

PR #255, opened 2021-03-27
Fix pybind11 version

GiulioRomualdi:
This follows https://github.com/dic-iit/bipedal-locomotion-framework/issues/246#issuecomment-808711753

Now we can use pybind11 installed from apt in ubuntu 20.04


traversaro (approved):
Great!

PR #256, opened 2021-03-28
Implement Block, BlockRunner and SharedResource classes

GiulioRomualdi:
This PR introduces the concept of Block, BlockRunner and SharedResource inside the repo. Imagine the Block as a standard Simulink block with input and output. Here we want to run several blocks in parallel with a different sampling time. This may be required in the case of a cascade control loop where the inner control loop runs faster than the other. 
The `BlockRunner` is a class that helps you in creating a periodic thread. 

Finally, the `SharedResource` is the way that can be used to communicate between two blocks running in separate threads. Indeed here it is crucial to guarantee correct access to the resource avoiding collisions. 

Here a picture that should explain the concept

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ All notable changes to this project are documented in this file.
 - Implement python bindings for VariablesHandler class (https://github.com/dic-iit/bipedal-locomotion-framework/pull/234)
 - Implement `PerceptionFeatures` library and implement `ArucoDetector`. (https://github.com/dic-iit/bipedal-locomotion-framework/pull/159)
 - Implement FixedBaseDynamics class (https://github.com/dic-iit/bipedal-locomotion-framework/pull/242)
-
+- Implement Block, BlockRunner and SharedResource classes in System component (https://github.com/dic-iit/bipedal-locomotion-framework/pull/256)
 
 ### Changed
 - Move all the Contacts related classes in Contacts component (https://github.com/dic-iit/bipedal-locomotion-framework/pull/204)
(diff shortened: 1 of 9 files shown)


S-Dafarra:
To be honest, I think that ``Block`` is a bit too similar to ``Advanceable``. What would be the advantage of having these two things separate?

GiulioRomualdi:
The main difference is that Block has input advance no

S-Dafarra:
> The main difference is that Block has input advance no

So why not add that method also to ``Advanceable``?

GiulioRomualdi:
Because we should change an huge  amount of code in the repo, furthermore some advanceables do not have inputs (sensorbridge) or have multiple inputs (planners).

S-Dafarra:
I think there might be easier options. What about
```c++
template<typename Input, typename Output>
class Block: public Advanceable<Output>
```
It is really a pity to start duplicating code at such early stages

GiulioRomualdi:
I try  to do that, even if I've have to change some part of the code of the PR that I'm already using 😢 

Notice that having multiple template argument is no so scalable for this kind of classes. If we want to add another template argument we should change all the classes while using CRPT we should just change the macro and the base class. 

By the way an advanceable is almost everything. While I seen a block only as a something usable with the block runner.

GiulioRomualdi:
Discussing with @traversaro and @S-Dafarra we reach the following points:
1. Block should be the Advance. However for the time being I don't have enough time to change advanceable and all the child classes
2. Since I need the Block/BlockRunner for a final application I will create a separate repo where I test the block along with a walking application. Once the code will become stable we may think to port it in bipedal.

Issue #285, opened 2021-04-26
Discuss the drawback of using `const IParametersHandler` in the Advanceable class


GiulioRomualdi:
In #267 I added
```cpp
bool initialize(std::shared_ptr<const BipedalLocomotion::ParametersHandler::IParametersHandler>)
```
method.

The signature of the method is in theory fine since a block should not change the handler in the initialization phase, however, in practice, this creates several issues.

For instance, in [`centroidal-mpc-walking`](https://github.com/GiulioRomualdi/centroidal-mpc-walking/blob/bb0f42b822f3afc41d5dec12e2e5b9bfd6bad633/src/centroidal-mpc-walking/src/WholeBodyQPBlock.cpp#L268) I tried to instantiate the `LeggedOdometry` however I'm not able to do it because the` LeggedOdometry::initialize()` wants a `non-const` `IParameterHandler`.
https://github.com/dic-iit/bipedal-locomotion-framework/blob/4ad31d8d62ffa0e3da5b7fbef77a945124c7bb4e/src/Estimators/include/BipedalLocomotion/FloatingBaseEstimators/FloatingBaseEstimator.h#L160-L161

Here two possible solutions can be implemented:
1. Change the signature of 
    https://github.com/dic-iit/bipedal-locomotion-framework/blob/4ad31d8d62ffa0e3da5b7fbef77a945124c7bb4e/src/Estimators/include/BipedalLocomotion/FloatingBaseEstimators/FloatingBaseEstimator.h#L160-L161
   Into `bool initialize(std::weak_ptr<const BipedalLocomotion::ParametersHandler::IParametersHandler> handler, std::shared_ptr<iDynTree::KinDynComputations> kindyn); ` 
2. Implement the copy constructor/copy assignment operator for `IParameterHandler`.

This first solution is temporary since we may need to expand the content of a ParameterHandler and if the object is `const` this is not possible. 
On the other hand, I don't know how to implement the second solution. Since there is no way to get the type of a stored parameter is almost impossible to copy the content of a parameter handler into another one. (e.g. I don't know how to copy the content of a `YarpParameterHandler` in a `StdParameterHandler` and vice-versa)

cc @prashanthr05 @S-Dafarra and @traversaro

S-Dafarra:
I would go for the first solution. I would prefer if the user of the parameters handler does not modify anything.

GiulioRomualdi:
[...]

PR #288, opened 2021-04-26
Implement clone method in ParametersHandler classes

GiulioRomualdi:
Follow up of https://github.com/dic-iit/bipedal-locomotion-framework/issues/285#issuecomment-826923766

### DIFF
--- a/src/ParametersHandler/tests/ParametersHandlerYarpTest.cpp
+++ b/src/ParametersHandler/tests/ParametersHandlerYarpTest.cpp
@@ -204,4 +204,52 @@ TEST_CASE("Get parameters")
         }
     }
 
+    SECTION("Clone")
+    {
+        IParametersHandler::shared_ptr setGroup = std::make_shared<YarpImplementation>();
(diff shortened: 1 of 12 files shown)


src/ParametersHandler/tests/ParametersHandlerYarpTest.cpp
```diff
@@ -204,4 +204,52 @@ TEST_CASE("Get parameters")
         }
     }
 
+    SECTION("Clone")
+    {
+        IParametersHandler::shared_ptr setGroup = std::make_shared<YarpImplementation>();
```
S-Dafarra, line 209: I think you forgot to clone here.

GiulioRomualdi: I call clone at line 217 :) 
```cpp
        auto newHandler = parameterHandler->clone();
```

S-Dafarra: Sorry I completely missed that :grin:

S-Dafarra (approved):

PR #301, opened 2021-05-08
Upload missing documentattion file

GiulioRomualdi:
#300 was merged without the main page of the documentation. This PR adds it.


traversaro (approved):
We should make sure to remember to keep the README and this in sync.

PR #323, opened 2021-06-07
Add `foot_take_off_acceleration` and `foot_take_off_velocity` parameters in the `SwingFootPlanner` class

GiulioRomualdi:
This PR introduces `foot_take_off_acceleration` and `foot_take_off_velocity` parameters in the `SwingFootPlanner` class.

I also remove `std::cerr` in favour if `log()->error(...)`

### DIFF
--- a/src/Planners/src/SwingFootPlanner.cpp
+++ b/src/Planners/src/SwingFootPlanner.cpp
@@ -43,6 +43,20 @@ bool SwingFootPlanner::initialize(std::weak_ptr<const ParametersHandler::IParame
         return false;
     }
 
+    double footTakeOffVelocity;
+    if (!ptr->getParameter("foot_take_off_velocity", footTakeOffVelocity))
+    {
+        log()->error("{} Unable to initialize the foot take-off velocity.", logPrefix);
+        return false;
+    }
+
+    double footTakeOffAcceleration;
+    if (!ptr->getParameter("foot_take_off_acceleration", footTakeOffAcceleration))
+    {
+        log()->error("{} Unable to initialize the foot take-off acceleration.", logPrefix);
+        return false;
+    }
+
     double footLandingVelocity;
     if (!ptr->getParameter("foot_landing_velocity", footLandingVelocity))
     {
@@ -58,7 +72,7 @@ bool SwingFootPlanner::initialize(std::weak_ptr<const ParametersHandler::IParame
     }
 
     // check the parameters passed to the planner
-    bool ok = (m_dT > 0) && ((m_footApexTime > 0) && (m_footApexTime < 1));
+    const bool ok = (m_dT > 0) && ((m_footApexTime > 0) && (m_footApexTime < 1));
     if (!ok)
     {
         log()->error("{} The sampling time should be a positive number and the foot_apex_time must "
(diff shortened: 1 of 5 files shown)


src/Planners/src/SwingFootPlanner.cpp
```diff
@@ -164,9 +178,11 @@ bool SwingFootPlanner::updateSE3Traj()
 
 bool SwingFootPlanner::advance()
 {
+    constexpr auto logPrefix = "[SwingFootPlanner::advance]";
+
     // update the time taking into account the limits
-    m_currentTrajectoryTime
-        = std::min(m_dT + m_currentTrajectoryTime, m_contactList.lastContact()->deactivationTime);
+    m_currentTrajectoryTime = std::min(m_dT + m_currentTrajectoryTime, //
```
S-Dafarra, line 184:
```suggestion
    m_currentTrajectoryTime = std::min(m_dT + m_currentTrajectoryTime,
```

src/Planners/src/SwingFootPlanner.cpp
```diff
@@ -212,16 +228,17 @@ bool SwingFootPlanner::advance()
...
-            std::cerr << "[SwingFootPlanner::advance] Unable to set the knots for the planar "
-                         "planner."
-                      << std::endl;
+            log()->error("{} Unable to set the knots for the planar planner.", logPrefix);
             return false;
         }
 
-        const double footHeightViaPointPos = (m_currentContactPtr->pose.translation()(2) +
-                                              nextContactPtr->pose.translation()(2)) / 2.0 + m_stepHeight;
+        const double footHeightViaPointPos = (m_currentContactPtr->pose.translation()(2) //
```
S-Dafarra, line 235:
```suggestion
        const double footHeightViaPointPos = (m_currentContactPtr->pose.translation()(2)
```

src/Planners/src/SwingFootPlanner.cpp
```diff
@@ -212,16 +228,17 @@ bool SwingFootPlanner::advance()
...
+                                              + nextContactPtr->pose.translation()(2))
+                                                 / 2.0
+                                             + m_stepHeight;
 
-        const double footHeightViaPointTime = m_footApexTime * T + m_currentContactPtr->deactivationTime;
+        const double footHeightViaPointTime = m_footApexTime * T //
```
S-Dafarra, line 240:
```suggestion
        const double footHeightViaPointTime = m_footApexTime * T
```

S-Dafarra (approved):
I noticed that you are adding some ``//`` before ending a line. Not sure if it is wanted

GiulioRomualdi:
Hi @S-Dafarra `// ` are useful to trick clang format to have more readable code

S-Dafarra:
> Hi @S-Dafarra `// ` are useful to trick clang format to have more readable code
> 
> I discovered it from @diegoferigo

Ok, then ignore the comments above

paolo-viceconte (approved):

diegoferigo:
> Hi @S-Dafarra `// ` are useful to trick clang format to have more readable code
> 
> I discovered it from @diegoferigo

clang-format uses a cost-based scheme in edge cases to select the less worst case between options. For how blf is configured, a long function names with string arguments gets broken weirdly, and I mean that the string is split in two parts instead of going all together to a new line. This is quite odd and I do not exclude that the configuration has either something off or something that can be optimized. This is not the case of the current PR, tho it could be related.

diegoferigo:
I didn't review this PR, but next time could we try to make these changes backward compatible e.g. by using default values? I just realized that all the code using the `SwingFootPlanner` has a partial configuration now.

PR #338, opened 2021-06-14
Add common Python files to gitignore

diegoferigo:


diegoferigo:
ping @GiulioRomualdi @S-Dafarra

S-Dafarra:
Really sorry for the delay, I missed it (twice). Overall looks good to me. My only concern is that the entries are much more, many of them are not directly related to python. Would it be possible to have those folders generated in a subfolder, from which it would be easier to ignore?

diegoferigo:
The new content is a direct copy-paste from the [official gitignores](https://github.com/github/gitignore) that Github uses for Python projects. What kind of problems do you think could happen? We're already using this ignores in many C++ / Python projects. If in the future there would be a clash with a new file / folder (quite uncommon), it would be trivial to spot the problem since the files would not show up in `git status`.

S-Dafarra:
> The new content is a direct copy-paste from the [official gitignores](https://github.com/github/gitignore) that Github uses for Python projects. What kind of problems do you think could happen? We're already using this ignores in many C++ / Python projects. If in the future there would be a clash with a new file / folder (quite uncommon), it would be trivial to spot the problem since the files would not show up in `git status`.

That's also a good point. Ok for me then.

S-Dafarra (approved):

diegoferigo:
Awesome, thanks! Resolved the conflict with the new online interface. It's the first time I use it, very helpful :)

GiulioRomualdi:
The CI failure is not related to this PR. please check here https://github.com/dic-iit/bipedal-locomotion-framework/issues/353

I think we can merge it

PR #340, opened 2021-06-15
Add COM Kalman Filter fusing kinematics and IMU measurements

prashanthr05:
During some walking experiments, it was observed that the DCM velocity (Divergent Component of Motion) suffered from sudden spikes causing the walking controller to fail. This quantity is computed using the Center of Mass velocity which in turn in measured using robot momenta using kinematics. This boils down to computing link velocities using the joint velocity measurements which are usually noisy.  
We may improve the CoM velocity estimation by either avoiding to use the joint velocities through a fusion of forward kinematics and IMU based fusion or do a complete fusion, assuming a simplified model for COM evolution and that the IMU lies close to the CoM of the robot in most cases. 
This is the approach taken in the work "Center of Mass Estimator for Humanoids and its Application in Modelling Error Compensation, Fall Detection and Prevention, Xinjilefu et al".

In this PR, we implement this estimator for estimating the center of mass position, velocity and offset by fusing IMU and kinematics measurements with a constant COM height LIPM model.

cc @GiulioRomualdi


GiulioRomualdi:
Hi @prashanthr05, first of all, thank you for the contribution 😄 . 

Is the PR ready to be reviewed?

prashanthr05:
> Hi @prashanthr05, first of all, thank you for the contribution  .
> 
> Is the PR ready to be reviewed?

A suitable test is pending.

PR #344, opened 2021-06-21
Implement cubic spline class

GiulioRomualdi:
This PR introduces the `Spline` class that is an abstract class. `QuinticSpline` and `CubicSpline` inherit from `Spline`. 

I finally modify the SwingFoot planner, so the user can choose between the two spline implementations.


src/Planners/src/CubicSpline.cpp
```diff
@@ -0,0 +1,768 @@
...
+    {
+        Eigen::VectorXd position; /**< Knot position */
+        Eigen::VectorXd velocity; /**< first derivative of spline computed at t@knot */
+        Eigen::VectorXd acceleration; /**< second derivative of spline computed at t@knot */
+
+        double timeInstant; /**< Knot time (it is an absolute time ) */
+    };
+
+    /**
+     * Description of a 5-th polynomial. It contains the coefficients, the initial/final knots and
```
paolo-viceconte, line 42: Another 5-th order polynomial to be replaced with 3-rd

src/Planners/src/CubicSpline.cpp
```diff
@@ -0,0 +1,768 @@
+/**
+ * @file CubicSpline.cpp
+ * @authors Giulio Romualdi
+ * @copyright 2020 Istituto Italiano di Tecnologia (IIT). This software may be modified and
```
paolo-viceconte, line 4: Another 2020 to be replaced here

src/Planners/src/CubicSpline.cpp
```diff
@@ -0,0 +1,768 @@
...
+
+CubicSpline::CubicSpline()
+{
+    m_pimpl = std::make_unique<Impl>();
+}
+
+CubicSpline::~CubicSpline() = default;
+
+bool CubicSpline::setInitialConditions(Eigen::Ref<const Eigen::VectorXd> velocity,
+                                       Eigen::Ref<const Eigen::VectorXd> /**acceleration*/)
```
paolo-viceconte, line 183: Not sure whether the second parameter is left on purpose here. If so, how is it exploited?

src/Planners/src/CubicSpline.cpp
```diff
@@ -0,0 +1,768 @@
...
+    m_pimpl->currentTrajectory.acceleration = Eigen::VectorXd::Zero(velocity.size());
+
+    // The initial conditions changed. The coefficients are outdated.
+    m_pimpl->areCoefficientsComputed = false;
+
+    return true;
+}
+
+bool CubicSpline::setFinalConditions(Eigen::Ref<const Eigen::VectorXd> velocity,
+                                     Eigen::Ref<const Eigen::VectorXd> /**acceleration*/)
```
paolo-viceconte, line 204: Same as above

paolo-viceconte (approved):
It looks good to me, I just spotted a few potential fixes in the documentation. I will try to exploit this class soon, thanks @GiulioRomualdi!

src/Planners/src/CubicSpline.cpp
GiulioRomualdi: `acceleration` is not used however is required to keep the same interface between the two splines

PR #359, opened 2021-07-07
Stream motor state, joint torques and PID state in SensorBridge

isorrentino:
This PR is meant to add the possibility to log motor encoders, current sensors, WBD torque estimates, motor PWMs, PID references.

### DIFF
--- a/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
+++ b/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
@@ -212,11 +212,9 @@ public:
      * Get all joints' positions in radians
      * @param[out] parameter all joints' position in radians
      * @param[out] receiveTimeInSeconds time at which the measurement was received
-     *
      * @warning the size is decided at the configuration and remains fixed,
      * and internal checks must be done at the implementation level by the Derived class.
      * This means that the user must pass a resized argument "jointPositions" to this method
-     *
      * @return true/false in case of success/failure
      */
     bool getJointPositions(Eigen::Ref<Eigen::VectorXd> jointPositions,
@@ -237,11 +235,9 @@ public:
      * Get all joints' velocities in rad/s
      * @param[out] parameter all joints' velocities in radians per second
      * @param[out] receiveTimeInSeconds time at which the measurement was received
-     *
      * @warning the size is decided at the configuration and remains fixed,
      * and internal checks must be done at the implementation level by the Derived class.
      * This means that the user must pass a resized argument "jointVelocties" to this method
-     *
      * @return true/false in case of success/failure
      */
     bool getJointVelocities(Eigen::Ref<Eigen::VectorXd> jointVelocties,
@@ -356,11 +352,9 @@ public:
      * Get all motors' currents in ampere
      * @param[out] motorCurrents all motors' current in ampere
      * @param[out] receiveTimeInSeconds time at which the measurement was received
-     *
      * @warning the size is decided at the configuration and remains fixed,
      * and internal checks must be done at the implementation level by the Derived class.
      * This means that the user must pass a resized argument "motorCurrents" to this method
-     *
      * @return true/false in case of success/failure
      */
     bool getMotorCurrents(Eigen::Ref<Eigen::VectorXd> motorCurrents,
@@ -381,20 +375,18 @@ public:
      * Get all motors' PWMs
      * @param[out] motorPWMs all motors' PWM in ampere
      * @param[out] receiveTimeInSeconds time at which the measurement was received
-     *
      * @warning the size is decided at the configuration and remains fixed,
      * and internal checks must be done at the implementation level by the Derived class.
-     * This means that the user must pass a resized argument "motorPWMs" to this method
-     *
+     * This means that the user must pass a resized argument "motorPWMs" to this methodW
      * @return true/false in case of success/failure
      */
     bool getMotorPWMs(Eigen::Ref<Eigen::VectorXd> motorPWMs,
                            OptionalDoubleRef receiveTimeInSeconds = {}) final;
 
     /**
-     * Get motor torque in NewtonMeter
+     * Get motor torque in Nm
      * @param[in] jointName name of the joint
-     * @param[out] jointTorque motor torque in NewtonMeter
+     * @param[out] jointTorque motor torque in Nm
      * @param[out] receiveTimeInSeconds time at which the measurement was received
      * @return true/false in case of success/failure
      */
@@ -403,14 +395,12 @@ public:
                           OptionalDoubleRef receiveTimeInSeconds = {}) final;
 
     /**
-     * Get all motors' torques in NewtonMeter
-     * @param[out] jointTorques all motors' torque in NewtonMeter
-     * @param[out] receiveTimeInSeconds time at which the measurement was received
-     *
+     * Get all motors' torques in Nm
+     * @param[out] jointTorques all motors' torque in Nm
+     * @param[out] receiveTimeInSeconds time at which the measurement was receivedW
      * @warning the size is decided at the configuration and remains fixed,
      * and internal checks must be done at the implementation level by the Derived class.
      * This means that the user must pass a resized argument "jointTorques" to this method
-     *
      * @return true/false in case of success/failure
      */
     bool getJointTorques(Eigen::Ref<Eigen::VectorXd> jointTorques,
@@ -431,11 +421,9 @@ public:
      * Get all pid positions in rad
      * @param[out] pidPositions all pid positions in radians
      * @param[out] receiveTimeInSeconds time at which the measurement was received
-     *
      * @warning the size is decided at the configuration and remains fixed,
      * and internal checks must be done at the implementation level by the Derived class.
-     * This means that the user must pass a resized argument "pidPositions" to this method
-     *
+     * This means that the user must pass a resized argument "pidPositions" to this methodW
      * @return true/false in case of success/failure
      */
     virtual bool getPidPositions(Eigen::Ref<Eigen::VectorXd> pidPositions,
@@ -456,11 +444,9 @@ public:
      * Get all pid position errors in rad
      * @param[out] pidPositionErrors all pid position errors in radians
      * @param[out] receiveTimeInSeconds time at which the measurement was received
-     *
      * @warning the size is decided at the configuration and remains fixed,
      * and internal checks must be done at the implementation level by the Derived class.
      * This means that the user must pass a resized argument "pidPositionErrors" to this method
-     *
      * @return true/false in case of success/failure
      */
     virtual bool getPidPositionErrors(Eigen::Ref<Eigen::VectorXd> pidPositionErrors,
@@ -481,11 +467,9 @@ public:
      * Get all motors' positions in rad
      * @param[out] parameter all motors' position in radians
      * @param[out] receiveTimeInSeconds time at which the measurement was received
-     *
      * @warning the size is decided at the configuration and remains fixed,
      * and internal checks must be done at the implementation level by the Derived class.
      * This means that the user must pass a resized argument "motorPositions" to this method
-     *
      * @return true/false in case of success/failure
      */
     virtual bool getMotorPositions(Eigen::Ref<Eigen::VectorXd> motorPositions,
@@ -506,11 +490,9 @@ public:
      * Get all motors' velocities in rad/s
      * @param[out] parameter all motors' velocities in radians per second
      * @param[out] receiveTimeInSeconds time at which the measurement was received
-     *
      * @warning the size is decided at the configuration and remains fixed,
      * and internal checks must be done at the implementation level by the Derived class.
      * This means that the user must pass a resized argument "motorVelocties" to this method
-     *
      * @return true/false in case of success/failure
      */
     virtual bool getMotorVelocities(Eigen::Ref<Eigen::VectorXd> motorVelocties,
--- a/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
+++ b/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
@@ -375,7 +375,7 @@ public:
      * @param[out] receiveTimeInSeconds time at which the measurement was received
      * @warning the size is decided at the configuration and remains fixed,
      * and internal checks must be done at the implementation level by the Derived class.
-     * This means that the user must pass a resized argument "motorPWMs" to this methodW
+     * This means that the user must pass a resized argument "motorPWMs" to this method
      * @return true/false in case of success/failure
      */
     bool getMotorPWMs(Eigen::Ref<Eigen::VectorXd> motorPWMs,
--- a/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
+++ b/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
@@ -361,6 +361,31 @@ public:
     bool getMotorCurrents(Eigen::Ref<Eigen::VectorXd> motorCurrents,
                            OptionalDoubleRef receiveTimeInSeconds = {}) final;
 
+    /**
+     * Get motor torque in NewtonMeter
+     * @param[in] jointName name of the joint
+     * @param[out] motorTorque motor current in NewtonMeter
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     * @return true/false in case of success/failure
+     */
+    bool getMotorTorque(const std::string& jointName,
+                          double& motorTorque,
+                          OptionalDoubleRef receiveTimeInSeconds = {}) final;
+
+    /**
+     * Get all motors' torques in NewtonMeter
+     * @param[out] motorTorques all motors' torque in NewtonMeter
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     *
+     * @warning the size is decided at the configuration and remains fixed,
+     * and internal checks must be done at the implementation level by the Derived class.
+     * This means that the user must pass a resized argument "motorTorques" to this method
+     *
+     * @return true/false in case of success/failure
+     */
+    bool getMotorTorques(Eigen::Ref<Eigen::VectorXd> motorTorques,
+                           OptionalDoubleRef receiveTimeInSeconds = {}) final;
+
 private:
     /** Private implementation */
     struct Impl;
--- a/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
+++ b/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
@@ -60,8 +60,6 @@ namespace RobotInterface
  * |                            |stream_pids                      | boolean           |Flag to activate the attachment to remapped control boards for pids reading      |
  * |                            |stream_motor_states              | boolean           |Flag to activate the attachment to remapped control boards for motor states reading      |
  * |                            |stream_motor_PWM                 | boolean           |Flag to activate the attachment to remapped control boards for PWM reading      |
- * |                            |stream_joint_torques             | boolean           |Flag to activate the attachment to remapped control boards for joint torques reading      |
- * |                            |stream_current_sensors           | boolean           |Flag to activate the attachment to remapped control boards for current sensors reading      |
  * |RemoteControlBoardRemapper  |                                 |                   |Expects only one remapped remotecontrolboard device attached to it, if there multiple remote control boards, then  use a remapper to create a single remotecontrolboard |
  * |                            |joints_list                      | vector of strings |This parameter is **optional**. The joints list used to open the remote control board remapper. If the list is not passed, the order of the joint stored in the PolyDriver is used       |
  * |InertialSensors             |                                 |                   |Expects IMU to be opened as genericSensorClient devices communicating through the inertial server and other inertials as a part multiple analog sensors remapper ("multipleanalogsensorsremapper") |
--- a/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
+++ b/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
@@ -52,11 +52,16 @@ namespace RobotInterface
  * |     Group                  |         Parameter               | Type              |                   Description                   |
  * |:--------------------------:|:-------------------------------:|:-----------------:|:---------------------------------------------- :|
  * |                            |check_for_nan                    | boolean           |flag to activate checking for NANs in the incoming measurement buffers, not applicable for images|
- * |                            |stream_joint_states              | boolean           |Flag to activate the attachment to IMU sensor devices       |
+ * |                            |stream_joint_states              | boolean           |Flag to activate the attachment to remapped control boards for joint states reading     |
  * |                            |stream_inertials                 | boolean           |Flag to activate the attachment to IMU sensor devices       |
  * |                            |stream_cartesian_wrenches        | boolean           |Flag to activate the attachment to Cartesian wrench related devices       |
  * |                            |stream_forcetorque_sensors       | boolean           |Flag to activate the attachment to six axis FT sensor devices       |
  * |                            |stream_cameras                   | boolean           |Flag to activate the attachment to Cameras devices       |
+ * |                            |stream_pids                      | boolean           |Flag to activate the attachment to remapped control boards for pids reading      |
+ * |                            |stream_motor_states              | boolean           |Flag to activate the attachment to remapped control boards for motor states reading      |
+ * |                            |stream_motor_PWM                 | boolean           |Flag to activate the attachment to remapped control boards for PWM reading      |
+ * |                            |stream_joint_torques             | boolean           |Flag to activate the attachment to remapped control boards for joint torques reading      |
+ * |                            |stream_current_sensors           | boolean           |Flag to activate the attachment to remapped control boards for current sensors reading      |
  * |RemoteControlBoardRemapper  |                                 |                   |Expects only one remapped remotecontrolboard device attached to it, if there multiple remote control boards, then  use a remapper to create a single remotecontrolboard |
  * |                            |joints_list                      | vector of strings |This parameter is **optional**. The joints list used to open the remote control board remapper. If the list is not passed, the order of the joint stored in the PolyDriver is used       |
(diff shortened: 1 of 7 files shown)


src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
```diff
@@ -52,11 +52,16 @@ namespace RobotInterface
...
  * |                            |stream_cartesian_wrenches        | boolean           |Flag to activate the attachment to Cartesian wrench related devices       |
  * |                            |stream_forcetorque_sensors       | boolean           |Flag to activate the attachment to six axis FT sensor devices       |
  * |                            |stream_cameras                   | boolean           |Flag to activate the attachment to Cameras devices       |
+ * |                            |stream_pids                      | boolean           |Flag to activate the attachment to remapped control boards for pids reading      |
+ * |                            |stream_motor_states              | boolean           |Flag to activate the attachment to remapped control boards for motor states reading      |
+ * |                            |stream_motor_PWM                 | boolean           |Flag to activate the attachment to remapped control boards for PWM reading      |
+ * |                            |stream_joint_torques             | boolean           |Flag to activate the attachment to remapped control boards for joint torques reading      |
```
GiulioRomualdi, line 63: Why this one is not inside `stream_joint_states`

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
```diff
@@ -52,11 +52,16 @@ namespace RobotInterface
...
+ * |                            |stream_current_sensors           | boolean           |Flag to activate the attachment to remapped control boards for current sensors reading      |
```
GiulioRomualdi, line 64: And this inside `stream_motor_states`

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
```diff
@@ -361,6 +366,156 @@ class YarpSensorBridge : public ISensorBridge,
...
+    /**
+     * Get motor PWM
+     * @param[in] jointName name of the joint
+     * @param[out] motorPWM motor PWM
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     * @return true/false in case of success/failure
+     */
+    bool getMotorPWM(const std::string& jointName,
+                          double& motorPWM,
+                          OptionalDoubleRef receiveTimeInSeconds = {}) final;
```
GiulioRomualdi, line 378: Could you please indent the code as specified in the `.clang-format` file?

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
```diff
@@ -361,6 +366,156 @@ class YarpSensorBridge : public ISensorBridge,
...
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     *
+     * @warning the size is decided at the configuration and remains fixed,
+     * and internal checks must be done at the implementation level by the Derived class.
+     * This means that the user must pass a resized argument "motorPWMs" to this method
+     *
+     * @return true/false in case of success/failure
+     */
+    bool getMotorPWMs(Eigen::Ref<Eigen::VectorXd> motorPWMs,
+                           OptionalDoubleRef receiveTimeInSeconds = {}) final;
```
GiulioRomualdi, line 392: As above 😄

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
```diff
@@ -361,6 +366,156 @@ class YarpSensorBridge : public ISensorBridge,
...
+     */
+    bool getMotorPWM(const std::string& jointName,
+                          double& motorPWM,
+                          OptionalDoubleRef receiveTimeInSeconds = {}) final;
+
+    /**
+     * Get all motors' PWMs
+     * @param[out] motorPWMs all motors' PWM in ampere
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     *
```
GiulioRomualdi, line 384: I think you can remove this empty line

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
```diff
@@ -361,6 +366,156 @@ class YarpSensorBridge : public ISensorBridge,
...
+     * and internal checks must be done at the implementation level by the Derived class.
+     * This means that the user must pass a resized argument "motorPWMs" to this method
+     *
+     * @return true/false in case of success/failure
+     */
+    bool getMotorPWMs(Eigen::Ref<Eigen::VectorXd> motorPWMs,
+                           OptionalDoubleRef receiveTimeInSeconds = {}) final;
+
+    /**
+     * Get motor torque in NewtonMeter
```
GiulioRomualdi, line 395:
```suggestion
     * Get motor torque in Nm
```

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
```diff
@@ -361,6 +366,156 @@ class YarpSensorBridge : public ISensorBridge,
...
+     * @param[in] jointName name of the joint
+     * @param[out] jointTorque motor torque in NewtonMeter
```
GiulioRomualdi, line 397:
```suggestion
     * @param[out] jointTorque motor torque in Nm
```

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
```diff
@@ -361,6 +366,156 @@ class YarpSensorBridge : public ISensorBridge,
...
+     */
+    bool getJointTorque(const std::string& jointName,
+                          double& jointTorque,
+                          OptionalDoubleRef receiveTimeInSeconds = {}) final;
+
+    /**
+     * Get all motors' torques in NewtonMeter
+     * @param[out] jointTorques all motors' torque in NewtonMeter
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     *
```
GiulioRomualdi, line 409: As above 😸

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
```diff
@@ -23,7 +23,11 @@
 #include <yarp/dev/IAnalogSensor.h>
 #include <yarp/dev/ICurrentControl.h>
 #include <yarp/dev/IGenericSensor.h>
+#include <yarp/dev/ITorqueControl.h>
 #include <yarp/dev/MultipleAnalogSensorsInterfaces.h>
+#include <yarp/dev/IMotorEncoders.h>
+#include <yarp/dev/IPidControl.h>
+#include <yarp/dev/IAmplifierControl.h>
```
GiulioRomualdi, line 30: They are not in alphabetic order 😺. You can use `clang-format` for that

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
```diff
@@ -1318,6 +1556,234 @@ struct YarpSensorBridge::Impl
...
+        {
+            if (nanExistsInVec(controlBoardRemapperMeasures.motorPWMsUnordered,
+                               logPrefix,
+                               "MotorPWMs"))
+            {
+                return false;
+            }
+        }
+
+        controlBoardRemapperMeasures.motorPWMs.noalias()
```
GiulioRomualdi, line 1683: Is this valid also for coupled joints?

src/RobotInterface/YarpImplementation/src/YarpSensorBridge.cpp
```diff
@@ -475,3 +545,213 @@ bool YarpSensorBridge::getMotorCurrents(Eigen::Ref<Eigen::VectorXd> motorCurrent
...
+        receiveTimeInSeconds.value().get()
+            = m_pimpl->controlBoardRemapperMeasures.receivedTimeInSeconds;
+
+    return true;
+}
+
+bool YarpSensorBridge::getPidPositions(Eigen::Ref<Eigen::VectorXd> pidPositions,
+                                       OptionalDoubleRef receiveTimeInSeconds)
+{
+
```
GiulioRomualdi, line 644:
```suggestion
```

src/RobotInterface/YarpImplementation/src/YarpSensorBridge.cpp
```diff
@@ -475,3 +545,213 @@ bool YarpSensorBridge::getMotorCurrents(Eigen::Ref<Eigen::VectorXd> motorCurrent
...
+        receiveTimeInSeconds.value().get()
+            = m_pimpl->controlBoardRemapperMeasures.receivedTimeInSeconds;
+
+    return true;
+}
+
+bool YarpSensorBridge::getPidPositionErrors(Eigen::Ref<Eigen::VectorXd> pidPositionErrors,
+                                            OptionalDoubleRef receiveTimeInSeconds)
+{
+
```
GiulioRomualdi, line 679:
```suggestion
```

src/RobotInterface/YarpImplementation/src/YarpSensorBridge.cpp
```diff
@@ -475,3 +545,213 @@ bool YarpSensorBridge::getMotorCurrents(Eigen::Ref<Eigen::VectorXd> motorCurrent
...
+        receiveTimeInSeconds.value().get()
+            = m_pimpl->controlBoardRemapperMeasures.receivedTimeInSeconds;
+
+    return true;
+}
+
+bool YarpSensorBridge::getMotorPositions(Eigen::Ref<Eigen::VectorXd> motorPositions,
+                                         OptionalDoubleRef receiveTimeInSeconds)
+{
+
```
GiulioRomualdi, line 714:
```suggestion
```

src/RobotInterface/YarpImplementation/src/YarpSensorBridge.cpp
```diff
@@ -475,3 +545,213 @@ bool YarpSensorBridge::getMotorCurrents(Eigen::Ref<Eigen::VectorXd> motorCurrent
...
+        receiveTimeInSeconds.value().get()
+            = m_pimpl->controlBoardRemapperMeasures.receivedTimeInSeconds;
+
+    return true;
+}
+
+bool YarpSensorBridge::getMotorVelocities(Eigen::Ref<Eigen::VectorXd> motorVelocties,
+                                          OptionalDoubleRef receiveTimeInSeconds)
+{
+
```
GiulioRomualdi, line 749:
```suggestion
```

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -365,6 +370,192 @@ class ISensorBridge
...
+    /**
+     * Get motor PWM
+     * @param[in] jointName name of the joint
+     * @param[out] motorPWM motor PWM
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     * @return true/false in case of success/failure
+     */
+    virtual bool getMotorPWM(const std::string& jointName,
+                                  double& motorPWM,
+                                  OptionalDoubleRef receiveTimeInSeconds = {})
```
GiulioRomualdi, line 383: `clang-format` 😄

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -365,6 +370,192 @@ class ISensorBridge
...
+    {
+        return false;
+    };
+
+    /**
+     * Get all motors' PWM
+     * @param[out] motorPWMs all motors' PWM
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     *
```
GiulioRomualdi, line 392:
```suggestion
```

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -365,6 +370,192 @@ class ISensorBridge
...
+                                  OptionalDoubleRef receiveTimeInSeconds = {})
+    {
+        return false;
+    };
+
+    /**
+     * Get all joints' torque in NewtonMeter
+     * @param[out] jointTorques all motors' torque in NewtonMeter
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     *
```
GiulioRomualdi, line 423:
```suggestion
```

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -365,6 +370,192 @@ class ISensorBridge
...
+     * @warning the size is decided at the configuration and remains fixed,
+     * and internal checks must be done at the implementation level by the Derived class.
+     * This means that the user must pass a resized argument "jointTorques" to this method
+     *
```
GiulioRomualdi, line 427:
```suggestion
```

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -365,6 +370,192 @@ class ISensorBridge
...
+                                  OptionalDoubleRef receiveTimeInSeconds = {})
+    {
+        return false;
+    };
+
+    /**
+     * Get all pid positions in rad
+     * @param[out] pidPositions all pid positions in radians
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     *
```
GiulioRomualdi, line 454:
```suggestion
```

src/RobotInterface/include/BipedalLocomotion/RobotInterface/ISensorBridge.h
```diff
@@ -365,6 +370,192 @@ class ISensorBridge
...
+     * @warning the size is decided at the configuration and remains fixed,
+     * and internal checks must be done at the implementation level by the Derived class.
+     * This means that the user must pass a resized argument "pidPositions" to this method
+     *
```
GiulioRomualdi, line 458:
```suggestion
```

GiulioRomualdi (commented):
Some minor comments. Could you also update the CHANGELOG?

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
isorrentino: I thought of keeping it separate because the device used to read the measurements is different.

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
isorrentino: > I thought of keeping it separate because the device used to read the measurements is different.

Same here.

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
isorrentino: Actually I'm not sure about it. I saw something similar in 
https://github.com/robotology/yarp-telemetry/blob/7e2e551bbd1de58cd048c4aab6d4cf8b422a3b13/src/telemetryDeviceDumper/TelemetryDeviceDumper.cpp#L569-L572
but they call it `jointPWM` instead of `motorPWM`.
The documentation is not very clear to me https://github.com/robotology/yarp/blob/2a0eacdf9b2e6f9be04cd2ebcbe097208b758838/src/libYARP_dev/src/yarp/dev/IAmplifierControl.h#L142-L148

prashanthr05:
General comment from a first pass,

I don't think individual attach methods for current sensors and motor sensors are necessary as implemented in this PR.
We can add only a few lines in 
https://github.com/dic-iit/bipedal-locomotion-framework/blob/dba32d34b46ccd4400349fc2e8f7a447d980e232/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h#L849-L858 

to view the corresponding interfaces according to the desired configuration.

isorrentino:
> General comment from a first pass,
> 
> I don't think individual attach methods for current sensors and motor sensors are necessary as implemented in this PR.
> We can add only a few lines in
> https://github.com/dic-iit/bipedal-locomotion-framework/blob/dba32d34b46ccd4400349fc2e8f7a447d980e232/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h#L849-L858
> 
> to view the corresponding interfaces according to the desired configuration.
> 
> Because, all of these interfaces, if available are supposed to be read from the same device which is the attached Remote Control Board Remapper.

@prashanthr05 thanks for the suggestion, I modified the code accordingly.

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
```diff
@@ -870,6 +871,11 @@ struct YarpSensorBridge::Impl
         {
             ok = ok && devList[devIdx]->poly->view(controlBoardRemapperInterfaces.axis);
             ok = ok && devList[devIdx]->poly->view(controlBoardRemapperInterfaces.encoders);
+            ok = ok && devList[devIdx]->poly->view(controlBoardRemapperInterfaces.currsensors);
+            ok = ok && devList[devIdx]->poly->view(controlBoardRemapperInterfaces.amp);
+            ok = ok && devList[devIdx]->poly->view(controlBoardRemapperInterfaces.motorEncoders);
```
prashanthr05, line 876:
``` cpp
if (metaData.bridgeOptions.isMotorEncodersEnabled)
{
    ok = ok && devList[devIdx]->poly->view(controlBoardRemapperInterfaces.motorEncoders);
}
```

and similar to other non-default entries?

src/RobotInterface/YarpImplementation/src/YarpSensorBridge.cpp
```diff
@@ -52,6 +52,76 @@ bool YarpSensorBridge::initialize(std::weak_ptr<const IParametersHandler> handle
...
+                                   m_pimpl->metaData,
+                                   m_pimpl->metaData.bridgeOptions.isCurrentSensorsEnabled);
+    if (!ret)
+    {
+        log()->info("{} Skipping the configuration of configureRemoteControlBoardRemapper. "
+                    "YarpSensorBridge "
+                    "will not stream relevant measures.",
+                    logPrefix);
+    }
+
```
prashanthr05, line 124: I was wondering if there might be a cleaner way to do this, since within the `configureRemoteControlBoardRemapper` function, we clear the joint lists in the `metaData` everytime and read the same parameter from the configuration file.

Instead, we could just use part of the `subConfigLoader` logic somewhere enable non-default streams provided by the remote control board remapper? 

Open to discussion cc @GiulioRomualdi @S-Dafarra

S-Dafarra: To be honest, I did not understand the problem

prashanthr05: The problem is quite minor.

[SubConfigLoader](https://github.com/dic-iit/bipedal-locomotion-framework/blob/71e37a822c9c7aa1bf23a91eeabceb6f66ba3165/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h#L193-L228) is a helper function that checks for a `enable_stream` parameter (for example `enable_joint_states`) from the configuration file and then uses a `configureSomeTypeOfSensors` function pointer to load other associated parameters to this stream (in this case `configureRemoteControlBoardRemapper`).

Here, within the [`configureRemoteControlBoardRemapper`](https://github.com/dic-iit/bipedal-locomotion-framework/blob/71e37a822c9c7aa1bf23a91eeabceb6f66ba3165/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h#L230-L266), the configuration that is being read is `jointsLists` only.  However, since now there other optional joints related quantities expected to be obtained from the same RemoteControlBoardRemapper, repeating this `subConfigLoader` calls basically tries to clear the `jointsList` and re-read the `jointsList` everytime. Maybe, it's ok to do so.. since we are only in the configuration step. But maybe, we could do better.

S-Dafarra: To be honest, after reading the comments and the code several times, I had a hard time understanding the whole logic (clearly I am a bit slow :grin:). I wonder if it is possible to simplify the whole ``SubConfigLoader``/``configureRemoteControlBoardRemapper`` logic. If I understood well, the purpose of ``subConfigLoader`` was to reduce the use of repeated patterns. On the other hand, it is called repeatedly itself, while making the code much less readable. If it also has side effects, then it's becoming a little problematic.

Here is how I see it. 
Either we leave it as it is. Reading the same parameter a lot of times from the configuration parameters is not a drama. The complexity to read a configuration parameter is most probably linear in the ``yarp`` case. So it is not completely free, a bit of a waste, but not problematic.

Another possibility could be that instead of the ``SubConfigLoader``/``configureRemoteControlBoardRemapper`` logic we directly call a method that takes care of reading all the parameters, setting the relevant variables. 

What do you think?

isorrentino: > To be honest, after reading the comments and the code several times, I had a hard time understanding the whole logic (clearly I am a bit slow grin). I wonder if it is possible to simplify the whole SubConfigLoader/configureRemoteControlBoardRemapper logic. If I understood well, the purpose of subConfigLoader was to reduce the use of repeated patterns. On the other hand, it is called repeatedly itself, while making the code much less readable. If it also has side effects, then it's becoming a little problematic.

To be honest, I took much more time than expected for this PR because of the same problem. I had a hard time understanding too (but consider that I'm super slow).

prashanthr05: > On the other hand, it is called repeatedly itself, while making the code much less readable. If it also has side effects, then it's becoming a little problematic.

I suppose the problem of readability arises only with the `RemoteControlBoardRemapper` since it exposes a variety of interfaces while being just one type of device. 

> Either we leave it as it is. Reading the same parameter a lot of times from the configuration parameters is not a drama. The complexity to read a configuration parameter is most probably linear in the yarp case. So it is not completely free, a bit of a waste, but not problematic.

This should be fine, I guess.

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
GiulioRomualdi: I'm not super convinced about this choice. Indeed even if the interfaces are different the joint torque is _de facto_ a joint state. We do not have `steram_joint_position` or `stream_joint_velocity`

isorrentino: Ok, you convinced me. I'll change it.

src/RobotInterface/YarpImplementation/src/YarpSensorBridge.cpp
isorrentino: Should we handle this in a different PR?
@prashanthr05 @GiulioRomualdi @S-Dafarra

prashanthr05: I'd say, since you're going to handle the suggested change in https://github.com/dic-iit/bipedal-locomotion-framework/pull/359#discussion_r667845184, it would be easier to handle this change as well.

But, It's anyways not blocking the prescribed functionality of this PR. So you can decide. :)

isorrentino: I changed a bit the logic of `configureRemoteControlBoardRemapper`, `attachRemappedRemoteControlBoard` and `readControlBoardInterface` in `YarpSensorBridgeImpl.h`. I hope it is clear and it works fine. The commit is https://github.com/dic-iit/bipedal-locomotion-framework/pull/359/commits/a5f3bd21e818d1629b0ef4e482f63beb588559da.
@prashanthr05 @GiulioRomualdi @S-Dafarra

prashanthr05: The changes looks good. And also the check on the number of joints before reading the joint list parameter was a simple and effective solution.

prashanthr05 (approved):

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
```diff
@@ -351,16 +350,152 @@ class YarpSensorBridge : public ISensorBridge,
...
+     * Get all pid positions in rad
+     * @param[out] pidPositions all pid positions in radians
+     * @param[out] receiveTimeInSeconds time at which the measurement was received
+     * @warning the size is decided at the configuration and remains fixed,
+     * and internal checks must be done at the implementation level by the Derived class.
+     * This means that the user must pass a resized argument "pidPositions" to this methodW
+     * @return true/false in case of success/failure
+     */
+    virtual bool getPidPositions(Eigen::Ref<Eigen::VectorXd> pidPositions,
+                                  OptionalDoubleRef receiveTimeInSeconds = {}) final;
```
S-Dafarra, line 428: What is the position of a PID?

GiulioRomualdi (approved):

GiulioRomualdi:
Thank you @prashanthr05 for the review and @isorrentino for iterating 😃 

Merging!

S-Dafarra (commented):
Whoops I arrived late. My only comment was that it is not clear what is the PidPosition. Is the reference or the measured quantity? Also, a PID does not have a "position" semantically speaking

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridge.h
GiulioRomualdi: Opening a new issue for this

GiulioRomualdi: https://github.com/dic-iit/bipedal-locomotion-framework/issues/367

PR #365, opened 2021-07-12
Rename macOS-latest in macos-latest in GitHub CI

GiulioRomualdi:
As suggested by @S-Dafarra in https://github.com/dic-iit/bipedal-locomotion-framework/pull/364#issuecomment-878394900

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
     strategy:
       matrix:
         build_type: [Release, Debug]
-        os: [ubuntu-20.04, windows-latest, macOS-latest]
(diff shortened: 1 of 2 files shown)


S-Dafarra:
I noticed that the conda actions instead were running fine. So probably it was not case sensitive and it was a glitch?

S-Dafarra (approved):

GiulioRomualdi:
> I noticed that the conda actions instead were running fine. So probably it was not case sensitive and it was a glitch?

Maybe. However, as you pointed out `macos-latest` is the right name :)

traversaro (approved):

PR #369, opened 2021-07-14
Implement FeasibleContactWrenchTask for TSID component

GiulioRomualdi:
This PR implements a task for the TSID component ensuring a feasible contract  wrench

### DIFF
--- a/src/TSID/include/BipedalLocomotion/TSID/FeasibleContactWrenchTask.h
+++ b/src/TSID/include/BipedalLocomotion/TSID/FeasibleContactWrenchTask.h
@@ -1,0 +1,132 @@
+/**
+ * @file FeasibleContactWrenchTask.h
+ * @authors Giulio Romualdi
+ * @copyright 2021 Istituto Italiano di Tecnologia (IIT). This software may be modified and
+ * distributed under the terms of the GNU Lesser General Public License v2.1 or any later version.
+ */
+
+#ifndef BIPEDAL_LOCOMOTION_TSID_FEASIBLE_CONTACT_WRENCH_TASK_H
+#define BIPEDAL_LOCOMOTION_TSID_FEASIBLE_CONTACT_WRENCH_TASK_H
+
+#include <memory>
+
+#include <manif/manif.h>
+
+#include <BipedalLocomotion/TSID/TSIDLinearTask.h>
+#include <BipedalLocomotion/Math/ContactWrenchCone.h>
+
+
+#include <iDynTree/KinDynComputations.h>
+
+
+namespace BipedalLocomotion
+{
+namespace TSID
+{
+
+/**
+ * FeasibleContactWrenchTask is a concrete implementation of the TSIDLinearTask. Please use this
+ * element if you want to ensure that the contact wrench satisfies the constraints implemented in
+ * BipedalLocomotion::Math::ContactWrenchCone. Differently from the
+ * BipedalLocomotion::Math::ContactWrenchCone class, FeasibleContactWrenchTask requires also the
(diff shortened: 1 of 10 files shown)


src/TSID/include/BipedalLocomotion/TSID/FeasibleContactWrenchTask.h
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Enable the task. I.e. a contact wrench different from zero is allowed.
+     */
+    void enable();
+
+    /**
+     * Disable the task. I.e. the contact wrench is forced to be equal to zero.
+     */
+    void disable();
```
S-Dafarra, line 108: There is some potential confusion on whether it is the task to be disabled (hence the wrench is not constrained), or if the wrench cannot be used to control the robot. A possibility could be to allow setting directly the maximum normal component. Another possibility could be to have a method like ``setContactState`` with an enum. Otherwise also a different naming could do the trick, although I did not find any good alternatives myself.

GiulioRomualdi: I like `setContactState`! the enum could be `Active` `Inactive`, what do you think?

GiulioRomualdi: In theory the enum could be defined in https://github.com/dic-iit/bipedal-locomotion-framework/blob/d81b390683127a451669b0e09ff06da199f9490a/src/Contacts/include/BipedalLocomotion/Contacts/Contact.h#L19

S-Dafarra: That would be nice. On the other hand, we should try to be coherent and use that enum also inside the class. Right now it is a boolean https://github.com/dic-iit/bipedal-locomotion-framework/blob/d81b390683127a451669b0e09ff06da199f9490a/src/Contacts/include/BipedalLocomotion/Contacts/Contact.h#L102
and only in the ``EstimatedContact``.

For the time being, we could simply add the boolean and start using it in separate PRs.

GiulioRomualdi: In 9b8662b I implemented the function we discussed.

GiulioRomualdi:
@S-Dafarra let mIf I can merge the PR 😃

src/TSID/include/BipedalLocomotion/TSID/FeasibleContactWrenchTask.h
```diff
@@ -98,14 +95,10 @@ class FeasibleContactWrenchTask : public TSIDLinearTask
...
+     * Set if the contact is active or not.
+     * @param isActive true if the contact is active, false otherwise.
      */
-    void enable();
-
-    /**
-     * Disable the task. I.e. the contact wrench is forced to be equal to zero.
-     */
-    void disable();
+    void isContactActive(bool isActive);
```
S-Dafarra, line 101: I realized later that ``isContactActive`` is more in the form of a question, so I would expect a return value. Maybe ``setContactActive`` would suggest more that you are changing an internal state. Does it make sense?

GiulioRomualdi: Fixed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/369/commits/1277a6264d7dc26fc4812a7152ac8b071905bbab

S-Dafarra (approved):

S-Dafarra:
None of the jobs passed, did you check the output?

PR #378, opened 2021-07-23
YarpSensorBridge: add checks in control board getter methods

### LINKED ISSUE #376: SensorBridge crashes if a data is required and the stream is disabled
I was with @isorrentino and we noticed that `getMotorsCurrent` segs fault if `stream_motor_state` is not set in the config file. 

The function should return false and not crashes. 
The same logic is applied in all the methods of SensorBridge 

cc @prashanthr05

prashanthr05:
Fixes #376


GiulioRomualdi (approved):
Thank you @prashanthr05

GiulioRomualdi:
Windows is failing but don't worry is a known issue of vcpkg #374

PR #379, opened 2021-07-23
Add the possibility to use manif objects in the ForwardEuler integrator

GiulioRomualdi:
This PR introduces the possibility to use ForwardEuler integrator with manif objects. Thanks to this I removed the  Baumgarte stabilization in FloatingBaseSystemKinematics class


Giulero (approved):

traversaro (approved):

prashanthr05:
In the need for extension towards also right trivialization,

Maybe you could allow the user to set the trivialization as done [here](https://github.com/dic-iit/element_floating-base-estimation/blob/315e2c601bc02f8d14e298684e476da70a2dbe76/code/KinDynFusion/src/system/include/KinDynFusion/System/LieGroupDynamicalSystem.h#L33).

and use the corresponding integration accordingly, as done in an example of [RKMK4  Liegroup integrator, here](https://github.com/dic-iit/element_floating-base-estimation/blob/315e2c601bc02f8d14e298684e476da70a2dbe76/code/KinDynFusion/src/system/include/KinDynFusion/System/RKMK4Explicit.tpp#L25-L34).

PR #385, opened 2021-08-03
Allow changing the log verbosity

diegoferigo:
I exposed all the spdlogs levels to a new blf enum at the C++ level. Then, the blf enum is exposed to Python.


GiulioRomualdi (approved):
Thank you @diegoferigo

PR #390, opened 2021-08-11
Fix the `TOML` and `YARP` implementation of the parameters handler when a `std::vector<bool>` is passed to the `setParameter()` 

GiulioRomualdi:
This fixes the problem spotted by @traversaro in #381. 
It implements what is explained in https://github.com/dic-iit/bipedal-locomotion-framework/pull/381#issuecomment-896756059

- [x] Enable TOMLPlusPlus in CI


GiulioRomualdi:
In #391, I'm checking the tomlplusplus installation

S-Dafarra (approved):

PR #393, opened 2021-08-12
Release 0.3.0

GiulioRomualdi:


S-Dafarra (approved):

S-Dafarra:
Btw, I think that in time, the mention to the ``Framework`` target in the README completely disappeared, although it is still created https://github.com/dic-iit/bipedal-locomotion-framework/blob/master/CMakeLists.txt#L91-L93. Does it makes sense to bring it back?

PR #397, opened 2021-08-19
Add virtual destructor to Advanceable class

prashanthr05:
This allows appropriate dynamic allocation of memory for the classes derived from the `Advanceable` class.


GiulioRomualdi:
Thank you @prashanthr05. The descrutor is actually needed  Notice that this invalidate the rule of zero. I think it is necessary to define also the other 4 functions. (Not supersure)

prashanthr05:
> Thank you @prashanthr05. The descrutor is actually needed Notice that this invalidate the rule of zero. I think it is necessary to define also the other 4 functions. (Not supersure)
> 
> https://stackoverflow.com/questions/26039907/does-rule-of-three-five-apply-to-inheritance-and-virtual-destructors

I was thinking in any case, it was going to be set to the default cases and so it's safe to avoid them. 
This might however become super-important if we have a destructor definition and dealing with pointers (for the necessary deep-copy).

But in any case, it's not going to do us any harm add a few lines. I will do it.

prashanthr05:
> > Thank you @prashanthr05. The descrutor is actually needed Notice that this invalidate the rule of zero. I think it is necessary to define also the other 4 functions. (Not supersure)
> > https://stackoverflow.com/questions/26039907/does-rule-of-three-five-apply-to-inheritance-and-virtual-destructors
> 
> I was thinking in any case, it was going to be set to the default cases and so it's safe to avoid them.
> This might however become super-important if we have a destructor definition and dealing with pointers (for the necessary deep-copy).
> 
> ~But in any case, it's not going to do us any harm add a few lines.~ I went back and brushed up on rule of zero/three/five. There's a reason why it is a "rule". I will add these lines.

Done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/397/commits/f859754d6543358f6cc043a16831ca8932ae5841.

GiulioRomualdi:
The failure of the test should be related to this PR

GiulioRomualdi:
I rebased the PR on top of master. If the CI passes I will merge it

src/System/include/BipedalLocomotion/System/Advanceable.h
```diff
@@ -36,6 +36,13 @@ template <class _Input, class _Output> class BipedalLocomotion::System::Advancea
     using Input = _Input;
     using Output = _Output;
 
+    Advanceable() = default;
```
GiulioRomualdi, line 39: This shouldn't be required

prashanthr05: Removed in https://github.com/dic-iit/bipedal-locomotion-framework/pull/397/commits/1fcb4d038bd83d56b79ed3116034fce9109f03f6

prashanthr05: Looks like it's needed afterall!

diegoferigo:
I'm a bit skeptical about the declaration of the copy / move operators on the base `Advanceable` class. I'm not sure what would be the consequences of creating a new System, downcast it to `Advanceable`, and then try to copy the downcasted object. What kind of data would be copied? The entire original class? Just the members of the `Advanceable` class? Do you have any idea about how would happen? With these doubts, I'd rather keep copy / move disabled. cc @traversaro

prashanthr05:
> I'm a bit skeptical about the declaration of the copy / move operators on the base `Advanceable` class. I'm not sure what would be the consequences of creating a new System, downcast it to `Advanceable`, and then try to copy the downcasted object. What kind of data would be copied? The entire original class? Just the members of the `Advanceable` class? Do you have any idea about how would happen? With these doubts, I'd rather keep copy / move disabled. cc @traversaro

We could maybe disable the copy constructors, but move maybe not. Since, for using the AdvanceableRunner, we create an advanceable and move the ownership to the runner.

diegoferigo:
> > I'm a bit skeptical about the declaration of the copy / move operators on the base `Advanceable` class. I'm not sure what would be the consequences of creating a new System, downcast it to `Advanceable`, and then try to copy the downcasted object. What kind of data would be copied? The entire original class? Just the members of the `Advanceable` class? Do you have any idea about how would happen? With these doubts, I'd rather keep copy / move disabled. cc @traversaro
> 
> We could maybe disable the copy constructors, but move maybe not. Since, for using the AdvanceableRunner, we create an advanceable and move the ownership to the runner.

If there's already the need to do that, feel free to test the behavior and leave the PR with the new definitions. My main point is that I'm not 100% sure of what would actually be copied. In the same SO question you linked, a [reply below](https://stackoverflow.com/a/26040064) the accepted mentions of possible problems related to [_slicing_](https://schneide.blog/2015/07/06/object-slicing-breaking-polymorphic-objects-in-c/).

S-Dafarra:
On this, I agree with @diegoferigo. I would go for a ``default`` destructor and deleted copy/move constructors/assignement (hence the rule of five would be satisfied). The derived class can define its own if needed

GiulioRomualdi:
However the C++ documentation (https://en.cppreference.com/w/cpp/language/rule_of_three
) says the following about obeying the rule of 5 for polymorphism. 

> However, this makes the class prone to slicing, which is why polymorphic classes often define copy as deleted (see [C.67: A polymorphic class should suppress public copy/move](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c67-a-polymorphic-class-should-suppress-public-copymove) in C++ Core Guidelines), which leads to the following generic wording for the Rule of Five:
> 
> [C.21: If you define or =delete any default operation, define or =delete them all.](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all)

S-Dafarra (approved):

GiulioRomualdi:
Let's wait for the CI 🤞🏻

GiulioRomualdi:
macOS failure due to https://github.com/ami-iit/bipedal-locomotion-framework/issues/685

PR #398, opened 2021-08-20
Expose const ref of resource and mutex in SharedResource

prashanthr05:
This will allow the proper usage of `SharedResource` with smart pointers across multiple threads.

### DIFF
--- a/src/System/include/BipedalLocomotion/System/SharedResource.h
+++ b/src/System/include/BipedalLocomotion/System/SharedResource.h
@@ -21,7 +21,9 @@ namespace System
  * shared resource you can use it only with shared_ptr. For this reason the only way to create a
  * shared resource is to use the SharedResource::create() static method. If want a visual
  * interpretation of shared resource you may imagine it as the wire connecting two blocks running in
- * two separate threads. The implementation of SharedResource is thread safe.
+ * two separate threads. The implementation of SharedResource is thread safe, except when getting the
+ * resource by reference or when the resource type is a smart pointer.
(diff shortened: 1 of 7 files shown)


src/System/include/BipedalLocomotion/System/SharedResource.h
```diff
@@ -44,6 +46,21 @@ template <class T> class SharedResource
...
+     * Get the resource using const reference.
+     * @note one must not forget to access the mutex and
+     *       perform necessary locking and unlocking
+     * @return the const reference of the object inside the shared resource.
+     */
+    const T& get() const;
+
+    /**
+     * Get the mutex by reference.
+     * @note one must be careful not to modify the mutex
```
S-Dafarra, line 59: > std::mutex is neither copyable nor movable.
From https://en.cppreference.com/w/cpp/thread/mutex

So in principle, there is no way to edit it, right?

S-Dafarra (approved):

src/System/include/BipedalLocomotion/System/SharedResource.h
prashanthr05: Great! Then, I will remove that line.

prashanthr05: Thanks, done in https://github.com/dic-iit/bipedal-locomotion-framework/pull/398/commits/7959a8d53b630076205238f2a2e5917366a5380a.

GiulioRomualdi:
I think this PR is too old and currently not required. I will close for the time being

PR #421, opened 2021-09-28
Change icubSim to icub in config file of blf-calibration-delta-updater application

isorrentino:
…r app


isorrentino:
It was already icub instead of icubSim. Closing.

PR #429, opened 2021-10-21
Add the possibility to set the name of each element of a variable stored in the variables handler

GiulioRomualdi:
As written in the title with this PR it is possible to set the label of each element of a variable stored in the handler. Thanks to this it will be easier to retrieve the index of the variable

The feature is already implemented also for the python bindings
```
In [1]: import bipedal_locomotion_framework.bindings as blf                                         
In [2]: a = blf.system.VariablesHandler()                                                           
In [3]: a.add_variable("joint_acceleration_with_names", 3, ["torso", "hip", "ankle"])               
In [4]: a.add_variable("joint_acceleration", 10)                                                    

In [5]: print(a)                                                                                    
joint_acceleration size: 10, offset: 3 elements name: joint_acceleration_0 joint_acceleration_1 joint_acceleration_2 joint_acceleration_3 joint_acceleration_4 joint_acceleration_5 joint_acceleration_6 joint_acceleration_7 joint_acceleration_8 joint_acceleration_9joint_acceleration_with_names size: 3, offset: 0 elements name: torso hip ankle
```

### DIFF
--- a/src/System/src/VariablesHandler.cpp
+++ b/src/System/src/VariablesHandler.cpp
@@ -121,9 +121,9 @@ bool VariablesHandler::initialize(std::weak_ptr<const IParametersHandler> handle
             }
         } else
         {
-            log()->info("{} The parameter {}_elements_name is not found. The default one is used",
-                        logPrefix,
-                        names[i]);
+            log()->debug("{} The parameter {}_elements_name is not found. The default one is used",
+                         logPrefix,
+                         names[i]);
 
             if (!this->addVariable(names[i], sizes[i]))
             {


src/System/include/BipedalLocomotion/System/VariablesHandler.h
```diff
@@ -19,21 +20,35 @@ namespace BipedalLocomotion
...
         bool isValid() const;
+
+        std::ptrdiff_t getElementIndex(const std::string& name) const;
+        std::ptrdiff_t getElementIndex(std::ptrdiff_t localIndex) const;
+
         static VariableDescription InvalidVariable();
+
+    private:
+        std::unordered_map<std::string, std::ptrdiff_t> elementsNameMap;
+        std::vector<std::string> elementsName;
```
S-Dafarra, line 49: I would consider adding the ``m_`` in front of the name since they are private.

src/System/src/VariablesHandler.cpp
```diff
@@ -72,22 +103,63 @@ bool VariablesHandler::initialize(std::weak_ptr<const IParametersHandler> handle
...
+            {
+                log()->error("{} Unable to add the variable named {} having a size equal to {}.",
+                             logPrefix,
+                             names[i],
+                             sizes[i]);
+                return false;
+            }
+        } else
+        {
+            log()->info("{} The parameter {}_elements_name is not found. The default one is used",
```
S-Dafarra, line 123: ``info`` probably would be too much verbose. ``debug``?

GiulioRomualdi:
By the way we should be carefull in the case a new variable is added in an already existing group. All the next existing variables offset must be shifted of the size of the new variable and the group containing that variable must be expanded

PR #456, opened 2021-11-17
Install texlive-core in gh-page workflow

GiulioRomualdi:
The current documentation is broken since `m.css` is not able to find latex 

https://github.com/ami-iit/bipedal-locomotion-framework/runs/4235942651?check_suite_focus=true#step:6:3407

This should fix it. Before merging it I will check the artifact updated by the action


GiulioRomualdi:
I don't know why but the action called `gh-page` is not running anymore 😭

PR #475, opened 2021-12-02
Add the possibility to change the weight in TSID/IK

### LINKED ISSUE #474: Make the weights editable in the IK/TSID
The weights are now loaded by the configuration file. The only thing missing to using blf-Ik  in the walking module is to add the possibility to change the cost function wights at runtime.

Here is what I propose:
1. keep the weight in the configuration file. This is the default weight that will be loaded by IK/TSID if the user does not specify a new one
2. Create a method `setWeight(std::string taskName, Vector weight)` in IK/TSID this will set the new weight
3. The user has to generate the trajectory of the weight.  He can use a minJerkTrajectory or something else. We can implement this class in blf but we have to decide where to put it (`math` component?) and give a name to the class

cc @traversaro @S-Dafarra @isorrentino

GiulioRomualdi:
This closes #474

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project are documented in this file.
 ## [Unreleased]
 ### Added
 - Add the reading of the orientation of the head IMU in `YarpRobotLoggerDevice` (https://github.com/ami-iit/bipedal-locomotion-framework/pull/471)
-
+- Add the possibility to change the weight in TSID/IK (https://github.com/ami-iit/bipedal-locomotion-framework/pull/475)
 ### Changed
 - Use yarp clock instead of system clock in `YarpRobotLoggerDevice` (https://github.com/ami-iit/bipedal-locomotion-framework/pull/473)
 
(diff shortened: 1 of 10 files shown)


traversaro:
@traversaro did you wanted to ask for a review?

GiulioRomualdi:
> @GiulioRomualdi did you wanted to ask for a review instead of assigning?

Ops 😃

S-Dafarra (commented):
The code looks good to me. What I find misleading is that ``weight`` is singular, but it is expecting a vector instead of a number. In general, I did not understand what is the effect of this weight? In a cost function we could have two weights
$$
w1 c(x)^T diag(w2) c(x) +  w1' c(x)'^T diag(w2') c(x)'.
$$
Which weight are we changing?

GiulioRomualdi:
Hi @S-Dafarra

The weight I'm referring to is the one associated with the norm. In TSID (same for the IK) each component of the cost function is a `l2` weighted norm of the form

image

where `W`, in our case is a diagonal matrix. In `setTaskWeight` the diagonal of the diagonal matrix is passed to the solver

S-Dafarra:
Ok I see. Basically, it is the w2 in https://github.com/ami-iit/bipedal-locomotion-framework/pull/475#pullrequestreview-822378830. Can we also set the w1? If not, then there is no ambiguity

GiulioRomualdi:
> Ok I see. Basically, it is the w2 in [#475 (review)](https://github.com/ami-iit/bipedal-locomotion-framework/pull/475#pullrequestreview-822378830). Can we also set the w1? If not, then there is no ambiguity

`w1` is always 1 in tsid/ik

S-Dafarra (approved):

PR #487, opened 2021-12-13
Use telemetry in YarpLoggerDevice

### LINKED ISSUE #189: Start using yarp-telemetry in the framework
[`yarp-telemetry`](https://github.com/robotology-playground/yarp-telemetry) is a brand new library that allows logging data in a fancy and simple way.

It may be worth starting to think about how to use it in our library. A possible example is to extend `advanceable` and add a log function to store the dataset.

These are the limitations that I see
- [x] `yarp-telemetry` cannot store indefinitely long dataset https://github.com/robotology-playground/yarp-telemetry/pull/55
- [x] Right now the interface requires a reference to a std::vector as input. https://github.com/robotology-playground/yarp-telemetry/issues/49 A possible solution is to use `iDynTree::span` and `iDynTree::MatrixView` 

cc @nicogene @traversaro @S-Dafarra @isorrentino @prashanthr05

GiulioRomualdi:
You can try to use it. It depends on https://github.com/robotology/yarp-telemetry/pull/146

This closes #189 

cc @isorrentino @S-Dafarra @traversaro

### DIFF
--- a/devices/YarpRobotLoggerDevice/app/robots/iCubGazeboV3/yarp-robot-logger.xml
+++ b/devices/YarpRobotLoggerDevice/app/robots/iCubGazeboV3/yarp-robot-logger.xml
@@ -12,6 +12,16 @@ GNU Lesser General Public License v2.1 or any later version. -->
     <param name="save_period">600.0</param>
   </group>
 
+  <group name="ExogenousSignals">
+    <param name="exogenous_inputs">("Walking")</param>
+
+    <group name="Walking">
+      <param name="port_name">"/yarp-robot-logger/exogenous_signals/walking:i"</param>
+      <param name="signal_name">"walking"</param>
+    </group>
+
+  </group>
+
   <group name="RobotSensorBridge">
     <param name="check_for_nan">false</param>
     <param name="stream_joint_states">true</param>
--- a/devices/YarpRobotLoggerDevice/app/robots/iCubGazeboV3/yarp-robot-logger.xml
+++ b/devices/YarpRobotLoggerDevice/app/robots/iCubGazeboV3/yarp-robot-logger.xml
@@ -8,6 +8,10 @@ GNU Lesser General Public License v2.1 or any later version. -->
   <param name="sampling_period_in_s">0.01</param>
   <param name="port_prefix">/yarp-robot-logger</param>
 
+  <group name="Telemetry">
+    <param name="save_period">600.0</param>
+  </group>
+
   <group name="RobotSensorBridge">
     <param name="check_for_nan">false</param>
     <param name="stream_joint_states">true</param>
--- a/devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp
+++ b/devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp
@@ -3,10 +3,14 @@
  * distributed under the terms of the GNU Lesser General Public License v2.1 or any later version.
  */
 
+#include <cstddef>
 #include <iomanip>
+#include <matioCpp/Span.h>
+#include <memory>
 #include <tuple>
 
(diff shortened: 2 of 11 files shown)


GiulioRomualdi:
@isorrentino @prashanthr05 the PR is ready to be reviewed

devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp
```diff
@@ -50,63 +61,152 @@ YarpRobotLoggerDevice::~YarpRobotLoggerDevice() = default;
...
+        return false;
+    }
+
+    // the telemetry will flush the content of its storage every config.save_period
+    // and this device runs every devicePeriod
+    // so the size of the telemetry buffer must be at least config.save_period / devicePeriod
+    // to be sure we are not going to lose data the buffer will be 10% longer
+    constexpr double percentage = 0.1;
+    config.n_samples = static_cast<int>(ceil((1 + percentage) //
+                                             * (config.save_period / devicePeriod)));
```
prashanthr05, line 164:
```suggestion
    config.n_samples = static_cast<int>(std::ceil((1 + percentage) //
                                             * (config.save_period / devicePeriod)));
```

available in <cmath>.

prashanthr05 (approved):
Minor suggesion.

devices/YarpRobotLoggerDevice/app/robots/iCubGazeboV3/yarp-robot-logger.xml
```diff
@@ -8,6 +8,20 @@ GNU Lesser General Public License v2.1 or any later version. -->
...
+    <param name="save_period">600.0</param>
+  </group>
+
+  <group name="ExogenousSignals">
+    <param name="exogenous_inputs">("Walking")</param>
+
+    <group name="Walking">
+      <param name="port_name">"/yarp-robot-logger/exogenous_signals/walking:i"</param>
+      <param name="signal_name">"walking"</param>
+    </group>
```
prashanthr05, line 21: @GiulioRomualdi I had a question regarding this. Why is this an input port?
Why don't we do any connection with an external port?
Could you explain a bit more about the usage here? I am sorry I missed this detail during the review.

prashanthr05: I was wondering, If there's an arbitrary port existing in the network, for example `/icub/xsens_inertial/analog:o` then is it possible to log data directly from that port through this exogenous signal connection? If yes, how do we do that?

prashanthr05: ping @GiulioRomualdi

GiulioRomualdi: Hi @prashanthr05 this is the port opened by the device this is the reason why it's an input port. The connection is left to the user. 
`/icub/xsens_inertial/analog:o` cannot be used as it is indeed the signal must be compliant with this format
 
https://github.com/ami-iit/bipedal-locomotion-framework/blob/5d1d6b81f498602ffc6c83e48e4d3c707a2e6cd5/src/YarpUtilities/thrifts/BipedalLocomotion/YarpUtilities/VectorsCollection.thrift#L3-L6

prashanthr05: Ah ok, so we might need a wrapper for the port `/icub/sens_inertial/analog:o`? Could you direct me to the link of how the `walking-controller` data is packed?

prashanthr05: Found it here: https://github.com/robotology/walking-controllers/pull/97.

PR #500, opened 2022-01-18
Implement python bindings for the classes that handle the clock

GiulioRomualdi:
This PR implements:
1. The python bindings for the Clock machinery and for the yarp clock

Furthermore, since the clock is implemented as a singleton, an error it will be returned if the user tries to change the clock type once the `clock()` has been already called once. Notice that `BipedalLocomotion::clock()` instantiate the singleton the first time is called (See Meyers' singleton implementation).

For instance, in python, we will have 
```
In [1]: import bipedal_locomotion_framework.bindings as blf

In [2]: blf.clock().now()
Out[2]: datetime.timedelta(days=19010, seconds=63353, microseconds=503687)

In [3]: blf.system.ClockBuilder.set_factory(blf.system.YarpClockFactory())
[2022-01-18 18:35:55.945] [thread: 66773] [blf] [error] [ClockBuilder::setFactory] The clock has been already called. Since the clock() returns a singleton it is not possible to set the factory anymore. Please call 'setFactory()' at the beginning of your application to avoid this problem.
Out[3]: False
```

```
In [1]: import bipedal_locomotion_framework.bindings as blf

In [2]: blf.system.ClockBuilder.set_factory(blf.system.YarpClockFactory())
Out[2]: True

In [3]: blf.system.ClockBuilder.set_factory(blf.system.YarpClockFactory())
Out[3]: True

In [4]: blf.clock().now()
Out[4]: datetime.timedelta(days=19010, seconds=63392, microseconds=416123)
```

Thanks to this PR it will be possible to easily synchronize yarp process also in python.

Last but not least the PR implements the `trampoline` pybind11 functionality to extend the clock class directly in python https://pybind11.readthedocs.io/en/stable/advanced/classes.html#combining-virtual-functions-and-inheritance. In other words  a python developer can implement its own clock by inheriting from the c++ binded  class 

cc @Giulero

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,11 @@ All notable changes to this project are documented in this file.
 
 ## [Unreleased]
 ### Added
+- Implement the python bindings for the clock machinery and for the yarp clock (https://github.com/ami-iit/bipedal-locomotion-framework/pull/500)
+
 ### Changed
+- An error it will be returned if the user tries to change the clock type once the `clock()` has been already called once (https://github.com/ami-iit/bipedal-locomotion-framework/pull/500)
+
 ### Fix
 
 ## [0.6.0] - 2022-12-10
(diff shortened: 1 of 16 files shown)


GiulioRomualdi:
The PR is ready to be reviewed

S-Dafarra (approved):

PR #507, opened 2022-02-04
Implement the `MultiStateWeightProvider`

GiulioRomualdi:
The `MultiStateWeightProvider` is a provider that will allow us to perform gain scheduling. The user should provide a configuration file containing the name and the weight for each state. Then the provider will interpolate among the states using a FirstOrder stable system. The settling time of the system should be set in the configuration file.

I also implemented the associated tests and the python bindings.

In order to improve the possibility to set a given weight provider in the `ILinearTaskSolver` can pass directly a weight provider


S-Dafarra:
> Before merging: I am not sure that WeightProvider is the right name of the class since it can be used also to perform gain scheduling for the weight of a controller and not just for the weight of the `ILinearTaskSolver`. For instance, it could be used to perform gain scheduling the CoM-ZMP controller.

Makes sense. What did you have in mind?

GiulioRomualdi:
`

> > Before merging: I am not sure that WeightProvider is the right name of the class since it can be used also to perform gain scheduling for the weight of a controller and not just for the weight of the `ILinearTaskSolver`. For instance, it could be used to perform gain scheduling the CoM-ZMP controller.
> 
> Makes sense. What did you have in mind?

`ControllerGainProvider`? mmmmh 😕 I was not able to find nothing

S-Dafarra:
> `ControllerGainProvider`? mmmmh confused I was not able to find nothing better

Thinking about it, now is a supersimple interface, with a simple ``getWeight()`` method. In the end, it can be seen as a simpler version of ``Advanceable``, but the IK never calls ``advance``. Similarly, the ``Advanceable`` class has the ``getOutput()``. So, what if we add a templated interface ``Output`` (or ``OutputSignal``) with only the ``getOutput`` virtual method. Then ``Advanceable`` could inherit from this ``Output``, while the IK tasks, or the controllers, will expect as input for the weight a shared pointer to an ``Output`` with a given type. In this way, the weight can be an output of any ``Advanceable`` you like. Does it make sense?

GiulioRomualdi:
Closing in favour of #555

PR #520, opened 2022-03-24
Implement a matlab function to migrate the dataset loaded with blf 0.6.0 to blf 0.7.0

GiulioRomualdi:


isorrentino (approved):

GiulioRomualdi:
> script because past log files will not be compatible with the new versions of the visualizer

The old log files are still working with the new visualizer but the joints indices are displayed instead of the joints names.

S-Dafarra (approved):

PR #538, opened 2022-06-17
Attempt to speedup the Ubuntu ci

GiulioRomualdi:
This PR moves the missing dependency compilation test into separate jobs. It creates a different workflow for a missing dependency 

With this PR
image

Without this PR
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -582,7 +582,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ${{ github.workspace }}/install/deps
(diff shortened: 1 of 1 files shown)


traversaro (approved):
Sorry, I had missed it. It is not great to duplicate the part on compilation from source of dependencies that is quite verbose, but it if gives a gain, why not.

By the way, I did not noticed that the check for missing dependencies is taking ~1 hour, that is a lot. In theory, enabling&disabling a dependency should not take so much time, as it should affect only the part that depend on that dependency, but probably there is something that gets regenerated and forces the recompilation of all the library?

PR #547, opened 2022-07-06
Add logging left upper leg FT and wrenches for iCubGazeboV3

isorrentino:


GiulioRomualdi:
Duplicated of https://github.com/ami-iit/bipedal-locomotion-framework/pull/546

PR #570, opened 2022-10-05
Automatic polling for exogenous signals

GiulioRomualdi:
Thanks to this PR the `YarpRobotLoggerDevice` will automatically connect to the exogenous ports. So no need to run it before the walking-controller

Moreover: 
- The remote control board will use udp as a carrier
- the readouts from the genericSensorClient (wrenches) are retrieved using udp
- the multipleanalogsensorsclient data are collected with fast_tcp
- the logging of the robot cameras is disabled
- The files are formatted


isorrentino (approved):

devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp
```diff
@@ -604,6 +652,80 @@ void YarpRobotLoggerDevice::unpackIMU(Eigen::Ref<const analog_sensor_t> signal,
...
+        }
+        for (const auto& [name, signal] : m_vectorSignals)
+        {
+            allPortsConnected = allPortsConnected && signal.connected;
+        }
+
+        if (!allPortsConnected)
+        {
+            // check for new messages
+            yarp::profiler::NetworkProfiler::getPortsList(yarpPorts);
```
S-Dafarra, line 690: Why do you need the full list of ports if you already know the name of the port to connect to?

GiulioRomualdi: to check if the port exists to establish the connection. maybe we can directly do the connection if not connected. I don't know if  we can improve the performances in this case

GiulioRomualdi: https://github.com/ami-iit/bipedal-locomotion-framework/pull/570/commits/05342243cffc707ee03c605ce8a7baa1b3328d69 should implement what you suggested

S-Dafarra: That's pretty neat!

S-Dafarra (approved):

PR #580, opened 2022-12-18
Use std::int64_t instead of long in TextLoggingEntry class

### LINKED ISSUE #579: Unable to compile YarpRobotLoggerDevice on Windows
As in https://github.com/conda-forge/bipedal-locomotion-framework-feedstock/pull/6 when I try to compile the device in windows I got the following error

image

I notice that commenting the following line, the compilation went fine 

https://github.com/ami-iit/bipedal-locomotion-framework/blob/c5112abf18e30817cf6c18f60632381eb38b8d4c/devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp#L901

msg is a 

https://github.com/ami-iit/bipedal-locomotion-framework/blob/c5112abf18e30817cf6c18f60632381eb38b8d4c/devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp#L877

and in theory matioCpp should be able to handle it thanks to 

https://github.com/ami-iit/bipedal-locomotion-framework/blob/c5112abf18e30817cf6c18f60632381eb38b8d4c/devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp#L43-L61

GiulioRomualdi:
This fixes #579 moreover thanks to this https://github.com/ami-iit/matio-cpp/pull/63 is no more required

### DIFF
--- a/devices/YarpRobotLoggerDevice/include/BipedalLocomotion/YarpTextLoggingUtilities.h
+++ b/devices/YarpRobotLoggerDevice/include/BipedalLocomotion/YarpTextLoggingUtilities.h
@@ -7,6 +7,7 @@
 #define BIPEDAL_LOCOMOTION_FRAMEWORK_YARP_ROBOT_LOGGER_DEVICE_YARP_TEXT_LOGGING_UTILITIES_H
 
 #include <string>
+#include <cstdint>
 
 #include <yarp/os/Bottle.h>
 
@@ -29,7 +30,7 @@ struct TextLoggingEntry
     std::string cmd;
     std::string args;
     int pid{0};
-    long thread_id{0};
+    std::int64_t thread_id{0};
     std::string component;
     std::string id;
     double systemtime{0.0};


traversaro (approved):

S-Dafarra (approved):

GiulioRomualdi:
macOS is not compiling because the code I wrote is not c++17 compliant (there seems that it uses a feature supported since c++20) see here: https://stackoverflow.com/questions/46114214/lambda-implicit-capture-fails-with-variable-declared-from-structured-binding

S-Dafarra:
May it be related by the fact that you are passing by reference ``cameraName`` and ``writer``, but they are local? Would it  make sense to simply pass them by copy? Or copy directly the iterator?

GiulioRomualdi:
https://github.com/ami-iit/bipedal-locomotion-framework/pull/580/commits/ead93eb884c8d296c9d9aa020dbac473b4c60c9f should fix the issue, lets wait for the CI

PR #596, opened 2023-01-31
Add a tutorial for the inverse kinematics

GiulioRomualdi:
This is a step forward toward https://github.com/ami-iit/bipedal-locomotion-framework/issues/501

cc @traversaro


S-Dafarra (approved):

S-Dafarra:
Actually, discussing with @Iaxama a few days ago, we noticed that when you have no task at level 0 and/or no task at level 1, the IK fails with a cryptic error about the non-zeros of the hessian/constrain matrix

PR #603, opened 2023-02-10
Implement the JointLimitsTask for the IK

### LINKED ISSUE #530: QPInverseKinematics and joint limit constraints?
Has been using QPInverseKinematics for a while. Usage is similar to IK/tests/QPInverseKinematicsTest.cpp. 

Say I define a CoM task with highpriority and JointTracking task with lowpriority. It happens a lot that CoM is reached but the resulting joint positions violate urdf joint limits.

Is it possible for the solver to take into account joint limit constraint as well? 

Thanks.

GiulioRomualdi:
This closes #530 

cc @shikui08

### DIFF
--- a/src/IK/src/JointLimitsTask.cpp
+++ b/src/IK/src/JointLimitsTask.cpp
@@ -1,0 +1,271 @@
+/**
+ * @file JointLimitsTask.cpp
+ * @authors Giulio Romualdi
+ * @copyright 2023 Istituto Italiano di Tecnologia (IIT). This software may be modified and
+ * distributed under the terms of the BSD-3-Clause license.
+ */
+
+#include <types.h>
+#include <vector>
+
+#include <OsqpEigen/Constants.hpp>
+
+#include <BipedalLocomotion/IK/JointLimitsTask.h>
+#include <BipedalLocomotion/TextLogging/Logger.h>
+
+#include <iDynTree/Core/EigenHelpers.h>
+#include <iDynTree/Model/Model.h>
+
+using namespace BipedalLocomotion::ParametersHandler;
+using namespace BipedalLocomotion::IK;
+
+bool JointLimitsTask::setKinDyn(std::shared_ptr<iDynTree::KinDynComputations> kinDyn)
+{
+    if ((kinDyn == nullptr) || (!kinDyn->isValid()))
+    {
+        log()->error("[JointLimitsTask::setKinDyn] Invalid kinDyn object.");
+        return false;
+    }
+
+    m_kinDyn = kinDyn;
+
+    // populate the limits with the one retrieved from the model
+    m_upperLimits = Eigen::VectorXd::Constant(m_kinDyn->getNrOfDegreesOfFreedom(), OsqpEigen::INFTY);
+    m_lowerLimits = Eigen::VectorXd::Constant(m_kinDyn->getNrOfDegreesOfFreedom(), -OsqpEigen::INFTY);
+
+    unsigned int jointsWithLimits{0};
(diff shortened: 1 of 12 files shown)


src/IK/src/JointLimitsTask.cpp
```diff
@@ -0,0 +1,266 @@
...
+        {
+            if (!joint->getPosLimits(dof,
+                                     m_lowerLimits[joint->getDOFsOffset() + dof],
+                                     m_upperLimits[joint->getDOFsOffset() + dof]))
+            {
+                continue;
+            } else
+            {
+                m_isLimitConsideredForAllJoints = true;
+            }
```
S-Dafarra, line 57: It seems that is enough that ``joint->getPosLimits`` returns true for only one joint in order to have ``m_isLimitConsideredForAllJoints = true``. Is this wanted?

src/IK/src/JointLimitsTask.cpp
```diff
@@ -0,0 +1,266 @@
...
+        return false;
+    }
+
+    // check if the
+    bool useModelLimits = true;
+    if (!ptr->getParameter("use_model_limits", useModelLimits))
+    {
+        log()->error("{} Error while retrieving the 'use_model_limits' parameter", errorPrefix);
+        return false;
+    }
```
S-Dafarra, line 167: If there is an error with a specific parameter, I would specify in the message the exact name of the parameter that triggered the error

src/IK/src/JointLimitsTask.cpp
GiulioRomualdi: Nope

S-Dafarra:
Thinking about it, does it make sense to have a test case in which we check that the joint limits are actually satisfied over time?

shikui08:
> This closes #530
> 
> cc @shikui08

Wow. Great work. Appreciate your efforts~!

src/IK/src/JointLimitsTask.cpp
GiulioRomualdi: Done 2865e2b6b4fc47582008579d5bee78e4e2877e39

src/IK/src/JointLimitsTask.cpp
GiulioRomualdi: Fixed here 2e8d668ed148b44c4f13dc2a42057d6f81b449fd thank you for noticing

GiulioRomualdi:
> Thinking about it, does it make sense to have a test case in which we check that the joint limits are actually satisfied over time?

This part is currently missing

S-Dafarra:
> > Thinking about it, does it make sense to have a test case in which we check that the joint limits are actually satisfied over time?
> 
> This part is currently missing

Are you planning to add it, or you prefer to avoid it?

GiulioRomualdi:
> Are you planning to add it, or you prefer to avoid it?

I'm trying to do it

GiulioRomualdi:
> > Are you planning to add it, or you prefer to avoid it?
> 
> I'm trying to do it

While writing the test I found a bug that I fixed in https://github.com/ami-iit/bipedal-locomotion-framework/pull/603/commits/4139b0de6403a370b0b871b2a70da3bc2f0f13ef

GiulioRomualdi:
In https://github.com/ami-iit/bipedal-locomotion-framework/pull/603/commits/9fea6654a47d4ca6b9b9ddcc3a9bd87705a63c1d I added a test where I check that the joint limit task is satisfied. I forced the joint to be

$$
q^- \le q \le q^+
$$

where I asked

$$
q^- = q^+ = q(t = 0)
$$

for some joints

S-Dafarra (approved):

PR #608, opened 2023-02-23
Fix attachRemappedRemoteControlBoard in YarpSensorBridge

GiulioRomualdi:
This fix is required when the the YarpSensorBridge is attached to a list of polydriver and the remote control board remapper is not the first one.

### DIFF
--- a/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
+++ b/src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
@@ -977,51 +977,74 @@ struct YarpSensorBridge::Impl
         // expects only one remotecontrolboard device attached to it, if found break!
         // if there multiple remote control boards, then  use a remapper to create a single
         // remotecontrolboard
-        bool ok{true};
+        bool okJointsSensor = !metaData.bridgeOptions.isJointSensorsEnabled;
+        bool okPWM = !metaData.bridgeOptions.isPWMControlEnabled;
+        bool okMotorsSensor = !metaData.bridgeOptions.isMotorSensorsEnabled;
+        bool okPID = !metaData.bridgeOptions.isPIDsEnabled;
+
         for (int devIdx = 0; devIdx < devList.size(); devIdx++)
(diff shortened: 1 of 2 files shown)


src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
```diff
@@ -977,51 +977,77 @@ struct YarpSensorBridge::Impl
...
+                     "PWM sensors: {} - {}.",
+                     logPrefix,
+                     isSensorFound(okJointsSensor),
+                     isSensorRequired(metaData.bridgeOptions.isJointSensorsEnabled),
+                     isSensorFound(okMotorsSensor),
+                     isSensorRequired(metaData.bridgeOptions.isMotorSensorsEnabled),
+                     isSensorFound(okPID),
+                     isSensorRequired(metaData.bridgeOptions.isPIDsEnabled),
+                     isSensorFound(okPWM),
+                     isSensorRequired(metaData.bridgeOptions.isPWMControlEnabled));
```
S-Dafarra, line 1048: I suspect that this error message could be a bit misleading. Suppose that ``metaData.bridgeOptions.isJointSensorsEnabled = false``, we will have
```
Joint sensors: found - not required,
```
but actually we never checked if the joint sensor interfaces are actually found. In my opinion, it would be better to have only
```
Joint sensors: not required,
```

GiulioRomualdi: What do you think of https://github.com/ami-iit/bipedal-locomotion-framework/pull/608/commits/c8c6e0c23a5fc99c5a5112d300a069bbcb69cf79 ?

S-Dafarra (approved):

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpSensorBridgeImpl.h
S-Dafarra: Looks good!

PR #615, opened 2023-03-02
Add dynamic models composing the UKF process model

isorrentino:
This PR implements some state dynamics which can be used to build a process model for a UKF.

### DIFF
--- a/src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/Dynamics.h
+++ b/src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/Dynamics.h
@@ -1,0 +1,198 @@
+/**
+ * @file Dynamics.h
+ * @authors Ines Sorrentino
+ * @copyright 2023 Istituto Italiano di Tecnologia (IIT). This software may be modified and
+ * distributed under the terms of the BSD-3-Clause license.
+ */
+
+#ifndef BIPEDAL_LOCOMOTION_ESTIMATORS_DYNAMICS_H
+#define BIPEDAL_LOCOMOTION_ESTIMATORS_DYNAMICS_H
+
+#include <memory>
+#include <string>
+
+#include <Eigen/Dense>
+
+#include <BipedalLocomotion/System/Factory.h>
+#include <BipedalLocomotion/System/Advanceable.h>
+#include <BipedalLocomotion/ParametersHandler/IParametersHandler.h>
+#include <BipedalLocomotion/System/VariablesHandler.h>
+#include <BipedalLocomotion/RobotDynamicsEstimator/SubModel.h>
+#include <BipedalLocomotion/RobotDynamicsEstimator/SubModelKinDynWrapper.h>
+
+/**
+ * BLF_REGISTER_DYNAMICS is a macro that can be used to register a Dynamics. The key of
+ * the dynamics will be the stringified version of the Dynamics C++ Type
+ * @param _model the model of the dynamics
+ * @param _baseModel the base model from which the _model inherits.
+ */
+#define BLF_REGISTER_DYNAMICS(_model, _baseModel)               \
+    static std::shared_ptr<_baseModel> _model##FactoryBuilder() \
+{                                                               \
+    return std::make_shared<_model>();                          \
+    };                                                          \
+    static std::string _model##BuilderAutoRegHook               \
+    = ::BipedalLocomotion::Estimators::                         \
+    RobotDynamicsEstimator::DynamicsFactory::registerBuilder    \
+    (#_model, _model##FactoryBuilder);
+
+namespace BipedalLocomotion
+{
+namespace Estimators
+{
+namespace RobotDynamicsEstimator
+{
+
+/**
+ * @brief The UKFInput struct represents the input of the ukf needed to update the dynamics.
+ */
+struct UKFInput
+{
+    Eigen::VectorXd robotJointPositions; /**< Vector of joint positions. */
+    manif::SE3d robotBasePose; /**< Robot base position and orientation. */
+    manif::SE3d::Tangent robotBaseVelocity; /**< Robot base velocity. */
+    manif::SE3d::Tangent robotBaseAcceleration; /**< Robot base acceleration. */
+};
+
+/**
+ * UkfInputProvider describes the provider for the inputs of the ukf
+ */
+class UkfInputProvider : public System::Advanceable<UKFInput, UKFInput>
+{
+private:
+    UKFInput m_ukfInput;
+
+public:
+    /**
+     * Get the inputes
+     * @return A struct containing the inputs for the ukf
+     */
+    const UKFInput& getOutput() const override;
+
+    /**
+     * Set the state which represents the state of the provider.
+     * @param input is a struct containing the input of the ukf.
+     */
+    bool setInput(const UKFInput& input) override;
+
+    /**
+     * @brief Advance the internal state. This may change the value retrievable from getOutput().
+     * @return True if the advance is successfull.
+     */
+    bool advance() override;
+
+    /**
+     * @brief Determines the validity of the object retrieved with getOutput()
+     * @return True if the object is valid, false otherwise.
+     */
+    bool isOutputValid() const override;
+
+}; // class UkfProcessInputProvider
+
+/**
+ * Dynamics describes a generic dynamic model for a UKF state or measurement.
+ */
+class Dynamics
+{
+protected:
+    int m_size; /**< Size of the variable associate to the Dynamics object. */
+    Eigen::VectorXd m_updatedVariable; /**< Updated variable computed using the dynamic model. */
+    std::string m_description{"Generic Dynamics Element"}; /**< String describing the type of the dynamics */
+    std::string m_name; /**< Name of dynamics. */
+    Eigen::VectorXd m_covariances; /**< Vector of covariances. */
+    std::vector<std::string> m_elements = {}; /**< Elements composing the variable vector. */
+    std::string m_dynamicModel;
+    bool m_isInitialized{false}; /**< True if the dynamics has been initialized. */
+    System::VariablesHandler m_stateVariableHandler; /**< Variable handler describing the variables and the sizes in the ukf state/measurement vector. */
+    bool m_isStateVariableHandlerSet{false}; /**< True if setVariableHandler is called. */
+    UKFInput m_ukfInput;
+
+    /**
+      * Controls whether the variable handler contains the variables on which the dynamics depend.
+      * @return True in case all the dependencies are contained in the variable handler, false otherwise.
+      */
+    virtual bool checkStateVariableHandler();
+
+
+public:
+    /**
+     * Initialize the task.
+     * @param paramHandler pointer to the parameters handler.
+     * @return True in case of success, false otherwise.
+     */
+    virtual bool
+    initialize(std::weak_ptr<const ParametersHandler::IParametersHandler> paramHandler);
+
+    /**
+     * Finalize the Dynamics.
+     * @param handler object describing the variables in the vector for which all the dynamics are defined.
+     * @note You should call this method after you add ALL the dynamics to the variable handler.
+     * @return true in case of success, false otherwise.
+     */
+    virtual bool finalize(const System::VariablesHandler& handler);
+
+    /**
+     * Set the SubModelKinDynWrapper object.
+     * @param kinDynWrapper pointer to a SubModelKinDynWrapper object.
+     * @return True in case of success, false otherwise.
+     */
+    virtual bool setSubModels(const std::vector<SubModel>& subModelList, const std::vector<std::shared_ptr<SubModelKinDynWrapper>>& kinDynWrapperList);
+
+    /**
+     * Update the dynamics of the variable.
+     * @return True in case of success, false otherwise.
+     */
+    virtual bool update();
+
+    /**
+     * Get the next value m_updatedVariable.
+     * @return a const reference to the next variable value.
+     */
+    Eigen::Ref<const Eigen::VectorXd> getUpdatedVariable() const;
+
+    /**
+     * Get the size of the state.
+     * @return the size of the state.
+     */
+    int size() const;
+
+    /**
+     * Set the state of the ukf needed to update the dynamics.
+     * @param ukfState reference to the ukf state.
+     */
+    virtual void setState(const Eigen::Ref<const Eigen::VectorXd> ukfState) = 0;
+
+    /**
+     * Set a `UKFInput` object.
+     * @param ukfInput reference to the UKFInput struct.
+     */
+    virtual void setInput(const UKFInput & ukfInput) = 0;
+
+    /**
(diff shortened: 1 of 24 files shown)


GiulioRomualdi:
Can you rebase the PR on top of master and removing all the commits already merged in past PR?

isorrentino:
Hi @GiulioRomualdi @S-Dafarra I noticed that the measurement and state dynamics classes are exactly the same. I put the PR in draft in order to change the `StateDynamics` class name into `Dynamics` and apply all the needed modifications. I'll ping you as soon as the modifications are pushed and the PR is again ready to be reviewed.

isorrentino:
> Hi @GiulioRomualdi @S-Dafarra I noticed that the measurement and state dynamics classes are exactly the same. I put the PR in draft in order to change the `StateDynamics` class name into `Dynamics` and apply all the needed modifications. I'll ping you as soon as the modifications are pushed and the PR is again ready to be reviewed.

Changes took less than expected. Changing back to ready.

src/Estimators/tests/RobotDynamicsEstimator/UKFModelCreatorTest.cpp
```diff
@@ -0,0 +1,61 @@
...
+    REQUIRE(groupUKFMeasurement != nullptr);
+
+    // Model Creator
+
+    // State Model --> variable handler, lista_dinamiche_stati
+
+    // Measurement Model --> variable handler, lista_dinamiche_misure
+
+    // Crea un oggetto ukfmodelcreator(stateHandler, measurementHandler)
+    // crea variablehandler per stato e per misure
```
S-Dafarra, line 60: Leftover?

src/Estimators/tests/RobotDynamicsEstimator/FrictionTorqueDynamicsTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <BipedalLocomotion/RobotDynamicsEstimator/FrictionTorqueDynamics.h>
+
+using namespace BipedalLocomotion::Estimators::RobotDynamicsEstimator;
+using namespace BipedalLocomotion::ParametersHandler;
+using namespace BipedalLocomotion::System;
+
+double random_double()
+{
+    return 1.0 * ((double)rand() - RAND_MAX / 2) / ((double)RAND_MAX);
+}
```
S-Dafarra, line 28: ``catch`` should have dedicated methods to generate a random double https://github.com/catchorg/Catch2/blob/0a0ebf50034831718ffc63c6d12abb1fa469d6e9/docs/generators.md

Also, it is suggested to fix the seed, to make sure that the test is repeatable (meaning that the value is no more random, but at least you avoid cases where some test fails once in a while)

src/Estimators/tests/RobotDynamicsEstimator/JointVelocityDynamicsTest.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include <BipedalLocomotion/RobotDynamicsEstimator/JointVelocityDynamics.h>
+
+using namespace BipedalLocomotion::Estimators::RobotDynamicsEstimator;
+using namespace BipedalLocomotion::ParametersHandler;
+using namespace BipedalLocomotion::System;
+
+double random_double()
+{
+    return 1.0 * ((double)rand() - RAND_MAX / 2) / ((double)RAND_MAX);
+}
```
S-Dafarra, line 34: Same as above

src/Estimators/tests/RobotDynamicsEstimator/ZeroVelocityDynamicsTest.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include <BipedalLocomotion/RobotDynamicsEstimator/ZeroVelocityDynamics.h>
+
+using namespace BipedalLocomotion::Estimators::RobotDynamicsEstimator;
+using namespace BipedalLocomotion::ParametersHandler;
+using namespace BipedalLocomotion::System;
+
+double random_double()
+{
+    return 1.0 * ((double)rand() - RAND_MAX / 2) / ((double)RAND_MAX);
+}
```
S-Dafarra, line 27: As above

isorrentino:
Hi @GiulioRomualdi and @S-Dafarra, I modified most of the code because using it I noticed that some things were missing or were uncomfortable to use. I also added some missing tests. I'm going to apply the reviews made by Stefano, but you will probably have to do the review again.

S-Dafarra:
> Hi @GiulioRomualdi and @S-Dafarra, I modified most of the code because using it I noticed that some things were missing or were uncomfortable to use. I also added some missing tests. I'm going to apply the reviews made by Stefano, but you will probably have to do the review again.

Sounds good @isorrentino, when you are ready you can re-request the review
Screenshot from 2023-03-16 12-21-24

isorrentino:
> Sounds good @isorrentino, when you are ready you can re-request the review

I requested the review again but I found a problem and I'm trying to fix it. Let's wait for the review. Thanks.

isorrentino:
Hi @S-Dafarra and @GiulioRomualdi, this PR can be reviewed again. Thank you.

isorrentino:
Friendly ping @S-Dafarra @GiulioRomualdi :)

src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/Dynamics.h
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ * BLF_REGISTER_DYNAMICS is a macro that can be used to register a Dynamics. The key of
+ * the dynamics will be the stringified version of the Dynamics C++ Type
+ * @param _model the model of the dynamics
+ * @param _baseModel the base model from which the _model inherits.
+ */
+#define BLF_REGISTER_DYNAMICS(_model, _baseModel)               \
+    static std::shared_ptr<_baseModel> _model##FactoryBuilder() \
+{                                                               \
```
S-Dafarra, line 31: Indentation problem

src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/Dynamics.h
```diff
@@ -0,0 +1,198 @@
...
+#include <BipedalLocomotion/RobotDynamicsEstimator/SubModel.h>
+#include <BipedalLocomotion/RobotDynamicsEstimator/SubModelKinDynWrapper.h>
+
+/**
+ * BLF_REGISTER_DYNAMICS is a macro that can be used to register a Dynamics. The key of
+ * the dynamics will be the stringified version of the Dynamics C++ Type
+ * @param _model the model of the dynamics
+ * @param _baseModel the base model from which the _model inherits.
+ */
+#define BLF_REGISTER_DYNAMICS(_model, _baseModel)               \
```
S-Dafarra, line 29: The term "dynamics" is quite abused and omnipresent. While it is safe to have a class named ``Dynamics`` in a specific namespace, this macro has to be in the root namespace. I would suggest using a more specific name, maybe adding an ``ESTIMATOR`` keyword

src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/Dynamics.h
```diff
@@ -0,0 +1,198 @@
...
+    Eigen::VectorXd robotJointPositions; /**< Vector of joint positions. */
+    manif::SE3d robotBasePose; /**< Robot base position and orientation. */
+    manif::SE3d::Tangent robotBaseVelocity; /**< Robot base velocity. */
+    manif::SE3d::Tangent robotBaseAcceleration; /**< Robot base acceleration. */
+};
+
+/**
+ * UkfInputProvider describes the provider for the inputs of the ukf
+ */
+class UkfInputProvider : public System::Advanceable<UKFInput, UKFInput>
```
S-Dafarra, line 60: I would suggest adding more details on what this class is needed for. From the documentation only I was not able to understand it.

src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/Dynamics.h
```diff
@@ -0,0 +1,198 @@
...
+    Eigen::VectorXd m_updatedVariable; /**< Updated variable computed using the dynamic model. */
+    std::string m_description{"Generic Dynamics Element"}; /**< String describing the type of the dynamics */
+    std::string m_name; /**< Name of dynamics. */
+    Eigen::VectorXd m_covariances; /**< Vector of covariances. */
+    std::vector<std::string> m_elements = {}; /**< Elements composing the variable vector. */
+    std::string m_dynamicModel;
+    bool m_isInitialized{false}; /**< True if the dynamics has been initialized. */
+    System::VariablesHandler m_stateVariableHandler; /**< Variable handler describing the variables and the sizes in the ukf state/measurement vector. */
+    bool m_isStateVariableHandlerSet{false}; /**< True if setVariableHandler is called. */
+    UKFInput m_ukfInput;
```
S-Dafarra, line 108: Looking at the header only, I find a little suspicious to have all these as protected members. Is it ok for the derived classes to access all these data. I would suggest eventually to keep the data as private as possible, and use protected setters and getters to structure the access to this data.

Then, looking at the implementation, I noticed that almost none of this data is used in the base class. I would suggest leaving the handling of the private attributes directly to the derived class, rather than forcing all the derived classes using some specific variables.

src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/FrictionTorqueStateDynamics.h
```diff
@@ -0,0 +1,135 @@
...
+     * @return true in case of success, false otherwise.
+     */
+    bool finalize(const System::VariablesHandler& stateVariableHandler) override;
+
+    /**
+     * Set the SubModelKinDynWrapper object.
+     * @param kinDynWrapper pointer to a SubModelKinDynWrapper object.
+     * @return True in case of success, false otherwise.
+     */
+    bool setSubModels(const std::vector<SubModel>& subModelList, const std::vector<std::shared_ptr<SubModelKinDynWrapper>>& kinDynWrapperList) override;
```
S-Dafarra, line 102: Same as above, missing doc for the second entry

src/Estimators/src/FrictionTorqueStateDynamics.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+    // Set the state process covariance
+    if (!ptr->getParameter("covariance", m_covariances))
+    {
+        log()->error("{} Error while retrieving the covariance variable.", errorPrefix);
+        return false;
+    }
+
+    // Set the dynamic model type
+    if (!ptr->getParameter("dynamic_model", m_dynamicModel))
```
S-Dafarra, line 45: Is this used anywhere?

S-Dafarra (commented):
Some preliminary comments. I haven't gone through all the files yet

src/Estimators/src/UkfState.cpp
```diff
@@ -0,0 +1,327 @@
...
+    {
+        log()->error("{} The ukf state is not well initialized.", logPrefix);
+        throw std::runtime_error("Error");
+    }
+
+    if (m_pimpl->ukfInputProvider == nullptr)
+    {
+        log()->error("{} The ukf input provider is not set.", logPrefix);
+        throw std::runtime_error("Error");
+    }
```
S-Dafarra, line 273: This differs from the error-catching mechanism we used in the rest of the library. In particular, we mainly used an output boolean to indicate a failure, rather than an exception. I also understand that you needed to override this method, so there was not so much of an option in the first place.

With the risk of sounding like a broken record, I wonder if we could have something like
```cpp
bool advance(){
    try 
    {
        propagate(inputs);
    } 
    catch (std::runtime_error)
    {
        return false;
    }
    return true;
}
```

In this way, you would use this method as all the other methods in the library.

src/Estimators/tests/RobotDynamicsEstimator/UkfStateTest.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+    REQUIRE(inputProvider->setInput(input));
+
+    stateModel->propagate(currentState, updatedState);
+
+    for (int idx = 0; idx < updatedState.size(); idx++)
+    {
+        REQUIRE(std::abs(updatedState[idx]) < 200);
+    }
+}
```
S-Dafarra, line 192: I would also suggest defining a test case where you expect some specific output, and check that you converge there. You could generate in some way some simple synthetic data (i.e. a set of constant joint torques), and check that the estimator converges to the quantities you used to generate the data (i.e. you check that you are able to reconstruct the initial set of joint torques)

S-Dafarra (commented):
I should have checked all the files now :blush:

isorrentino:
Hi @GiulioRomualdi @S-Dafarra, I'm finally back to this PR. I know that @S-Dafarra has already spent time on the review, but I thought it might make more sense to close the 4 PRs containing the code of the estimator (#615, #640, #646, #649) and open many much smaller ones to simplify and speed up the review process. Consider also that the code of the last PR (#649) changes also many parts of the code contained in the previous three PRs. What do you think?

S-Dafarra:
> Hi @GiulioRomualdi @S-Dafarra, I'm finally back to this PR. I know that @S-Dafarra has already spent time on the review, but I thought it might make more sense to close the 4 PRs containing the code of the estimator (#615, #640, #646, #649) and open many much smaller ones to simplify and speed up the review process. Consider also that the code of the last PR (#649) changes also many parts of the code contained in the previous three PRs. What do you think?

Clearly, smaller PRs are much easier to review, but anyway there is always a tradeoff with the development time. Up to you! I did not check the other PRs yet. If you close this, I would just ask you to consider my previous comments in the new PRs before asking me to review them (only if they still apply) :blush:

isorrentino:
> If you close this, I would just ask you to consider my previous comments in the new PRs before asking me to review them (only if they still apply) 

Of course, I would have not ignored your comments :) Actually, since the second PR is very similar to this one, they will apply also to that one :D

src/Estimators/src/UkfState.cpp
isorrentino: I see your point, but this is not applicable here because the method `propagate` is used directly from the 
[bayes-filters-lib](https://github.com/robotology/bayes-filters-lib) library. So, I can't have an additional method (`advance`) as I don't have the possibility to call it.

isorrentino:
As mentioned in https://github.com/ami-iit/bipedal-locomotion-framework/pull/615#issuecomment-1523654641, I'm going to close this PR without merging it.

src/Estimators/src/UkfState.cpp
S-Dafarra: Ah, I see. So you never expect the user to call ``propagate``, right?

isorrentino: > ever expect the user to call `propagate`

Yes, that’s right.

S-Dafarra: Sounds good then!

PR #624, opened 2023-03-13
Restructure ContactDetector machinery 

GiulioRomualdi:
This is the first PR that will bring the [CentroidalMPC](https://github.com/ami-iit/paper_romualdi_2022_icra_centroidal-mpc-walking) merged in master. 
While implementing the [`WholeBodyQPBlock`](https://github.com/ami-iit/paper_romualdi_2022_icra_centroidal-mpc-walking/blob/bd9f6a2537d45ca9c0c6bb6441bebbe4022087f3/src/centroidal-mpc-walking/src/WholeBodyQPBlock.cpp), I noticed that the behavior of `FixedFootDetector` (the class used to get the fixed foot given planned footsteps) was not coherent with the others advanceable in the project. Indeed, before this PR, a user had to call `FixedFootDetector::getFixedFoot()` before calling `FixedFootDetector::advance()`. This behavior was counterintuitive since the `Advanceable` should be able to provide its output only if `advance()` is called. Moreover, the `FixedFootDetector` internal timer was reset when a new sequence of contact was provided. This prevented the usage of `FixedFootDetector` without having a predefined contact sequence. 
Thanks to this refactory the `FixedFootDetector` usage becomes similar to the others
advanceable. Indeed now `advance()` considers the input set by the user and provides the
corresponding output.
⚠️   Even if this modification does not break the API the user may notice some strange behavior if `advance()` was called after getting the output of the detector. The only user affected by this modification should be (@GiulioRomualdi in [paper_romualdi_2022_icra_centroidal-mpc-walking](https://github.com/ami-iit/paper_romualdi_2022_icra_centroidal-mpc-walking))

While implementing [paper_romualdi_2022_icra_centroidal-mpc-walking](https://github.com/ami-iit/paper_romualdi_2022_icra_centroidal-mpc-walking)), I also noticed some undesired behavior due to the number representation of double. Indeed I noticed that the following code
```c++
double time = 0.0
double dt = 0.1;
time += dt;
time += dt;
time += dt;

bool is_lower_equal = time <= 0.3;
```
 was giving me `is_lower_equal == False`. This is due to [round-off error](https://en.wikipedia.org/wiki/Round-off_error) Indeed `time` was slightly lower than 0.3 (for instance `0.3 - 1e-8`). This commit https://github.com/ami-iit/bipedal-locomotion-framework/pull/624/commits/2b6ada8629517a2175dbdf32220be1ce7397590c robustifies the comparison by adding a tolerance (by default equal to 0) that considers the round-off error.

Last but not least in the refractory I also moved the implementation of the `SchmittTrigger` into the math component since it can be used also by other components.

### DIFF
--- a/src/Contacts/src/ContactPhaseList.cpp
+++ b/src/Contacts/src/ContactPhaseList.cpp
@@ -140,12 +140,12 @@ bool ContactPhaseList::setLists(const std::initializer_list<ContactList>& contac
     return true;
 }
 
-ContactPhaseList::const_iterator ContactPhaseList::getPresentPhase(double time) const
+ContactPhaseList::const_iterator ContactPhaseList::getPresentPhase(double time, double tolerance /*= 0*/) const
 {
     // With the reverse iterator we find the last phase such that the begin time is smaller that
     // time
-    auto presentReverse = std::find_if(rbegin(), rend(), [time](const ContactPhase& a) -> bool {
-        return a.beginTime <= time;
+    auto presentReverse = std::find_if(rbegin(), rend(), [time, tolerance](const ContactPhase& a) -> bool {
+        return a.beginTime < time || std::abs(a.beginTime - time) <= tolerance;
     });
 
     if (presentReverse == rend())
--- a/src/Contacts/src/ContactPhaseList.cpp
+++ b/src/Contacts/src/ContactPhaseList.cpp
@@ -140,13 +140,16 @@ bool ContactPhaseList::setLists(const std::initializer_list<ContactList>& contac
     return true;
 }
 
-ContactPhaseList::const_iterator ContactPhaseList::getPresentPhase(double time, double tolerance /*= 0*/) const
+ContactPhaseList::const_iterator
+ContactPhaseList::getPresentPhase(double time,
+                                  double tolerance /*= std::numeric_limits<double>::min();*/) const
 {
     // With the reverse iterator we find the last phase such that the begin time is smaller that
     // time
-    auto presentReverse = std::find_if(rbegin(), rend(), [time, tolerance](const ContactPhase& a) -> bool {
-        return a.beginTime < time || std::abs(a.beginTime - time) <= tolerance;
-    });
+    auto presentReverse
+        = std::find_if(rbegin(), rend(), [time, tolerance](const ContactPhase& a) -> bool {
+              return a.beginTime <= time + tolerance;
+          });
 
     if (presentReverse == rend())
     {
--- a/src/Contacts/src/ContactPhaseList.cpp
+++ b/src/Contacts/src/ContactPhaseList.cpp
@@ -140,9 +140,9 @@ bool ContactPhaseList::setLists(const std::initializer_list<ContactList>& contac
     return true;
 }
 
-ContactPhaseList::const_iterator
-ContactPhaseList::getPresentPhase(double time,
-                                  double tolerance /*= std::numeric_limits<double>::min();*/) const
+ContactPhaseList::const_iterator ContactPhaseList::getPresentPhase(
+    double time,
+    double tolerance /*= BipedalLocomotion::Math::AbsoluteEqualityDoubleTolerance*/) const
 {
     // With the reverse iterator we find the last phase such that the begin time is smaller that
     // time
--- a/src/Contacts/src/FixedFootDetector.cpp
+++ b/src/Contacts/src/FixedFootDetector.cpp
@@ -56,8 +56,7 @@ bool FixedFootDetector::updateFixedFoot()
     constexpr auto logPrefix = "[FixedFootDetector::updateFixedFoot]";
 
     // search the phase associated to the current time
-    constexpr auto tolerance = std::numeric_limits<double>::min();
-    auto phase = m_contactPhaselist.getPresentPhase(m_currentTime, tolerance);
+    auto phase = m_contactPhaselist.getPresentPhase(m_currentTime);
 
     if(phase == m_contactPhaselist.end())
     {
--- a/src/Contacts/src/FixedFootDetector.cpp
+++ b/src/Contacts/src/FixedFootDetector.cpp
@@ -10,14 +10,21 @@
 #include <BipedalLocomotion/ParametersHandler/IParametersHandler.h>
 #include <BipedalLocomotion/TextLogging/Logger.h>
 
+#include <limits>
 #include <memory>
 
 using namespace BipedalLocomotion::Contacts;
 using namespace BipedalLocomotion::ParametersHandler;
 
-bool FixedFootDetector::customInitialization(std::weak_ptr<const IParametersHandler> handler)
+bool FixedFootDetector::initialize(std::weak_ptr<const IParametersHandler> handler)
 {
     constexpr auto logPrefix = "[FixedFootDetector::customInitialization]";
+
+    if (m_detectorState != State::NotInitialized)
+    {
+        log()->error("{} The contact detector already seems to be initialized.", logPrefix);
+        return false;
+    }
 
     auto ptr = handler.lock();
     if (ptr == nullptr)
@@ -28,45 +35,54 @@ bool FixedFootDetector::customInitialization(std::weak_ptr<const IParametersHand
 
     if (!ptr->getParameter("sampling_time", m_dT))
     {
-        log()->error("{} Unable to find the sampling time.", logPrefix);
+        log()->error("{} Unable to find the 'sampling_time' parameter.", logPrefix);
         return false;
     }
 
     if (m_dT <= 0)
     {
-        log()->error("{} The sampling time must be a strictly positive number.", logPrefix);
+        log()->error("{} The parameter 'sampling_time' must be a strictly positive number.",
+                     logPrefix);
         return false;
     }
+
+    m_detectorState = State::Initialized;
 
     return true;
 }
 
 bool FixedFootDetector::updateFixedFoot()
 {
-    assert(m_currentTime >= m_contactPhaselist.firstPhase()->beginTime
-           && "[FixedFootDetector::updateFixedFoot] The current time must be greater than "
-              "equal the first contact phase. If you read this assert there is a bug in the code");
+    constexpr auto logPrefix = "[FixedFootDetector::updateFixedFoot]";
 
     // search the phase associated to the current time
-    auto phase = std::find_if(m_contactPhaselist.cbegin(),
-                              m_contactPhaselist.cend(),
-                              [&](const auto& phase) -> bool {
-                                  return phase.beginTime <= m_currentTime
-                                      && m_currentTime <= phase.endTime;
-                              });
+    constexpr auto tolerance = std::numeric_limits<double>::min();
+    auto phase = m_contactPhaselist.getPresentPhase(m_currentTime, tolerance);
 
-    // take the last phase if not found
-    if (phase == m_contactPhaselist.cend())
-        phase = m_contactPhaselist.lastPhase();
+    if(phase == m_contactPhaselist.end())
+    {
+        log()->error("{} No phase has the begin time lower than the specified time.", logPrefix);
+        return false;
+    }
+
+    if (phase->activeContacts.size() > 2)
+    {
+        log()->error("{} The base detector can be used only for bipedal locomotion and it cannot "
+                     "handle more than two actives contacts.",
+                     logPrefix);
+        return false;
+    }
 
     // if the robot is in single support only one contact is enabled
     if (phase->activeContacts.size() == 1)
     {
         // get the enabled contact
         const auto it = phase->activeContacts.cbegin();
-        assert(m_contactStates.find(it->first) != m_contactStates.end()
-               && "[FixedFootDetector::updateFixedFoot] Unable to find the contact. This "
-                  "should not be possible. If you read this assert there is a bug in the code");
+        if (m_contactStates.find(it->first) == m_contactStates.end())
+        {
+            log()->error("{} Unable to find the contact.", logPrefix);
+            return false;
+        }
 
         // update the contacts
         for (auto& [name, contact] : m_contactStates)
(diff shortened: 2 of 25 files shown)


GiulioRomualdi:
Windows is failing with the following error:
```
D:\a\bipedal-locomotion-framework\bipedal-locomotion-framework\src\Contacts\src\SchmittTriggerDetector.cpp(99): error C7555: use of designated initializers requires at least '/std:c++20'
D:\a\bipedal-locomotion-framework\bipedal-locomotion-framework\src\Contacts\src\SchmittTriggerDetector.cpp(258): error C7555: use of designated initializers requires at least '/std:c++20'
D:\a\bipedal-locomotion-framework\bipedal-locomotion-framework\src\Contacts\src\SchmittTriggerDetector.cpp(274): error C7555: use of designated initializers requires at least '/std:c++20'
```
I wasn't aware of this. I need to restructure the code

S-Dafarra:
> While implementing [paper_romualdi_2022_icra_centroidal-mpc-walking](https://github.com/ami-iit/paper_romualdi_2022_icra_centroidal-mpc-walking)), I also noticed some undesired behavior due to the number representation of double. Indeed I noticed that the following code
> 
> ```c++
> double time = 0.0
> double dt = 0.1;
> time += dt;
> time += dt;
> time += dt;
> 
> bool is_lower_equal time <= 0.3;
> ```
> 
> was giving me `is_lower_equal == False`. This is due to [round-off error](https://en.wikipedia.org/wiki/Round-off_error) Indeed `time` was slightly lower than 0.3 (for instance `0.3 - 1e-8`). This commit [2b6ada8](https://github.com/ami-iit/bipedal-locomotion-framework/commit/2b6ada8629517a2175dbdf32220be1ce7397590c) robustifies the comparison by adding a tolerance (by default equal to 0) that considers the round-off error.

I thought it would be interesting to add that 0.1 is a periodic number in binary notation. Hence, they are always "approximated". That's why their sum is different from the product. A good explanation can be found here: https://youtu.be/PZRI1IfStY0?t=192

src/Contacts/src/ContactPhaseList.cpp
```diff
@@ -140,12 +140,12 @@ bool ContactPhaseList::setLists(const std::initializer_list<ContactList>& contac
...
 
-ContactPhaseList::const_iterator ContactPhaseList::getPresentPhase(double time) const
+ContactPhaseList::const_iterator ContactPhaseList::getPresentPhase(double time, double tolerance /*= 0*/) const
 {
     // With the reverse iterator we find the last phase such that the begin time is smaller that
     // time
-    auto presentReverse = std::find_if(rbegin(), rend(), [time](const ContactPhase& a) -> bool {
-        return a.beginTime <= time;
+    auto presentReverse = std::find_if(rbegin(), rend(), [time, tolerance](const ContactPhase& a) -> bool {
+        return a.beginTime < time || std::abs(a.beginTime - time) <= tolerance;
```
S-Dafarra, line 148: This should be equivalent to 
```
a.beginTime <= time + tolerance
```

In this way, you might also use negative tolerances to have the opposite effects, i.e. ``a.beginTime`` << ``time``.

src/Contacts/src/FixedFootDetector.cpp
```diff
@@ -28,45 +35,54 @@ bool FixedFootDetector::customInitialization(std::weak_ptr<const IParametersHand
...
 
     // search the phase associated to the current time
-    auto phase = std::find_if(m_contactPhaselist.cbegin(),
-                              m_contactPhaselist.cend(),
-                              [&](const auto& phase) -> bool {
-                                  return phase.beginTime <= m_currentTime
-                                      && m_currentTime <= phase.endTime;
-                              });
+    constexpr auto tolerance = std::numeric_limits<double>::min();
+    auto phase = m_contactPhaselist.getPresentPhase(m_currentTime, tolerance);
```
S-Dafarra, line 60: We might also consider ``std::numeric_limits<double>::min();`` as default tolerance

S-Dafarra (commented):
Some preliminary comments. I still need to check a bunch of files

src/Contacts/src/SchmittTriggerDetector.cpp
```diff
@@ -1,245 +1,261 @@
...
-        return false;
-    }
+    ok = ok && setupParam("contact_make_switch_times", switchOnAfter);
 
     std::vector<double> switchOffAfter;
-    if (! m_pimpl->setupParamV(handler, "contact_break_switch_times", switchOffAfter, printPrefix))
-    {
-        return false;
-    }
+    ok = ok && setupParam("contact_break_switch_times", switchOffAfter);
```
S-Dafarra, line 80: ``ok`` is no more used. I guess you are missing a 
```cpp
if (!ok)
    return false;
````
here

File: src/Math/include/BipedalLocomotion/Math/SchmittTrigger.h [hunk 4]
```diff
@@ -0,0 +1,137 @@
...
+                             from high to low in seconds.*/
+};
+
+/**
+ * SchmittTriggerState contains the input of the SchmittTrigger class.
+ */
+struct SchmittTriggerInput
+{
+    double time{0.0}; /**< Current time instant in seconds */
+    double rawValue{0.0} ; /**< Raw value that should  */
```
S-Dafarra, line 38: I think a piece is missing here.

src/Math/include/BipedalLocomotion/Math/SchmittTrigger.h
```diff
@@ -0,0 +1,137 @@
...
+                                     switchOffAfter time-units*/
+        double switchOnAfter{0.0}; /**< time units to wait for before switching to ON state from OFF
+                                      state. Ensure it's greater than sampling time. */
+        double switchOffAfter{0.0}; /**< time units to wait for before switching to OFF state from
+                                       ON state. Ensure it's greater than sampling time. */
+
+        /** Threshold used for the comparison of two time instant. Given two time instants if the
+         * error between the two is lower than the threshold, the time instants are considered
+         * equal. */
+        double timeComparisonThreshold{std::numeric_limits<double>::epsilon()};
```
S-Dafarra, line 66: Why ``epsilon`` and not ``min``?

src/Math/include/BipedalLocomotion/Math/SchmittTrigger.h
```diff
@@ -0,0 +1,137 @@
...
+    };
+
+    /**
+     * Initialize the SchmittTrigger block.
+     * @param handler pointer to the parameter handler.
+     * @note The following parameters are required
+     * |  Parameter Name  |   Type   |                                     Description                                     | Mandatory |
+     * |:----------------:|:--------:|:-----------------------------------------------------------------------------------:|:---------:|
+     * |  `on_threshold`  | `double` | High value threshold to initiate an ON state switch after switchOnAfter time-units  |    Yes    |
+     * | `off_threshold`  | `double` | Low value threshold to initiate an OFF state switch after switchOffAfter time-units |    Yes    |
```
S-Dafarra, line 76: Also here I think it is not clear what is meant by time-units

src/Contacts/src/ContactPhaseList.cpp
GiulioRomualdi: Yes you are right, I applied what you suggested in  9c1a49364634ca8d52439d19e444ce13c30efb85

src/Contacts/src/FixedFootDetector.cpp
GiulioRomualdi: Done in https://github.com/ami-iit/bipedal-locomotion-framework/commit/9c1a49364634ca8d52439d19e444ce13c30efb85

src/Math/src/SchmittTrigger.cpp
```diff
@@ -0,0 +1,163 @@
...
+{
+
+    auto robustEqual = [this](double a, double b) -> bool {
+        return (std::abs(a - b) <= this->m_params.timeComparisonThreshold);
+    };
+
+    // if the trigger is not active
+    if (!m_state.state)
+    {
+        // the state is deactivate this means that we can reset the fallingEdgeTimeInstant. We can
```
S-Dafarra, line 83:
```suggestion
        // the state is inactive this means that we can reset the fallingEdgeTimeInstant. We can
```

src/Math/src/SchmittTrigger.cpp
```diff
@@ -0,0 +1,163 @@
...
+                m_state.state = true;
+                m_state.switchTime = m_input.time;
+                m_state.edgeTime = m_risingEdgeTimeInstant;
+                m_timer = 0;
+            }
+        }
+    } else
+    {
+
+        // the state is deactivate this means that we can reset the risingEdgeTimeInstant. We can
```
S-Dafarra, line 111:
```suggestion
        // the state is inactive this means that we can reset the risingEdgeTimeInstant. We can
```

src/Math/include/BipedalLocomotion/Math/SchmittTrigger.h
GiulioRomualdi: You are right. At this point, I think that `epsilon` is better than min indeed as explained [here](https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon)

- **`std::numeric_limits<double>::epsilon()`** Returns the machine epsilon, that is, the difference between 1.0 and the next value representable by the floating-point type T.

- **`std::numeric_limits<double>::min()`** Returns the minimum finite value representable by the numeric type T.




In https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon, shows how to use `std::numeric_limits<double>::epsilon()` to compare floating-point values for equality.

GiulioRomualdi: We may implement something like this in `Math` component and use it everywhere is needed 
```c++
template<class T>
typename [std::enable_if](http://en.cppreference.com/w/cpp/types/enable_if)<![std::numeric_limits](http://en.cppreference.com/w/cpp/types/numeric_limits)<T>::is_integer, bool>::type
    almost_equal(T x, T y, int ulp)
{
    // the machine epsilon has to be scaled to the magnitude of the values used
    // and multiplied by the desired precision in ULPs (units in the last place)
    return [std::fabs](http://en.cppreference.com/w/cpp/numeric/math/fabs)(x - y) <= [std::numeric_limits](http://en.cppreference.com/w/cpp/types/numeric_limits)<T>::epsilon() * [std::fabs](http://en.cppreference.com/w/cpp/numeric/math/fabs)(x + y) * ulp
        // unless the result is subnormal
        || [std::fabs](http://en.cppreference.com/w/cpp/numeric/math/fabs)(x - y) < [std::numeric_limits](http://en.cppreference.com/w/cpp/types/numeric_limits)<T>::min();
}
```

S-Dafarra: That's where my doubt arose. ``epsilon`` seems to work better for checking if you are close to 1, while in https://github.com/ami-iit/bipedal-locomotion-framework/blob/f3241a5471b44e5f2815a32c06c9c82aeca51829/src/Math/src/SchmittTrigger.cpp#L77 you are checking a difference that should be close to zero. Note that the example of https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon have another part on the right hand side. In order to copy it you should do
```cpp
(std::abs(a - b) <= this->m_params.timeComparisonThreshold) * std::abs(a + b) )
```
but I don't think it is worth adding the additional multiplication. In addition, it makes the check more and more coarse the higher is ``a+b``. That's why they also added the 
```cpp
|| std::fabs(x - y) < std::numeric_limits<T>::min();
```
in the check (again using ``min``).

In any case, we are talking about an infinitesimal difference, so feel free to use the version you prefer :blush:

S-Dafarra: > We may implement something like this in `Math` component and use it everywhere is needed (Taken from https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon)
> 
> ```c++
> template<class T>
> typename std::enable_if<!std::numeric_limits<T>::is_integer, bool>::type
>     almost_equal(T x, T y, int ulp)
> {
>     // the machine epsilon has to be scaled to the magnitude of the values used
>     // and multiplied by the desired precision in ULPs (units in the last place)
>     return std::fabs(x - y) <= std::numeric_limits<T>::epsilon() * std::fabs(x + y) * ulp
>         // unless the result is subnormal
>         || std::fabs(x - y) < std::numeric_limits<T>::min();
> }
> ```

Loving it, but I guess it is a bit of an over-complexification :thinking:

GiulioRomualdi: Here I am! So I tried to do some tests following this [nice post](https://embeddeduse.com/2019/08/26/qt-compare-two-floats/) with a toy problem implemented in [floating-point-comparison](https://github.com/GiulioRomualdi/floating-point-comparison). Here I implemented for type of comparisons
```c++

template <typename T>
bool comparison0(T f1, T f2, T tolerance)
{
    return f1 == f2;
}

template <typename T>
bool comparison1(T f1, T f2, T tolerance)
{
    return std::abs(f1 - f2) <= tolerance;
}

template <typename T>
bool comparison2(T f1, T f2, T tolerance)
{
    return std::abs(f1 - f2) <= tolerance * std::max(std::abs(f1), std::abs(f2));
}

template <typename T>
bool comparison3(T f1, T f2, T tolerance)
{
    std::size_t ulp = 2;

    return std::fabs(f1 - f2) <= tolerance * std::fabs(f1 + f2) * ulp
        || std::fabs(f1 - f2) < std::numeric_limits<T>::min();
}
```
The equalities are tested for  from `9.0` to `99` in steps of `0.01` 

Setting the tolerance equal to **`1e-10`** We got the following results where **correct** means the number of comparisons that were correct 
| Comparison Function | correct | total |
|:---:|:--:|:---:|
| 0 | 10 | 10000 |
| 1 | 10000 | 10000 |
| 2 | 9999 | 10000 |
| 3 | 9999 | 10000 |

On the other hand setting the tolerance equal to **`std::numeric_limit<double>::epsilon()`**  I got the following

| Comparison Function | correct | total |
|:---:|:--:|:---:|
| 0 | 24 | 10000 |
| 1 | 42 | 10000 |
| 2 | 70 | 10000 |
| 3 | 274 | 10000 |

The following conclusion holds (not in general but in our case):
- **`std::numeric_limit<double>::epsilon()`** is too low tolerance
- even if [here](https://embeddeduse.com/2019/08/26/qt-compare-two-floats/), and [here](https://stackoverflow.com/questions/67842444/modern-practice-to-compare-double-float-for-equality-in-modern-c) suggested to avoid considering an absolute threshold for us `comparison1()` seems to be enough.

So we should decide the tolerance we want to guarantee in blf. (We can store it as parameter in Math component along with the Gravity)

GiulioRomualdi:
Hi @S-Dafarra, I applied most of the changes you suggested except https://github.com/ami-iit/bipedal-locomotion-framework/pull/624#discussion_r1137048108 since https://github.com/ami-iit/bipedal-locomotion-framework/pull/630 will revisit how the time is handled inside the entire project.

The plan is to merge this PR and then move to #630

S-Dafarra (approved):

PR #625, opened 2023-03-14
Handle the case without force/torque sensors 

isorrentino:
The case without force/torque sensors to split the model was not considered and if the list of sensors was empty we had a segmentation fault error. This PR fixes the problem and now if the user does not specify a list of sensors, the SubModelCreator creates a unique SubModel object corresponding to the full model.

### DIFF
--- a/src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/SubModel.h
+++ b/src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/SubModel.h
@@ -74,45 +74,100 @@ class SubModel
 
 public:
     /**
+     * Determines the validity of the object.
+     * @return True if the object is valid, false otherwise.
+     */
+    bool isValid() const;
+
+    /**
      * Getters
      */
 
     /**
-     * @brief Access model.
+     * @brief Get the `iDynTree::Model` instance.
+     * @note The actual implementation of the Model is currently stored in an `iDynTree::Model`
      * @return The model of the SubModel.
      */
     const iDynTree::Model& getModel() const;
 
     /**
-     * @brief Access jointListMapping.
+     * @brief Access the `std::vectot<int>` list.
      * @return the mapping between the joint indeces in the sub-model and the joint indeces in the
      * full-model.
      */
     const std::vector<int>& getJointMapping() const;
 
     /**
-     * @brief Access ftList.
+     * @brief Access the `std::vector<FT>` list.
      * @return the list of FT objects which is the list of force/torque sensors.
      */
     const std::vector<FT>& getFTList() const;
 
     /**
-     * @brief Access accelerometerList.
+     * @brief Access the `std::vector<Sensor>` list of acceletometer sensors.
      * @return a list of Sensor objects describing the accelerometers contained in the sub-model.
      */
     const std::vector<Sensor>& getAccelerometerList() const;
 
     /**
-     * @brief Access gyroscopeList.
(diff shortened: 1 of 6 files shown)


src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/SubModel.h
```diff
@@ -73,47 +73,101 @@ class SubModel
...
+     */
+    bool isValid() const;
+
     /**
      * Getters
      */
 
     /**
-     * @brief Access model.
+     * @brief Access m_model.
```
GiulioRomualdi, line 87: I would avoid mentioning `m_model` you can change the documentation in 
```suggestion
     * @brief Get the `iDynTree::Model` instance.
     * @note The actual implementation of the Model is currently stored in an `iDynTree::Model`
```

This is valid also for the others

src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/SubModel.h
```diff
@@ -73,47 +73,101 @@ class SubModel
...
+     * @brief Access m_externalContactList.
      * @return a list of strings describing frame names of the external contacts for the sub-model.
      */
     const std::vector<std::string>& getExternalContactList() const;
 
+    /**
+     * @brief access the length of m_ftList.
+     * @return the number of force/torque sensors in the sub-model.
+     */
+    int getNrOfFTSensor() const;
```
GiulioRomualdi, line 127:
```suggestion
    std::size_t getNrOfFTSensor() const;
```

This is valid also foe the others

src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/SubModel.h
isorrentino: Done

src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/SubModel.h
isorrentino: Done

src/Estimators/src/SubModel.cpp
```diff
@@ -371,27 +390,29 @@ bool blfEstim::SubModelCreator::createSubModels(
...
 }
 
-void blfEstim::SubModelCreator::setModel(const iDynTree::Model& model)
+void RDE::SubModelCreator::setModelAndSensors(iDynTree::ModelLoader& modelLoader)
 {
-    this->m_model = model;
+    m_model = modelLoader.model();
+
+    m_sensorList = modelLoader.sensors();
 }
```
S-Dafarra, line 413: From outside, it can be a bit misleading the fact that the method is ``setModelAndSensors``, but you are passing a single input. Moreover, in this way you limit yourself to the use of ``iDynTree::ModelLoader``, but that is not strictly necessary. I would suggest to change the signature to 
```cpp
void setModelAndSensors(const iDynTree::Model& mode, iDynTree::SensorsList& sensors);
```
and from outside you will simply call it with
```cpp
subModel.setModelAndSensors(mdldr.model(), mdldr.sensors())
```

isorrentino: Done!

S-Dafarra (approved):

GiulioRomualdi:
The first commit shouldn't be there 😭 

image

GiulioRomualdi:
Let's wait for the CI. (I should add the auto-merge CI 🤔 )

PR #626, opened 2023-03-14
Implement the DiscreteGeometryContact in Contacts component

GiulioRomualdi:
I don't really like `DiscreteGeometryContact`. however, I was not able to find a better name.  @S-Dafarra perhaps have some suggestions


S-Dafarra (approved):
The name looks reasonable to me. I checked quickly on https://en.wikipedia.org/wiki/Discrete_geometry. In general a discrete geometric object is constituted by a finite set of basic geometric objects. Maybe a more appropriate name could be ``PolytopeContact``, but in the end it depends if the corners can be considered also spheres (or other basic geometric shapes). In this case then, your initial proposition would be more accurate.

PR #630, opened 2023-03-17
Refactor how the time is handled in the entire framework

GiulioRomualdi:
Thanks to this PR we will change the page on how the time is considered in the framework.

Since now the time was considered as double, Now we move to `std::chrono::nanoseconds`. This will solve all the problems similar to what discussed in https://github.com/ami-iit/bipedal-locomotion-framework/pull/624#discussion_r1138761866

In detail this PR:
1. Add the support of `std::chrono` in The text logging
2. Add the possibility to retrieve and set duration from the `IParametersHandler`
   > With this feature it will be possible to specify the DateTime in an `ini` and a `toml` file. [`toml` officially support the local-time.](https://github.com/toml-lang/toml/blob/main/toml.md#local-time) Indeed, if you include only the time portion of an [RFC 3339](https://tools.ietf.org/html/rfc3339) formatted date-time, it will represent that time of day
   >  ```toml
   >  dt = 00:00:00.010000
   >  ```
   > means that `dt` is equal to `0.01 seconds` Following the same language specification ([RFC 3339](https://tools.ietf.org/html/rfc3339)) I also added the local time support to the yarp `ini` file specification (@traversaro, do you think this may be ported directly inside yarp?).
3. Restructure the Contacts component to handle time with `std::chrono::nanoseconds`
4. Restructure the Planners component to handle time with `std::chrono::nanoseconds`
5. Restructure the FloatingBaseEstimator component to handle time with `std::chrono::nanoseconds`
6. Update the `blf-position-tracking` to handle time with `std::chrono::nanoseconds`
7. Remove all the thresholds added in #624 in the `ContactDetectors` required by https://github.com/ami-iit/bipedal-locomotion-framework/pull/624#discussion_r1138761866
8. Restructure the python bindings

⚠️ Once this PR got merged I can help @paolo-viceconte, @LoreMoretti, @CarlottaSartore  to update the python code that uses the planner component

### DIFF
--- a/src/Contacts/include/BipedalLocomotion/Contacts/ContactList.h
+++ b/src/Contacts/include/BipedalLocomotion/Contacts/ContactList.h
@@ -11,10 +11,10 @@
 // BipedalLocomotion
 #include <BipedalLocomotion/Contacts/Contact.h>
 
-//iDynTree
 #include <manif/manif.h>
 
 // std
+#include <chrono>
 #include <functional>
 #include <set>
 #include <string>
@@ -27,8 +27,9 @@ namespace Contacts
 
 /**
  * @brief Class containing a list of contacts.
- * The contact are added such that the activation time is strictly growing. In addition, contacts cannot be overlapping.
- * It represents a series of contact activations and deactivations of a single entity.
+ * The contact are added such that the activation time is strictly growing. In addition, contacts
+ * cannot be overlapping. It represents a series of contact activations and deactivations of a
+ * single entity.
  */
 class ContactList
 {
@@ -45,7 +46,8 @@ class ContactList
                                                      structure for inserting and ordering
                                                      contacts. **/
     std::string m_defaultName{"ContactList"}; /** Default name for the contact list. **/
-    BipedalLocomotion::Contacts::ContactType m_defaultContactType{BipedalLocomotion::Contacts::ContactType::FULL}; /** Default contact type. **/
+    BipedalLocomotion::Contacts::ContactType m_defaultContactType{
+        BipedalLocomotion::Contacts::ContactType::FULL}; /** Default contact type. **/
     int m_defaultIndex{-1}; /**< Default Frame index of the contact */
 
 public:
@@ -94,7 +96,8 @@ public:
      * @brief Add a new contact to the list.
      * @param newContact The new contact
      * @return false if it was not possible to insert the contact.
-     * Possible failures: the activation time is greater than the deactivation time, or the new contact ovelaps with an existing contact.
+     * Possible failures: the activation time is greater than the deactivation time, or the new
+     * contact overlaps with an existing contact.
      */
     bool addContact(const BipedalLocomotion::Contacts::PlannedContact& newContact);
 
@@ -106,9 +109,12 @@ public:
      * @param activationTime The activation time.
      * @param deactivationTime The deactivation time.
      * @return false if it was not possible to insert the contact.
-     * Possible failures: the activation time is greater than the deactivation time, or the new contact ovelaps with an existing contact.
+     * Possible failures: the activation time is greater than the deactivation time, or the new
+     * contact overlaps with an existing contact.
      */
-    bool addContact(const manif::SE3d& newTransform, double activationTime, double deactivationTime);
+    bool addContact(const manif::SE3d& newTransform,
+                    const std::chrono::nanoseconds& activationTime,
+                    const std::chrono::nanoseconds& deactivationTime);
 
     /**
      * @brief Erase a contact
@@ -128,12 +134,14 @@ public:
     const_iterator cbegin() const;
 
     /**
-     * @brief Return a const reverse iterator to the contacts (basically starting from the last contact going backward).
+     * @brief Return a const reverse iterator to the contacts (basically starting from the last
+     * contact going backward).
      */
     const_reverse_iterator rbegin() const;
 
     /**
-     * @brief Return a const reverse iterator to the contacts (basically starting from the last contact going backward).
+     * @brief Return a const reverse iterator to the contacts (basically starting from the last
+     * contact going backward).
      */
     const_reverse_iterator crbegin() const;
 
@@ -193,7 +201,8 @@ public:
      * @brief Edit an existing contact.
      * @param element Iterator to the element to edit.
      * @param newContact The new contact
-     * @return false if the element is not valid or if the new contact timing would require a reordering of the list.
+     * @return false if the element is not valid or if the new contact timing would require a
+     * reordering of the list.
      */
     bool editContact(const_iterator element,
                      const BipedalLocomotion::Contacts::PlannedContact& newContact);
@@ -208,14 +217,14 @@ public:
      * @return an iterator to the last contact  having an activation time lower than time.
      * If no contact satisfy this condition, it returns a pointer to the end.
      */
-    const_iterator getPresentContact(double time) const;
+    const_iterator getPresentContact(const std::chrono::nanoseconds& time) const;
 
     /**
      * @brief Clear all the steps, except the one returned by getPresentContact
      * @param time The present time.
      * @return false if no contact is available at this time.
      */
-    bool keepOnlyPresentContact(double time);
+    bool keepOnlyPresentContact(const std::chrono::nanoseconds& time);
 
     /**
      * @brief Clear the contacts.
@@ -226,7 +235,6 @@ public:
      * @brief Remove only the last contact.
      */
     void removeLastContact();
-
 };
 
 /**
--- a/src/Contacts/src/ContactList.cpp
+++ b/src/Contacts/src/ContactList.cpp
@@ -67,7 +67,7 @@ bool ContactList::addContact(const PlannedContact &newContact)
         {
             log()->error("{} The new contact (activationTime: {} deactivationTime: {}) is not "
                          "compatible with an element already present in the list (activationTime: "
-                         "{} deactivationTime: {}).",
+                         "{}, deactivationTime: {}.).",
                          errorPrefix,
                          newContact.activationTime,
                          newContact.deactivationTime,
@@ -79,7 +79,9 @@ bool ContactList::addContact(const PlannedContact &newContact)
     return true;
 }
 
-bool ContactList::addContact(const manif::SE3d &newTransform, double activationTime, double deactivationTime)
+bool ContactList::addContact(const manif::SE3d& newTransform,
+                             const std::chrono::nanoseconds& activationTime,
+                             const std::chrono::nanoseconds& deactivationTime)
 {
     PlannedContact newContact;
     newContact.pose = newTransform;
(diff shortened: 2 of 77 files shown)


GiulioRomualdi:
Windows is failing with:
```
SchmittTriggerDetectorUnitTest.obj : error LNK2001: unresolved external symbol "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const Catch::Detail::unprintableString" (?unprintableString@Detail@Catch@@3V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@B) [D:\a\bipedal-locomotion-framework\bipedal-locomotion-framework\build\src\Contacts\tests\ContactDetectors\SchmittTriggerDetectorUnitTests.vcxproj]
```

I was wondering if this is due to the specific version of catch indeed `Catch::Detail::unprintableString` is defined [here](https://github.com/catchorg/Catch2/blob/6783411349db5a87718c9b7ba8b02e90e39d783e/src/catch2/catch_tostring.hpp#L51)

S-Dafarra:
> > Windows is failing with:
> > ```
> > 2023-03-20T12:21:21.4106528Z ParametersHandlerYarpTest.cpp.obj : error LNK2019: unresolved external symbol "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const Catch::Detail::unprintableString" (?unprintableString@Detail@Catch@@3V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@B) referenced in function "public: static class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl Catch::StringMaker<class std::chrono::duration<__int64,struct std::ratio<1,1000000000> >,void>::convert<class std::chrono::duration<__int64,struct std::ratio<1,1000000000> > >(class std::chrono::duration<__int64,struct std::ratio<1,1000000000> > const &)" (??$convert@V?$duration@_JU?$ratio@$00$0DLJKMKAA@@std@@@chrono@std@@@?$StringMaker@V?$duration@_JU?$ratio@$00$0DLJKMKAA@@std@@@chrono@std@@X@Catch@@SA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@AEBV?$duration@_JU?$ratio@$00$0DLJKMKAA@@std@@@chrono@3@@Z)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I was wondering if this is due to the specific version of catch indeed `Catch::Detail::unprintableString` is defined [here](https://github.com/catchorg/Catch2/blob/6783411349db5a87718c9b7ba8b02e90e39d783e/src/catch2/catch_tostring.hpp#L51)
> 
> [catchorg/Catch2#1039](https://github.com/catchorg/Catch2/issues/1039) may contain the solution to our problems

Maybe you can explicitly cast the comparison to ``bool`` to fix it.

src/Contacts/include/BipedalLocomotion/Contacts/ContactList.h
```diff
@@ -94,7 +96,8 @@ class ContactList
      * @brief Add a new contact to the list.
      * @param newContact The new contact
      * @return false if it was not possible to insert the contact.
-     * Possible failures: the activation time is greater than the deactivation time, or the new contact ovelaps with an existing contact.
+     * Possible failures: the activation time is greater than the deactivation time, or the new
+     * contact ovelaps with an existing contact.
```
S-Dafarra, line 100:
```suggestion
     * contact overlaps with an existing contact.
```

src/Contacts/src/ContactList.cpp
```diff
@@ -208,21 +210,28 @@ bool ContactList::editContact(ContactList::const_iterator element, const Planned
...
 }
 
-ContactList::const_iterator ContactList::getPresentContact(double time) const
+ContactList::const_iterator
+ContactList::getPresentContact(const std::chrono::nanoseconds& time) const
 {
-    // With the reverse iterator we find the last step such that the activation time is smaller that time
-    ContactList::const_reverse_iterator presentReverse = std::find_if(rbegin(), rend(),
-                                                        [time](const PlannedContact & a) -> bool { return a.activationTime <= time; });
+    // With the reverse iterator we find the last step such that the activation time is smaller that
```
S-Dafarra, line 216:
```suggestion
    // With the reverse iterator we find the last step such that the activation time is smaller than
```

src/ParametersHandler/YarpImplementation/src/YarpImplementation.cpp
```diff
@@ -15,6 +17,73 @@
...
 using namespace BipedalLocomotion::ParametersHandler;
 
+bool YarpImplementation::stringToDuration(const std::string& string, std::chrono::nanoseconds& time)
+{
+    using namespace std::chrono_literals;
+
+    constexpr auto logPrefix = "[YarpImplementation::stringToDuration]";
+    // \\. means that \ is used as escape character for the regexp and again as escape character for
+    // the string literal.
+    const std::regex expression("[0-9]{2}:[0-9]{2}:[0-9]{2}\\.[0-9]{6}\\b");
```
S-Dafarra, line 27: I would suggest adding an explanation of what this expression is expected to match. This might help the future you reading again the code.

src/ParametersHandler/YarpImplementation/src/YarpImplementation.cpp
```diff
@@ -15,6 +17,73 @@
...
+    if (!std::regex_match(string, expression))
+    {
+        log()->debug("{} The parameter do not match the excepted expression. Expected: "
+                     "'HH:MM:SS.xxxxxx'. Get: {}.",
```
S-Dafarra, line 31: Most of the time we work with seconds. I would suggest accepting also the ``SS.xxxxxxxx`` case.

src/ParametersHandler/YarpImplementation/src/YarpImplementation.cpp
S-Dafarra: Also, I guess we do not allow setting a ``nanosecond`` from configuration file, right?

src/Planners/src/CubicSpline.cpp
```diff
@@ -728,14 +748,15 @@ bool CubicSpline::evaluatePoint(const double& t,
...
-                                 - m_pimpl->polynomials.back().initialPoint->timeInstant;
+        const auto finalTime = m_pimpl->polynomials.back().finalPoint->timeInstant
+                               - m_pimpl->polynomials.back().initialPoint->timeInstant;
         m_pimpl->getPositionAtTime(finalTime, m_pimpl->polynomials.back(), position);
         m_pimpl->getVelocityAtTime(finalTime, m_pimpl->polynomials.back(), velocity);
         m_pimpl->getAccelerationAtTime(finalTime, m_pimpl->polynomials.back(), acceleration);
         return true;
     }
 
+    // TODO(Giulio) Change this algorithm with the reverse iterator
```
S-Dafarra, line 759: Note that in the end, its complexity is always linear in the dimension of the vector.

S-Dafarra (commented):
I would add that the use of ``std::chrono_literals`` is a pretty nice addition! It makes the time quantities very readable.

src/Estimators/src/LeggedOdometry.cpp
```diff
@@ -385,7 +387,8 @@ iDynTree::FrameIndex LeggedOdometry::Impl::getLastActiveContact(const std::map<i
         }
     }
 
-    double latestTime{std::numeric_limits<double>::max()}; // assuming time cannot be negative
+    std::chrono::nanoseconds latestTime{std::chrono::nanoseconds::max()}; // assuming time cannot be
```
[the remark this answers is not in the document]
GiulioRomualdi: I tried to replicate the same logic implemented before. Here we have `std::chrono::nanoseconds::max()` since we had  `std::numeric_limits<double>::max()`. while in 

https://github.com/ami-iit/bipedal-locomotion-framework/blob/060520ccff61f563030fd5a3c31de2dd0fb52080/src/Estimators/src/LeggedOdometry.cpp#L349

we have `std::chrono::nanoseconds::min()` because before we had {-1}. I can try to modify https://github.com/ami-iit/bipedal-locomotion-framework/blob/060520ccff61f563030fd5a3c31de2dd0fb52080/src/Estimators/src/LeggedOdometry.cpp#L349 with `std::chrono::nanoseconds{-1}`

src/ParametersHandler/YarpImplementation/src/YarpImplementation.cpp
GiulioRomualdi: Ok we can support this for the ini file but not in toml since the latest support only the [RFC 3339](https://tools.ietf.org/html/rfc3339) specification

src/Planners/src/CubicSpline.cpp
GiulioRomualdi: That's true

src/Estimators/src/LeggedOdometry.cpp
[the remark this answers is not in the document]
S-Dafarra: As you prefer. Were those values meant to specifically make some searches fail? Maybe in one case, it was meant to be the first occurrence of some event, while the other was the last occurrence?

src/ParametersHandler/YarpImplementation/src/YarpImplementation.cpp
S-Dafarra: Unless we allow casting a normal double into a "time" variable :thinking:

GiulioRomualdi: As discussed with @S-Dafarra, we decided to support also the conversion from `double` to `std::chrono::nanoseconds` for both the ini https://github.com/ami-iit/bipedal-locomotion-framework/pull/630/commits/e8b59ffd71cb7de875ae10e114dd0a73bfe89a2f and toml https://github.com/ami-iit/bipedal-locomotion-framework/pull/630/commits/67e74a1e44219d36578b93384a915dc49c296787 files

src/ParametersHandler/YarpImplementation/src/YarpImplementation.cpp
GiulioRomualdi: I added the explanation of the regex in https://github.com/ami-iit/bipedal-locomotion-framework/pull/630/commits/3a7b778950664255e59f8fe8028ffd1a2c39d21c

GiulioRomualdi:
Hi @S-Dafarra I think I applied all the modification you suggested, let me know if you are fine with that 😄

S-Dafarra (commented):
Nitpicks

GiulioRomualdi:
> To be honest I was thinking something like "It matches a string of the type HH:MM:SS.xxxxxx with H, M, S, x a number from 0 to 9", but I guess that also the verbose version works.

I added also the short explanation :)

S-Dafarra (approved):

PR #640, opened 2023-03-30
Add dynamic models composing the UKF measurement model

isorrentino:
Implemented classes and tests for the following models:
- joint velocity dynamic model
- accelerometer dynamic model
- gyroscope dynamic model

The classes are used to build the UkfMeasurement object.


isorrentino:
As mentioned in https://github.com/ami-iit/bipedal-locomotion-framework/pull/615#issuecomment-1523654641, I'm going to close this PR without merging it.

PR #652, opened 2023-04-14
Implement a class to perform inference with the MANN network

GiulioRomualdi:
Before proceeding with the review of the PR I would like to discuss with you (@traversaro @S-Dafarra @paolo-viceconte) how to handle these two facts:
1. In 7172fa3656d717522427d44484bb50c95bafa35e I implemented the `Findonnxruntime.cmake`. I tested only on linux. A possibility is to get `onnxruntime` with fetchcontent however:
   1.  the installation procedure is system dependent. In linux we have to unzip a compressed folder and move the `.so` and `h` in the installation folder. In windows via [`nuget`](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime)
   2. The installation does not ship any cmake machinery so I don't know how to handle the linking of the libraries in our framework (probably with [`add_library(<name> <type> IMPORTED [GLOBAL])`](https://cmake.org/cmake/help/latest/command/add_library.html#imported-libraries)
2. In 421da3123c14069c46189d9d0e995bb9af1695ca I also stored the model for the test. This is pretty heavy (6mb). we may think of storing the model somewhere perhaps in `mann-pytorch` and take with `fetchcontent`


S-Dafarra:
> 1. the installation procedure is system dependent. In linux we have to unzip a compressed folder and move the `.so` and `h` in the installation folder. In windows via [`nuget`](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime)

I guess we can stick to linux only for the time being



> 2\. In [421da31](https://github.com/ami-iit/bipedal-locomotion-framework/commit/421da3123c14069c46189d9d0e995bb9af1695ca) I also stored the model for the test. This is pretty heavy (6mb). we may think of storing the model somewhere perhaps in `mann-pytorch` and take with `fetchcontent`

That's not desirable. Can't you have a smaller model?

traversaro:
Regarding the installation procedure, I would go for:
* Binary provided by onnxruntime for Linux/apt . I hoped it was static, but even shared will work fine. Discussing with @paolo-viceconte and @GiulioRomualdi I suggested to put this installation on blf, but probably it make more sense to have it on the superbuild.
* For conda-forge, just install onnxruntime from conda-forge . At the moment onnxruntime is not available on Windows, but for now we can just enable it on Linux/macOS .
* On vcpkg and homebrew we do not support this, as life is already too short to try to support homebrew and vcpkg.

traversaro:
> Binary provided by onnxruntime for Linux/apt . I hoped it was static, but even shared will work fine. Discussing with @paolo-viceconte and @GiulioRomualdi I suggested to put this installation on blf, but probably it make more sense to have it on the superbuild.

Done in https://github.com/robotology/robotology-superbuild/pull/1387 .

GiulioRomualdi:
The PR is finally ready to be reviewed. The model is now downloaded while compiling 🚀

src/ML/src/MANN.cpp
```diff
@@ -0,0 +1,385 @@
...
+    {
+        return false;
+    }
+
+    // If one element of the covariance is less then the threshold we set it to 1
+    constexpr double threshold = 1e-4;
+    m_pimpl->structuredInput.std
+        = (m_pimpl->structuredInput.std.array() < threshold).select(1, m_pimpl->structuredInput.std);
+    m_pimpl->structuredOutput.std = (m_pimpl->structuredOutput.std.array() < threshold)
+                                        .select(1, m_pimpl->structuredOutput.std);
```
S-Dafarra, line 196: Why 1 and not the threshold?

src/ML/src/MANN.cpp
```diff
@@ -0,0 +1,385 @@
...
+    m_pimpl->output.jointPositions.resize(numberOfJoints);
+    m_pimpl->output.jointVelocities.resize(numberOfJoints);
+    m_pimpl->output.projectedBaseVelocity = manif::SE2Tangentd::Zero();
+
+    return true;
+}
+
+bool MANN::setInput(const MANNInput& input)
+{
+    return m_pimpl->populateInput(input);
```
S-Dafarra, line 319: Maybe it is worth checking that ``initialize`` has been called first.

src/ML/src/MANN.cpp
```diff
@@ -0,0 +1,385 @@
...
+              const std::size_t rows = matrix.rows();
+              const std::size_t cols = matrix.cols();
+
+              Eigen::Ref<const Eigen::VectorXd> mean
+                  = m_pimpl->structuredOutput.mean.segment(variable.offset, variable.size);
+              Eigen::Ref<const Eigen::VectorXd> std
+                  = m_pimpl->structuredOutput.std.segment(variable.offset, variable.size);
+
+              // TODO try to avoid to do memory allocation
+              Eigen::VectorXd temp
```
S-Dafarra, line 340: You could have a map with key the name (e.g. "future_base_positions") and as value a buffer. The call to this map can be hidden in the labda (since you are already capturing ``this``). You can avoid to initialize it, but at least it will survive after the first iteration.

src/ML/src/MANN.cpp
```diff
@@ -0,0 +1,385 @@
...
+
+bool MANN::isOutputValid() const
+{
+    return true;
+}
+
+const MANNOutput& MANN::getOutput() const
+{
+    return m_pimpl->output;
+}
```
S-Dafarra, line 385: Maybe you should check that ``output`` has been filled at least once

src/ML/tests/MANNTest.cpp
```diff
@@ -0,0 +1,635 @@
...
+                                              0.4268539481560512,
+                                              0.2815972129434677,
+                                              0.46879425595722346,
+                                              0.6579274923872912,
+                                              0.0,
+                                              3.5086802986759024e-05,
+                                              0.0,
+                                              0.35984343424788334,
+                                              0.29772736861306426,
+                                              0.5900169010665167});
```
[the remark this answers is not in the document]
GiulioRomualdi: You're right. This is mitigated by https://github.com/ami-iit/mann-pytorch/pull/5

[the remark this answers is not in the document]
GiulioRomualdi: Removed :)

src/ML/src/MANN.cpp
GiulioRomualdi: Done in https://github.com/ami-iit/bipedal-locomotion-framework/pull/652/commits/31b8e91aa7116c6b5642bd93c81420bd69fc8cb6

src/ML/src/MANN.cpp
GiulioRomualdi: After https://github.com/ami-iit/mann-pytorch/pull/5 I cleaned the code and temp is not required anymore :)

GiulioRomualdi:
I think I applied all your suggestions, so in theory, the PR could be merged. However we need https://github.com/ami-iit/mann-pytorch/pull/5 to got merged before this one

S-Dafarra (approved):

PR #655, opened 2023-04-20
Catch2 is now downloaded with fetchcontent if BUILD_TESTING is set to on

GiulioRomualdi:
When this PR will get merged there will be no need to install catch2 to compile the tests.

cc @G-Cervettini


cmake/AddBipedalLocomotionUnitTest.cmake
```diff
@@ -5,6 +5,23 @@
...
+  FetchContent_Declare(Catch2
+    GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+    GIT_TAG        v2.13.8)
+
+  FetchContent_GetProperties(Catch2)
+  if(NOT Catch2_POPULATED)
+    message(STATUS "Fetching Catch2...")
+    FetchContent_MakeAvailable(Catch2)
+  endif()
+endif()
```
S-Dafarra, line 23: What about the case in which ``Catch2`` is already available in the system?

GiulioRomualdi: I've tested it in my system. The version installed in the system is not considered.

cmake/AddBipedalLocomotionUnitTest.cmake
```diff
@@ -5,6 +5,23 @@
...
 
+# This software may be modified and distributed under the terms of the
+# GNU Lesser General Public License v2.1 or any later version.
+
+# Fetch catch2 only if the testing are built
+if (BUILD_TESTING)
+  include(FetchContent)
+  FetchContent_Declare(Catch2
+    GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+    GIT_TAG        v2.13.8)
```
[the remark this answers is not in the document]
S-Dafarra: From the tests side, you probably just need to change an inclusion. The rest is on the cmake side. See for example 
- https://github.com/ami-iit/matio-cpp/pull/56
- https://github.com/ami-iit/element_jet-cat-turbines/pull/416

Maybe it is worth a shot, and if something needs extra care, we can postpone it. Up to you :wink:

cmake/AddBipedalLocomotionUnitTest.cmake
S-Dafarra: You could add an option to allow using the system one. This can be useful if you want to test something with a different version of Catch. See for example https://github.com/ami-iit/matio-cpp/blob/1b41fe4705b4f3cb63732772ca3d5c408d6fb09b/CMakeLists.txt#L103-L118

cmake/AddBipedalLocomotionUnitTest.cmake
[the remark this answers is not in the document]
GiulioRomualdi: The entire repo has been updated to catch2 v3 https://github.com/ami-iit/bipedal-locomotion-framework/pull/655/commits/9d7582295758ac687fc5e8228526f767d2320b0b

cmake/AddBipedalLocomotionUnitTest.cmake
GiulioRomualdi: Done in https://github.com/ami-iit/bipedal-locomotion-framework/commit/9d7582295758ac687fc5e8228526f767d2320b0b

traversaro (approved):

S-Dafarra (approved):

PR #659, opened 2023-04-28
Add the possibility to use MANN in autoregressive fashion

GiulioRomualdi:
This PR implements the possibility of calling MANN in autoregressive fashion.

The following diagram shows how the blocks are connected. The output of MANN is given to two blocks, one computes the kinematically feasible base position, while the other blends it with te input provided by the user

image

Drawio diagram: [adherent_autregressive.zip](https://github.com/ami-iit/bipedal-locomotion-framework/files/11430346/adherent_autregressive.zip)


src/ML/src/MANNAutoregressive.cpp
```diff
@@ -0,0 +1,706 @@
...
+    ContinuousDynamicalSystem::ForwardEuler<ContinuousDynamicalSystem::SO3Dynamics> integrator;
+
+    Eigen::Vector3d gravity;
+
+    AutoregressiveState state;
+
+    struct ContactInfo
+    {
+        Contacts::DiscreteGeometryContact discreteGeometryContact;
+        Contacts::EstimatedContact* estimatedContactPtr; /**< Pointer to the estimated contact saved
```
S-Dafarra, line 45:
```suggestion
        Contacts::EstimatedContact* estimatedContactPtr{nullptr}; /**< Pointer to the estimated contact saved
```

src/ML/src/MANNAutoregressive.cpp
```diff
@@ -0,0 +1,706 @@
...
+    if (m_pimpl->fsmState != Impl::FSM::Reset && m_pimpl->fsmState != Impl::FSM::Running)
+    {
+        log()->error("{} You need to reset the network before calling advance the first time. "
+                     "Please call reset().",
+                     logPrefix);
+        return false;
+    }
+
+    // perform one iteration of MANN
+    if (!m_pimpl->mann.advance() || !m_pimpl->mann.isOutputValid())
```
S-Dafarra, line 493: What if I forgot to call ``initialize``?

src/ML/src/MANNAutoregressive.cpp
```diff
@@ -0,0 +1,706 @@
...
+}
+
+bool MANNAutoregressive::setInput(const Input& input)
+{
+    // If the output of the network is valid, then the network has been called at least once
+    constexpr auto logPrefix = "[MANNAutoregressive::setInput]";
+
+    if (m_pimpl->fsmState != Impl::FSM::Reset && !this->isOutputValid())
+    {
+        log()->error("{} The network has not been reset. Please call reset().", logPrefix);
```
S-Dafarra, line 341: Can't we do this automatically?

src/ML/src/MANNAutoregressive.cpp
```diff
@@ -0,0 +1,706 @@
...
+
+    // get the output
+    const auto& mannOutput = m_pimpl->mann.getOutput();
+
+    // the joint positions and velocities are considered as new input
+    m_pimpl->mannInput.jointPositions = m_pimpl->mann.getOutput().jointPositions;
+    m_pimpl->mannInput.jointVelocities = m_pimpl->mann.getOutput().jointVelocities;
+
+    // TODO here we assume that the x direction of the chest frame points forward. This may
+    // cause issues on some robots (e.g., iCubV2.5)
```
S-Dafarra, line 358: One suggestion could be to get the forward direction from config file (maybe passing directly a char 'x', 'y' or 'z')

src/ML/src/MANNAutoregressive.cpp
GiulioRomualdi: I updated the state machine. Now it is impossible to call reset if the advanceable has not been initialized. So in this case advance it will not run since the MANNAutoregressive will not be in the reset state.

src/ML/src/MANNAutoregressive.cpp
GiulioRomualdi: Reset needs some information required only if we want to reset the network, for instance, the past trajectory. They should be provided by the user

src/ML/tests/CMakeLists.txt
```diff
@@ -7,19 +7,21 @@ configure_file("${CMAKE_CURRENT_SOURCE_DIR}/FolderPath.h.in" "${CMAKE_CURRENT_BI
...
+#   set(MANN_ONNX_UPDATED FALSE)
+# endif()
+
+# if(NOT ${MANN_ONNX_UPDATED})
+#   message(STATUS "Fetching MANN onnx model from ami-iit/mann-pytorch...")
+#   file(DOWNLOAD https://raw.githubusercontent.com/ami-iit/mann-pytorch/7386952b61440c19459bbe680b818faf7f461ede/models/storage_20220909-131438/models/model_49.onnx
+#     ${CMAKE_CURRENT_BINARY_DIR}/model.onnx
+#     EXPECTED_MD5 ${MANN_ONNX_EXPECTED_MD5})
+# endif()
+
```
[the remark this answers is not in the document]
GiulioRomualdi: Removed thanks!

src/ML/src/MANNAutoregressive.cpp
GiulioRomualdi: Fixed thanks :)

src/ML/include/BipedalLocomotion/ML/MANNAutoregressive.h
```diff
@@ -0,0 +1,224 @@
...
+    MANNAutoregressive();
+
+    /**
+     * Destructor
+     */
+    ~MANNAutoregressive();
+
+    /**
+     * Set the robot model.
+     * @param model model of the robot considered by the network. Please load the very same model
```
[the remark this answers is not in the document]
GiulioRomualdi: In theory yes but I think it requires modification in pythorch-mann repo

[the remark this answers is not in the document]
GiulioRomualdi: https://github.com/ami-iit/mann-pytorch/issues/6

src/ML/src/MANNAutoregressive.cpp
S-Dafarra: I see, nevermind then

src/ML/src/MANNAutoregressive.cpp
GiulioRomualdi: Yes I agree, however, since we are really in the early stage we can proceed in this way:
1. expose the `forward)_direction` in the configuration file. the parameter will be a string contacting 'x', 'y' or 'z', and for the time being we support only 'x' .(i.e., it will return an error if the parameter is different from 'x'. So at least we avoid to give the user the possibility to test the network on robots in which the x-direction points forward

S-Dafarra: Sounds good!

S-Dafarra (approved):

PR #661, opened 2023-05-02
Implement robot dynamics estimator - PR1

isorrentino:
Hi @GiulioRomualdi and @S-Dafarra, as anticipated in https://github.com/ami-iit/bipedal-locomotion-framework/pull/615#issuecomment-1523654641 here is a first PR.

This PR modifies some interfaces and adds some methods to the already merged classes `SubModel` and `SubModelKinDynWrapper`.

### DIFF
--- a/src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/SubModel.h
+++ b/src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/SubModel.h
@@ -11,6 +11,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <unordered_map>
 
 // iDynTree
 #include <iDynTree/KinDynComputations.h>
@@ -56,6 +57,7 @@ struct FT : Sensor
 
     Direction forceDirection = Direction::NotSpecified; /**< Force direction depending on which side
                                                            of the sensor is considered (+1 or -1)*/
+    std::string associatedJoint; /**< Name of the fixed joint used to represent the ft sensor in the model. */
 };
 
 /**
@@ -67,9 +69,9 @@ class SubModel
     iDynTree::Model m_model; /**< iDynTree Model object describing the submodel */
     std::vector<int> m_jointListMapping; /**< Each element contains an index describing the joint
                                             position in the full model*/
-    std::vector<FT> m_ftList; /**< List of force/torque sensors in the submodel */
-    std::vector<Sensor> m_accelerometerList; /**< List of accelerometers in the submodel */
-    std::vector<Sensor> m_gyroscopeList; /**< List of gyroscopes in the submodel */
+    std::unordered_map<std::string, FT> m_ftList; /**< List of force/torque sensors in the submodel */
+    std::unordered_map<std::string, Sensor> m_accelerometerList; /**< List of accelerometers in the submodel */
+    std::unordered_map<std::string, Sensor> m_gyroscopeList; /**< List of gyroscopes in the submodel */
     std::vector<std::string> m_externalContactList; /**< List of the additional external contacts */
 
 public:
@@ -99,21 +101,21 @@ public:
 
     /**
      * @brief Access the `std::vector<FT>` list.
-     * @return the list of FT objects which is the list of force/torque sensors.
+     * @return a map of (string, FT) objects which is the list of force/torque sensors.
      */
-    const std::vector<FT>& getFTList() const;
+    const std::unordered_map<std::string, FT>& getFTList() const;
 
     /**
      * @brief Access the `std::vector<Sensor>` list of acceletometer sensors.
-     * @return a list of Sensor objects describing the accelerometers contained in the sub-model.
+     * @return a map of (string, Sensor) objects describing the accelerometers contained in the sub-model.
      */
-    const std::vector<Sensor>& getAccelerometerList() const;
+    const std::unordered_map<std::string, Sensor>& getAccelerometerList() const;
 
     /**
      * @brief Access the `std::vector<Sensor>` list of gyroscope sensors.
-     * @return a list of Sensor objects describing the gyroscope contained in the sub-model.
+     * @return a map of (string, Sensor) objects describing the gyroscope contained in the sub-model.
      */
-    const std::vector<Sensor>& getGyroscopeList() const;
+    const std::unordered_map<std::string, Sensor>& getGyroscopeList() const;
 
     /**
      * @brief Access the `std::vector<std::string>` list of frame names.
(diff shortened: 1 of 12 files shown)


src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/SubModel.h
```diff
@@ -56,6 +56,7 @@ struct FT : Sensor
 
     Direction forceDirection = Direction::NotSpecified; /**< Force direction depending on which side
                                                            of the sensor is considered (+1 or -1)*/
+    std::string associatedJoint; /**< Name of the fixd joint used to represent the ft sensor in the model. */
```
S-Dafarra, line 59:
```suggestion
    std::string associatedJoint; /**< Name of the fixed joint used to represent the ft sensor in the model. */
```

src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/SubModel.h
```diff
@@ -147,24 +148,70 @@ class SubModel
...
      * @return FT object associated with the specified index.
      */
     const FT& getFTSensor(const int index) const;
 
+    /**
+     * @brief Access an element of the force/torque sensor list.
+     * @param is the name of the force/torque sensor.
+     * @return FT object associated with the specified name.
+     */
+    const FT& getFTSensor(const std::string name) const;
```
S-Dafarra, line 161: I would suggest passing by reference. This hold also for all the methods below where you pass a ``const std::string`` instead of a ``const std::string&``

src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/SubModel.h
```diff
@@ -147,24 +148,70 @@ class SubModel
...
     /**
      * @brief Access an element of the gyroscope list.
+     * @param index is the index of the gyroscope in the submodel
      * @return a Sensor object corresponding to the gyroscope associated with the specified index.
      */
     const Sensor& getGyroscope(const int index) const;
 
+    /**
+     * @brief Access an element of the gyroscope list.
+     * @param is the name of the gyroscoper.
```
S-Dafarra, line 200:
```suggestion
     * @param is the name of the gyroscope.
```

src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/SubModelKinDynWrapper.h
```diff
@@ -81,7 +85,12 @@ class SubModelKinDynWrapper
      * updateDynamicsVariableState updates the value of all the member variables containing
      * information about the robot kinematics and dynamics
      */
-    bool updateDynamicsVariableState();
+    bool updateDynamicsVariableState(bool updateRobotDynamicsOnly);
```
S-Dafarra, line 88: Usually boolean argument is not really readable.
Imagine someone reading your code for the first time
```cpp
updateDynamicsVariableState(false);
```
can him/her guess what that ``false`` does?

Usually it is more convenient to use an enum class. Like
```cpp
enum class UpdateMode {
    Full,
    RobotDynamicsOnly
}
```
In this way, the code will become
```cpp
updateDynamicsVariableState(UpdateMode::Full);
```
that is more readable, and will be also easily extendible.

In this case, the method is protected so it should not matter a lot, but it is something you may want to consider.

src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/SubModelKinDynWrapper.h
```diff
@@ -102,10 +111,15 @@ class SubModelKinDynWrapper
...
+     * @brief updateState updates the state of the KinDynWrapper object.
+     * @param robotBaseAcceleration is a manif::SE3d::Tangent representing the robot base acceleration.
+     * @param robotJointAcceleration is a Eigen reference to a Eigen::VectorXd containing the joint accelerations.
+     * @param isCorrectStep is a boolean saying if the method is called during the predict step or the correct step.
      * @return a boolean value saying if the subModelList has been created correctly.
      */
-    bool updateInternalKinDynState();
+    bool updateState(manif::SE3d::Tangent& robotBaseAcceleration,
+                     Eigen::Ref<const Eigen::VectorXd> robotJointAcceleration,
+                     bool isCorrectStep);
```
S-Dafarra, line 122: Same as above for the boolean

src/Estimators/src/SubModel.cpp
```diff
@@ -494,16 +503,100 @@ const RDE::FT& RDE::SubModel::getFTSensor(const int index) const
     return m_ftList.at(index);
 }
 
+const RDE::FT& RDE::SubModel::getFTSensor(const std::string name) const
+{
+    for (int ftIndex = 0; ftIndex < m_ftList.size(); ftIndex++)
```
S-Dafarra, line 508: You might be interested in using an ``unordered_map`` . In this way, instead of performing a linear search, you can simply access the FT by name (provided that exists, but you can check it with the ``find`` method). This in general is beneficial when you have a lot of entries, but maybe it is not the case here. Another  difference is that the order is no more preserved. But if you are also interested in the order, you can simply have an additional vector with the associated names in the given order, or with the pointers to corresponding objects.

The same holds below

src/Estimators/include/BipedalLocomotion/RobotDynamicsEstimator/SubModel.h
```diff
@@ -67,9 +68,9 @@ class SubModel
     iDynTree::Model m_model; /**< iDynTree Model object describing the submodel */
     std::vector<int> m_jointListMapping; /**< Each element contains an index describing the joint
                                             position in the full model*/
-    std::vector<FT> m_ftList; /**< List of force/torque sensors in the submodel */
-    std::vector<Sensor> m_accelerometerList; /**< List of accelerometers in the submodel */
-    std::vector<Sensor> m_gyroscopeList; /**< List of gyroscopes in the submodel */
+    std::unordered_map<std::string, FT> m_ftList; /**< List of force/torque sensors in the submodel */
```
S-Dafarra, line 71: Consider adding 
```cpp
#include <unordered_map>
```

src/Estimators/src/SubModel.cpp
```diff
@@ -489,19 +498,73 @@ std::size_t RDE::SubModel::getNrOfExternalContact() const
...
 
-const RDE::FT& RDE::SubModel::getFTSensor(const int index) const
+const RDE::FT& RDE::SubModel::getFTSensor(const std::string& name)
+{
+    if (m_ftList.find(name) != m_ftList.end())
+    {
+        return m_ftList[name];
+    }
+
+    log()->error("[RobotDynamicsEstimator::getFTSensor] Sensor not found.");
```
S-Dafarra, line 508: I would suggest adding the name of the sensor that has not been found. The same below

src/Estimators/src/SubModel.cpp
```diff
@@ -489,19 +498,73 @@ std::size_t RDE::SubModel::getNrOfExternalContact() const
     return m_externalContactList.size();
 }
 
-const RDE::FT& RDE::SubModel::getFTSensor(const int index) const
+const RDE::FT& RDE::SubModel::getFTSensor(const std::string& name)
+{
+    if (m_ftList.find(name) != m_ftList.end())
```
S-Dafarra, line 503: Note that ``find`` already gives you the iterator to the object, if it has been found. I would suggest using its output if it is valid. The same holds below

src/Estimators/src/SubModel.cpp
```diff
@@ -489,19 +498,73 @@ std::size_t RDE::SubModel::getNrOfExternalContact() const
...
+}
+
+bool RDE::SubModel::hasFTSensor(const std::string& name) const
+{
+    if (m_ftList.find(name) != m_ftList.end())
+    {
+        return true;
+    }
+
+    return false;
```
S-Dafarra, line 521: This could be simply
```cpp
return m_ftList.find(name) != m_ftList.end();
```

The same holds below

src/Estimators/tests/RobotDynamicsEstimator/SubModelKinDynWrapperTest.cpp
```diff
@@ -264,11 +264,11 @@ TEST_CASE("SubModelKinDynWrapper")
...
                     Eigen::MatrixXd jac = jacobian.block(0, 6, 6, subModelList[idx].getModel().getNrOfDOFs());
 
-                    for (int ftIndex = 0; ftIndex < subModelList[idx].getNrOfFTSensor(); ftIndex++)
+                    for (auto & [ftName, ftObj] : subModelList[idx].getFTList())
                     {
-                        if (subModelList[idx].getFTSensor(ftIndex).name == key)
+                        if (ftName == key)
                         {
-                            contactTorques.noalias() += (int)subModelList[idx].getFTSensor(ftIndex).forceDirection * jac.transpose() * val;
+                            contactTorques.noalias() += (int)ftObj.forceDirection * jac.transpose() * val;
```
S-Dafarra, line 271: Consider using a ``static_cast``  (i.e. ``static_cast<int>(ftObj.forceDirection)``) instead of a classical casting. It is a little bit safer

GiulioRomualdi (approved):

GiulioRomualdi:
The failure of the CI is not related to this PR. Merging

PR #671, opened 2023-05-15
Add fts and imus of the legs to YarpRobotLoggerDevice

isorrentino:


isorrentino:
I tested the PR on the robot. It works fine. @GiulioRomualdi

PR #684, opened 2023-06-06
Use schmitt trigger to detect the contat activation and deactivation in MANN

GiulioRomualdi:
During our testing of MANNTrajectoryGenerator with @paolo-viceconte, we encountered an issue where unexpected contact was being generated. This was caused by the undesirable switch-bouncing effect triggered by MANN, which refers to the imperfect behavior of a switch leading to multiple transitions for a single input.

Switch-Bouncing-in-the-Circuit

To address this issue, we decided to filter the contact using a Schmitt trigger, which is already available in blf 🎉 . However, we faced a challenge in utilizing the Schmitt trigger within MANNAutoregressive. Specifically, we needed to implement the ability to completely control the internal state of the Schmitt trigger, including all internal timers.

Furthermore, the switch off and on parameters of the Schmitt trigger were set to negative values (-0.01 and -0.01) since we consider the negative of the foot height. In our context, the output of the Schmitt trigger is considered active when there is a contact. Therefore, a value of -0.01 indicates that the contact is active if the distance between the foot and the ground is lower than 1cm.


src/ML/src/MANNAutoregressive.cpp
```diff
@@ -607,10 +664,11 @@ bool MANNAutoregressive::advance()
     // we update the projected contact position in the world frame if and only if
     // 1. The support foot changed
     // 2. Tue support foot did not change but the lower corner yes
-    if ((supportFootPtr != m_pimpl->supportFootPtr)
-        || (supportFootPtr == m_pimpl->supportFootPtr
-            && supportCornerIndex != m_pimpl->supportCornerIndex))
+    if ((supportFootPtr != m_pimpl->supportFootPtr))
+    // || (supportFootPtr == m_pimpl->supportFootPtr
+    //     && supportCornerIndex != m_pimpl->supportCornerIndex))
```
S-Dafarra, line 669: Commented code

src/ML/src/MANNAutoregressive.cpp
```diff
@@ -663,54 +721,32 @@ bool MANNAutoregressive::advance()
...
-    }
 
-    // store the output
-    m_pimpl->output.leftFoot.lastUpdateTime = m_pimpl->currentTime;
-    m_pimpl->output.rightFoot.lastUpdateTime = m_pimpl->currentTime;
+    m_pimpl->updateContact(-leftFootHeight,
+                           m_pimpl->leftFootSchmittTrigger,
+                           m_pimpl->output.leftFootSchmittTriggerState,
+                           m_pimpl->output.leftFoot);
+    m_pimpl->updateContact(-rightFootHeight,
```
S-Dafarra, line 736: I would sugggest adding a contact to explain why the minus

src/ML/src/MANNTrajectoryGenerator.cpp
```diff
@@ -361,6 +371,15 @@ bool MANNTrajectoryGenerator::advance()
...
 
+
+/*     for (const auto[key, list] : m_pimpl->contactListMap)
+    {
+        std::cerr << key << std::endl;
+
+        for(auto contact : list)
+            std::cerr << "position " << contact.pose.translation().transpose() <<  " activation " << std::chrono::duration<double>(contact.activationTime).count() << " deactivation " << std::chrono::duration<double>(contact.deactivationTime).count() << std::endl;
+    }
+ */
```
S-Dafarra, line 382: Commented code

src/ML/tests/MANNTest.cpp
```diff
@@ -81,17 +81,15 @@ TEST_CASE("MANN")
...
-        -0.02287937013013993, -0.002927608219628366, 0.018047571904430328, -0.01034426546939903;
+    input.jointPositions << -0.10922017141063572, 0.05081325960010118, 0.06581966291990003,
+        -0.0898053099824925, -0.09324922528169599, -0.05110058859172172, -0.11021232812838086,
+        0.054291515925228385, 0.0735575862560208, -0.09509332143185895, -0.09833823347493076,
+        -0.05367281245082792, 0.1531558711397399, -0.001030634273454133, 0.0006584764419034815,
+        -0.0016821925351926288, -0.004284529460797688, 0.030389771690123243, -0.040592118429752494,
+        -0.1695472679986807, -0.20799422095574033, 0.045397975984119654, -0.03946672931050908,
+        -0.16795588539580256, -0.20911090583076936, 0.0419854257806720;
+
+    input.jointVelocities = Eigen::VectorXd::Zero(26);
```
S-Dafarra, line 92: Does it work only with zero joint velocities?

src/Math/include/BipedalLocomotion/Math/SchmittTrigger.h
```diff
@@ -95,10 +110,16 @@ class SchmittTrigger
     /**
      * Set the state of the SchmittTrigger.
      * @param state the state of the SchmittTrigger
-     * @note When the state is set the internal timer is reset as well. This function should be
-     * called only if the user want to force the state of the system.
+     * @return True in case of success, false otherwise.
+     * @note This function should be called only if the user want to force the state of the system.
+     */
+    bool setState(const SchmittTriggerState& state);
```
S-Dafarra, line 116: Maybe ``forceState``?

S-Dafarra:
> Furthermore, the switch off and on parameters of the Schmitt trigger were set to negative values (-0.01 and -0.01) since we consider the negative of the foot height. In our context, the output of the Schmitt trigger is considered active when there is a contact. Therefore, a value of -0.01 indicates that the contact is active if the distance between the foot and the ground is lower than 1cm.

Just as an additional comment, this is an implementation-dependent choice. If the user can set this value, I would suggest forcing it somehow inside. I feel like it breaks the "principle of least astonishment" a little :thinking:

src/ML/tests/MANNTest.cpp
GiulioRomualdi: no 😃 if you prefer, I can pass a random value for the joint state (position and velocity)

S-Dafarra: I see. You can also keep the previous values (i.e. random but fixed values)

GiulioRomualdi:
> > Furthermore, the switch off and on parameters of the Schmitt trigger were set to negative values (-0.01 and -0.01) since we consider the negative of the foot height. In our context, the output of the Schmitt trigger is considered active when there is a contact. Therefore, a value of -0.01 indicates that the contact is active if the distance between the foot and the ground is lower than 1cm.
> 
> Just as an additional comment, this is an implementation-dependent choice. If the user can set this value, I would suggest forcing the sign somehow inside. I feel like it breaks the "principle of least astonishment" a little thinking

I agree since it is counterintuitive I decided to ask the user to pass a positive number, since the foot height is always positive (at least in this scenario), and then convert it into negative to simplify the usage of the trigger inside the class

src/ML/src/MANNTrajectoryGenerator.cpp
GiulioRomualdi: Ops, removed in 5d9cc83df41de42783b30f656afa13314b0b58f3. Thank you 😄

src/ML/src/MANNAutoregressive.cpp
GiulioRomualdi: Done here: https://github.com/ami-iit/bipedal-locomotion-framework/commit/4634f98306381828c9eb9bf471e4402a3f970ad2

src/ML/src/MANNAutoregressive.cpp
GiulioRomualdi: Fixed in 7461bf0f7. Moreover, I added a comment that explains the rationale behind the if statement.

src/ML/tests/MANNTest.cpp
GiulioRomualdi: Done here: 7c582cf18

GiulioRomualdi:
Hi @S-Dafarra, I modified what you suggested except for https://github.com/ami-iit/bipedal-locomotion-framework/pull/684#discussion_r1221037158. Let me know if you are fine with this

paolo-viceconte (approved):
Just a couple of typos spotted, LGTM!

PR #692, opened 2023-06-23
Make the YarpSensorBridge compatible with MAS that use the same name between gyro ac magn and orientation

GiulioRomualdi:
This PR fixes https://github.com/ami-iit/bipedal-locomotion-framework/issues/638

⚠️ Before merging the PR I would like to test it on the robot.


GiulioRomualdi:
Tested on the robot, everything is working as expected

PR #703, opened 2023-07-10
Restructuring ReducedModelControllers 

mebbaid:
In this PR I will add the `StableCentroidalMPC` as derived class from a `BaseCentroidalMPC` together with keeping the original `CentroidalMPC` implementation unchanged.

The idea is also to add the `ParametrizedCentroidalMPC` derived class. However I will leave that to a separate PR till I fix the associated unit test issue.


mebbaid:
Closing in favor of #734

PR #742, opened 2023-10-24
Remove outdated tests - PR2

isorrentino:


S-Dafarra (approved):

GiulioRomualdi:
Hi @isorrentino it's not clear to me why they our outdated. If I'm not mistaken they are compiled and ran in the CI

isorrentino:
> Hi @isorrentino it's not clear to me why they our outdated. If I'm not mistaken they are compiled and ran in the CI

I do not know how they can run since one of the classes used by most of those tests does not exist anymore. For instance, https://github.com/ami-iit/bipedal-locomotion-framework/blob/935657dc6ab1c22853285c4ccaba49c8aafadbd4/src/Estimators/tests/RobotDynamicsEstimator/CMakeLists.txt#L14 is the test for the `SubModelKinDynWrapper` class which is no longer defined in the master branch (the latest PR removed it). Furthermore, many tests import the `SubModelKinDynWrapper` class and cannot run. It's not clear to me why the CI was not failing.

PR #745, opened 2023-10-25
Implement GlobalCoPEvaluator in contact Component

GiulioRomualdi:
This PR implements the `GlobalZMPEvaluator` class that can be used to compute the local and global ZMP given a set of contacts.

The class implements the same checks already in place in the walking-controller class. Such as:
- check for constant ZMP for a given amount of samples
- outlier rejection in the ZMP evaluation 

I implemented also the associated python bindings

⚠️ I was not able to find a better name for the class so if you find something feel free to suggest it :)


S-Dafarra:
Nitpick: the definition of the ZMP is slightly different, as it considers the acceleration of the CoM for example: see Eq. 21 of [1]. I think it would be more appropriate to use the CoP notation instead. The code of https://github.com/ami-iit/bipedal-locomotion-framework/blob/f7d94430b6d25c4c9cf0319ce973b662e5a5be3e/src/Contacts/src/Contact.cpp#L45-L47 corresponds to the definition of the CoP in case of planar ground. See Eq. (13) of [1]

## 
[1]: [Forces acting on a biped robot. Center of pressure-zero moment point](https://ieeexplore.ieee.org/document/1325327)

GiulioRomualdi:
After discussing with @S-Dafarra we decided to use the term XoP instead of ZMP. 

Still, one thing should be decided. Is `GlobalCoPEvaluator` the proper name or we can find something better?

S-Dafarra:
> Still, one thing should be decided. Is `GlobalCoPEvaluator` the proper name or we can find something better?

Maybe one thing to consider is to specify somewhere that by "global" we mean a weighted average, which may not follow the definition of "CoP"

GiulioRomualdi:
Yes I'm going to explain this in the class documentation

GiulioRomualdi:
Hi @S-Dafarra, I renamed the method from ZMP to CoP, let me know what you think :)

src/Math/include/BipedalLocomotion/Math/Wrench.h
```diff
@@ -66,6 +66,31 @@ template <class Scalar> class Wrench : public Eigen::Matrix<Scalar, 6, 1>
...
+     * bodies in contact. The CoP is a local quantity defined from interaction forces at the contact
+     * surface.
+     * @note A more detailed explanation can be found in: P. Sardain and G. Bessonnet, "Forces
+     * acting on a biped robot. Center of pressure-zero moment point," in IEEE Transactions on
+     * Systems, Man, and Cybernetics - Part A: Systems and Humans, vol. 34, no. 5, pp. 630-637,
+     * Sept. 2004, doi: 10.1109/TSMCA.2004.832811. An interested reader can also check the following
+     * https://scaron.info/robotics/zero-tilting-moment-point.html
+     * @warning This function assumes that the wrench is expressed in the body frame. I.e., a frame
+     * attached to the body belonging to the contact area and the z-component of the contact force
+     * is resultant pressure force.
```
S-Dafarra, line 82:
```suggestion
     * is the resultant pressure force.
```

src/Contacts/include/BipedalLocomotion/Contacts/GlobalCoPEvaluator.h
```diff
@@ -0,0 +1,114 @@
...
+public:
+    // clang-format off
+    /**
+     * Initialize the class
+     * @param handler pointer to a parameter handler
+     * @note the following parameters are required by the class
+     * |       Parameter Name       |       Type       |                                           Description                                     | Mandatory |
+     * |:--------------------------:|:----------------:|:-----------------------------------------------------------------------------------------:|:---------:|
+     * |   `minimum_normal_force`   |     `double`     |           Minimum normal force required to consider a contact active (in N)               |    Yes    |
+     * |   `cop_admissible_limits`  | `vector<double>` |  2D vector defines CoP region, comparing absolute CoP x and y to the 1st and 2nd elements |    Yes    |
```
S-Dafarra, line 48: I think it compares the local CoP not the global one

src/Contacts/src/GlobalCoPEvaluator.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+        if (m_constantCoPCounter >= m_constantCoPMaxCounter)
+        {
+            log()->error("{} The CoP was constant (in a {} m radius) for {} times.",
+                         logPrefix,
+                         m_constantCoPTolerance,
+                         m_constantCoPCounter);
+
+            return false;
+        }
```
S-Dafarra, line 168: Maybe, we should consider the case in which ``m_constantCoPMaxCounter`` is negative. In the walking controller, if the max counter was negative, we simply skipped the check.

S-Dafarra (commented):
Minor comments

src/Contacts/include/BipedalLocomotion/Contacts/GlobalCoPEvaluator.h
```diff
@@ -0,0 +1,114 @@
...
+{
+namespace Contacts
+{
+
+/**
+ * GlobalCoPEvaluator is a class that computes the global CoP given a set of contact wrenches.
+ * Following P. Sardain and G. Bessonnet, "Forces acting on a biped robot. Center of
+ * pressure-zero moment point," in IEEE Transactions on Systems, Man, and Cybernetics,
+ * we defined the global CoP as the weighted average of the CoP of each contact, with the weight
+ * determined by the normal force of the contact.
```
traversaro, line 29: Can you point to the formula used? I could not found the proposed formula in the paper.

src/Contacts/include/BipedalLocomotion/Contacts/GlobalCoPEvaluator.h
```diff
@@ -0,0 +1,114 @@
...
+ * @note This class assumes that the contact wrenches stored in Contacts::ContactWrench list are
+ * expressed in the body frame (left trivialized).
+ * @note In addition to evaluating the CoP, this class checks that at least one contact is active
```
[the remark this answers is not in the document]
traversaro: I added some assumptions that were not obvious from reading the docs.

traversaro (approved):
Minor comments.

src/Contacts/include/BipedalLocomotion/Contacts/GlobalCoPEvaluator.h
GiulioRomualdi: Hi @traversaro it should be the following equation. 

image

Still I can add the equation in the documentation

traversaro: But that is not the " weighted average of the CoP of each contact", right? This is the COP of the sum of the contact forces, and it can be proven that the two definitions are the same, with the assumptions that all the contacts occur in the same plane. So (nitpick) I would write "Consistently with", not "Following".

src/Contacts/src/GlobalCoPEvaluator.cpp
GiulioRomualdi: Hi @S-Dafarra it's the same here 
```cpp
    if (numberOfActiveSupports > 1 && m_constantCoPMaxCounter > 0)
    {

```

S-Dafarra: Ah sorry, I did not notice it. Is it worth mentioning this case in the docs?

src/Contacts/include/BipedalLocomotion/Contacts/GlobalCoPEvaluator.h
GiulioRomualdi: Here I wrote the math https://github.com/ami-iit/bipedal-locomotion-framework/pull/745/commits/881e92d4b9dd405b8f61b9dd9c888ca952a7f3b1

src/Contacts/src/GlobalCoPEvaluator.cpp
GiulioRomualdi: I would say yes

GiulioRomualdi: Done here: https://github.com/ami-iit/bipedal-locomotion-framework/pull/745/commits/95727d6dd7e4deebbefcfd0e9fe78de93f5522ce

src/Contacts/include/BipedalLocomotion/Contacts/GlobalCoPEvaluator.h
```diff
@@ -0,0 +1,137 @@
...
+public:
+    // clang-format off
+    /**
+     * Initialize the class
+     * @param handler pointer to a parameter handler
+     * @note the following parameters are required by the class
+     * |       Parameter Name       |       Type       |                                                     Description                                              | Mandatory |
+     * |:--------------------------:|:----------------:|:------------------------------------------------------------------------------------------------------------:|:---------:|
+     * |   `minimum_normal_force`   |     `double`     |                     Minimum normal force required to consider a contact active (in N)                        |    Yes    |
+     * |   `cop_admissible_limits`  | `vector<double>` |         2D vector defines CoP region, comparing the local CoP x and y to the 1st and 2nd elements            |    Yes    |
```
S-Dafarra, line 71:
```suggestion
     * |   `cop_admissible_limits`  | `vector<double>` |         2D vector defining the admissible CoP region, comparing the local CoP x and y to the 1st and 2nd elements            |    Yes    |
```

S-Dafarra (approved):
A small comment on the doc. Feel free to proceed

GiulioRomualdi:
There was an uninitialized variable in `GlobalCoPEvaluator::advance` I fixed it in https://github.com/ami-iit/bipedal-locomotion-framework/pull/745/commits/ba458367d2a4ea249ef89e8e41fc361bd3521a36

PR #748, opened 2023-11-03
Fix timestamp logging for the cameras

GiulioRomualdi:
With this PR, I saved the timestamps of the frames collected with the YarpRobotLoggerDevice in seconds instead of nanoseconds.


S-Dafarra (approved):

GiulioRomualdi:
The CI macOS fails
```
ModuleNotFoundError: No module named 'distutils'
CMake Error at bindings/CMakeLists.txt:38 (string):
  string sub-command STRIP requires two arguments.
```
but it's not related to this PR.

PR #767, opened 2023-11-20
Implement VectorsCollectionServer and VectorsCollectionClient classes

GiulioRomualdi:
This PR substitutes https://github.com/ami-iit/bipedal-locomotion-framework/pull/732. 

This PR introduces metadata for the `VectorsCollection`, allowing each component of the vector saved by the logger to have a specific label. This enhancement simplifies data analysis, as it enables the display of associated labels in the robot-log-visualizer, as shown in the following image:

image

To ensure metadata consistency, I've implemented it using the client-server paradigm in YARP. I declared the `VectorsCollectionMetadataService` in `VectorsCollection.thrift`. This change enables asynchronous metadata requests for a specific VectorsCollection. This is done thanks to the `VectorsCollectionServer` and `VectorsCollectionClient`

In this context, the `YarpRobotLoggerDevice` functions as a client (`VectorsCollectionClient`), while the server must be implemented in the application sending the data.

Here's an example of how to use `VectorsCollectionServer` in your code:

```cpp
#include <BipedalLocomotion/YarpUtilities/VectorsCollectionServer.h> 

class Module
{
    BipedalLocomotion::YarpUtilities::VectorsCollectionServer m_vectorsCollectionServer; /**< Logger server. */
public:
 // all the other functions you need
}
```
The `m_vectorsCollectionServer` helps you to handle the data you want to send and to populate the metadata. To use this functionality, call `BipedalLocomotion::YarpUtilities::VectorsCollectionServer::populateMetadata` during the configuration phase. Once you have finished populating the metadata you should call `BipedalLocomotion::YarpUtilities::VectorsCollectionServer::finalizeMetadata`
```cpp
//This code should go into the configuration phase
auto loggerOption = std::make_shared<BipedalLocomotion::ParametersHandler::YarpImplementation>(rf);
if (!m_vectorsCollectionServer.initialize(loggerOption->getGroup("WALKING_LOGGER")))
{
    log()->error("[WalkingModule::configure] Unable to configure the server.");
    return false;
}
 
m_vectorsCollectionServer.populateMetadata("dcm::position::measured", {"x", "y"});
m_vectorsCollectionServer.populateMetadata("dcm::position::desired", {"x", "y"});

m_vectorsCollectionServer.finalizeMetadata();
```

In the main loop, add the following code to prepare and populate the data:

```cpp
m_vectorsCollectionServer.clearData(); // optional see the documentation


// DCM
m_vectorsCollectionServer.populateData("dcm::position::measured", <signal>);
m_vectorsCollectionServer.populateData("dcm::position::desired", <signal>);

m_vectorsCollectionServer.sendData();
```

⚠️ This PR breaks all the code that uses the YarpRobotLoggerDevice for logging exogenous signals (cc @G-Cervettini @isorrentino)
This is an example of updated application that allows you to correctly fill the metadata: https://github.com/robotology/walking-controllers/pull/158


src/YarpUtilities/src/VectorsCollectionClient.cpp
```diff
@@ -0,0 +1,161 @@
...
+#include <yarp/os/Port.h>
+
+#include <unordered_set>
+
+using namespace BipedalLocomotion::YarpUtilities;
+
+struct VectorsCollectionClient::Impl
+{
+    yarp::os::BufferedPort<VectorsCollection> port; /**< Buffered port used to communicate with
+                                                            the client. */
```
S-Dafarra, line 22:
```suggestion
                                                            the server. */
```

src/YarpUtilities/src/VectorsCollectionClient.cpp
```diff
@@ -0,0 +1,161 @@
...
+    yarp::os::Port rpcPort; /**< RPC port used to communicate with the client. */
```
S-Dafarra, line 23:
```suggestion
    yarp::os::Port rpcPort; /**< RPC port used to communicate with the server. */
```

src/YarpUtilities/src/VectorsCollectionServer.cpp
```diff
@@ -0,0 +1,176 @@
...
+}
+
+bool VectorsCollectionServer::areMetadataReady()
+{
+    return m_pimpl->isMetadataFinalized;
+}
+
+VectorsCollectionMetadata VectorsCollectionServer::getMetadata()
+{
+    return m_pimpl->metadata;
```
S-Dafarra, line 175: I think we need to check whether the metadata is finalized first

src/YarpUtilities/src/VectorsCollectionServer.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+struct VectorsCollectionServer::Impl
+{
+    yarp::os::BufferedPort<VectorsCollection> port; /**< Buffered port used to communicate with
+                                                            the client. */
+    yarp::os::Port rpcPort; /**< RPC port used to communicate with the client. */
+
+    VectorsCollectionMetadata metadata; /**< Metadata of the vectors collection. */
+
+    bool isMetadataFinalized{false}; /**< True if the metadata has been finalized. */
```
S-Dafarra, line 26: I would suggest using an atomic variable here to avoid concurrency issues with clients that get connected before the population of the metadata is finalized

src/YarpUtilities/src/VectorsCollectionServer.cpp
GiulioRomualdi: This is defined in the thrift. I don't know if it is possible to return two values (probably with a pair).
What do you think?

S-Dafarra: In case you could just return an empty ``VectorsCollectionMetadata``. I guess it is better than a malformed one. This would also allow avoiding concurrency issues with clients that try to access the metadata before they are finalized.

GiulioRomualdi: Ok perfect! I will go for an empty one :)

GiulioRomualdi:
Hi @S-Dafarra I modified the code as you suggested, what do you think?

src/YarpUtilities/include/BipedalLocomotion/YarpUtilities/VectorsCollectionServer.h
```diff
@@ -0,0 +1,137 @@
...
+    VectorsCollectionMetadata getMetadata() override;
+
+    /**
+     * Check if the metadata is ready.
+     * @return true if the metadata is ready, false otherwise.
+     */
+    bool areMetadataReady() override;
+
+    /**
+     * Send the data filled with VectorsCollection::setData.
```
S-Dafarra, line 115:
```suggestion
     * Send the data filled with populateData
```

src/YarpUtilities/include/BipedalLocomotion/YarpUtilities/VectorsCollectionServer.h
```diff
@@ -0,0 +1,137 @@
...
+     * @param forceStrict If this is true, wait until any previous sends are complete. If false, the
+     * current object will not be sent on connections that are currently busy.
+     */
+    void sendData(bool forceStrict = false);
+
+    /**
+     * Deallocates the memory held by the internal buffer used to store data.
+     * @note If you have previously populated the object using VectorsCollection::setData and intend
```
S-Dafarra, line 123:
```suggestion
     * @note If you have previously populated the object using populateData and intend
```

src/YarpUtilities/src/VectorsCollectionServer.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    // check if the key exists
+    if (m_pimpl->setOfKeys.find(key) == m_pimpl->setOfKeys.end())
+    {
+        log()->error("{} The key {} does not exist.", logPrefix, key);
+        return false;
+    }
+
+    // prepare the data
+    BipedalLocomotion::YarpUtilities::VectorsCollection& collection = m_pimpl->port.prepare();
```
S-Dafarra, line 152: I just realized this. Wasn't it a problem to call ``prepare`` multiple times?

GiulioRomualdi: Well, I tested and everything seems working fine

S-Dafarra: I dug a little bit inside the ``prepare`` mechanism, and I guess this might explode in memory. Here is the trace of calls:
1. https://github.com/robotology/yarp/blob/5b7014770df1da99c48b4dd27871b5a7a3dc5655/src/libYARP_os/src/yarp/os/BufferedPort-inl.h#L113
2. https://github.com/robotology/yarp/blob/5b7014770df1da99c48b4dd27871b5a7a3dc5655/src/libYARP_os/src/yarp/os/PortWriterBuffer.h#L98-L102
3. https://github.com/robotology/yarp/blob/5b7014770df1da99c48b4dd27871b5a7a3dc5655/src/libYARP_os/src/yarp/os/PortWriterBufferBase.cpp#L185
4. https://github.com/robotology/yarp/blob/5b7014770df1da99c48b4dd27871b5a7a3dc5655/src/libYARP_os/src/yarp/os/PortWriterBufferBase.cpp#L69-L93
5. https://github.com/robotology/yarp/blob/5b7014770df1da99c48b4dd27871b5a7a3dc5655/src/libYARP_os/src/yarp/os/impl/PortCorePackets.cpp#L35-L54

In particular, I suspect that the problem is with https://github.com/robotology/yarp/blob/5b7014770df1da99c48b4dd27871b5a7a3dc5655/src/libYARP_os/src/yarp/os/impl/PortCorePackets.cpp#L37-L42, where, if it does not find any ``inactive`` packet, it allocates a new one.

At this point, to be on the safe side, I would suggest keeping an internal buffer, fill it in with every ``populateData``, and then do the ``prepare`` and a copy in the ``sendData``. What do you think?

S-Dafarra: Actually, it should not explode. I just saw https://github.com/robotology/yarp/blob/5b7014770df1da99c48b4dd27871b5a7a3dc5655/src/libYARP_os/src/yarp/os/PortWriterBufferBase.cpp#L71-L77. In particular 
```
            // oops, there is already a prepared and unwritten
            // object.  best remove it.
```
so I suppose that the multiple ``prepare`` case is already handled.

So feel free to proceed as you prefer

PR #782, opened 2023-12-14
Implement LinearSpline in Math component

GiulioRomualdi:
This PR implements the `LinearSpline` class in the `math` component, where `LinearSpline` is a [piecewise linear function](https://en.wikipedia.org/wiki/Piecewise_linear_function) of the form. 
```math
s(t) = a_0 + a_1 t
```
where $t \in [0, T_i]$ and $T_i$ is the duration of the i-th polynomial.


paolo-viceconte (approved):
Looks good to me

PR #784, opened 2023-12-15
Add the support of QP problems with no constraint in QPInverseKinematics and QPTSID 

GiulioRomualdi:
This addresses https://github.com/ami-iit/bipedal-locomotion-framework/issues/723


davidegorbani (approved):
It seems good to me!

PR #785, opened 2023-12-18
Add the possibility to set the SwingFootPlanner position and orientation tolerances as configuration parameters

GiulioRomualdi:
As per the title, this PR adds the possibility of setting the orientation and position tolerances for the swing foot planner. These tolerances are used to check if the position of a contact changes when it is already active. Having a tunable tolerance will make the planner more robust when the contact location are computed by an optimization procedure, for instance `sqpmethod`

### DIFF
--- a/src/Planners/src/SwingFootPlanner.cpp
+++ b/src/Planners/src/SwingFootPlanner.cpp
@@ -176,9 +176,8 @@ bool SwingFootPlanner::setContactList(const ContactList& contactList)
     {
         if (!(activeContact->pose.translation() - activeContactNewList->pose.translation())
                  .isZero(m_positionTolerance)
-            || !(activeContact->pose.asSO3() - activeContactNewList->pose.asSO3())
-                    .coeffs()
-                    .isZero(m_orientationTolerance))
+            || (activeContact->pose.asSO3() - activeContactNewList->pose.asSO3()).coeffs().norm()
+                   > m_orientationTolerance)
         {
             log()->error("{} The pose of the contact in the new contact list is different from the "
                          "pose of the contact in the original contact list. Given the contact "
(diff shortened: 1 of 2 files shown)


src/Planners/src/SwingFootPlanner.cpp
```diff
@@ -157,23 +169,33 @@ bool SwingFootPlanner::setContactList(const ContactList& contactList)
...
         && activeContactNewList != contactList.cend() // i.e., the new list contains an active
                                                       // contact
     )
     {
-        if (!(activeContact->pose - activeContactNewList->pose).coeffs().isZero(tolerance))
+        if (!(activeContact->pose.translation() - activeContactNewList->pose.translation())
+                 .isZero(m_positionTolerance)
+            || !(activeContact->pose.asSO3() - activeContactNewList->pose.asSO3())
+                    .coeffs()
+                    .isZero(m_orientationTolerance))
```
S-Dafarra, line 181: Is ``asSO3`` returning a rotation matrix? If so, the rotation tolerance is not really an angle, right?

S-Dafarra: Rather than the norm, what about using 
```math
\arccos\left(\frac{\text{trace}\left(R_1^\top R_2\right) - 1}{2}\right)
```

See https://en.wikipedia.org/wiki/Rotation_matrix#Determining_the_angle

GiulioRomualdi: I think its the same 

The equation you provide computes the angle of the angle axis representation, on the other hand, in case of SO(3) the log returns the axis angle representation ( see also
Section 2.1. in https://cvg.cit.tum.de/_media/members/demmeln/nurlanov2021so3log.pdf)

See: 
- https://en.wikipedia.org/wiki/Axis%E2%80%93angle_representation#Log_map_from_SO(3)_to_%F0%9D%94%B0%F0%9D%94%AC(3)
- Example 4 of https://arxiv.org/pdf/1812.01537.pdf

S-Dafarra: Ah sorry, I did not understand that you wanted to compute the norm of the log and not the norm of the difference. That makes sense then

S-Dafarra (approved):

src/Planners/src/SwingFootPlanner.cpp
GiulioRomualdi: ops :( sorry my fault!

PR #790, opened 2024-01-08
Avoid to call BufferedPort::prepare every time VectorsCollectionServer::populateData is called

GiulioRomualdi:


GiulioRomualdi:
If this code approach is fine I will update the walking-controller and the application in blf that streams the data

GiulioRomualdi:
I tested the modification on the robot, and everything seems to be working as expected. I'm going to finalize the PR this morning, merge it and release the software

GiulioRomualdi:
In windows the IK test fails with this nice error: https://github.com/ami-iit/bipedal-locomotion-framework/actions/runs/7582705428/job/20652803791?pr=790#step:14:312

GiulioRomualdi:
Let's see if the tests are bugged https://github.com/ami-iit/bipedal-locomotion-framework/pull/797

GiulioRomualdi:
The test failing is the QP IK with gravity task and for sure it is not related to this PR so I'm going to merge it. However I will open an issue for this since there could be something not consider cc @EhsanRanjbari and @S-Dafarra

S-Dafarra:
> The test failing is the QP IK with gravity task and for sure it is not related to this PR so I'm going to merge it. However I will open an issue for this since there could be something not consider cc @EhsanRanjbari and @S-Dafarra

It is very weird, that tests started failing all of a sudden and only on Windows. It also fails quite largely, so it is not just a problem of tolerances

PR #794, opened 2024-01-09
Real time logging

nicktrem:
Added the ability for the YARPRobotLoggerDevice to publish data to a YARP port for realtime logging capabilities.

### DIFF
--- a/devices/YarpRobotLoggerDevice/include/BipedalLocomotion/YarpRobotLoggerDevice.h
+++ b/devices/YarpRobotLoggerDevice/include/BipedalLocomotion/YarpRobotLoggerDevice.h
@@ -161,6 +161,7 @@ private:
     // for realtime data streaming
     bool streamRealTime{true};
     yarp::os::BufferedPort<yarp::os::Bottle> realtimeLoggingPort;
+    BipedalLocomotion::YarpUtilities::VectorsCollection* externalSignalCollection;
 
     std::mutex m_bufferManagerMutex;
     robometry::BufferManager m_bufferManager;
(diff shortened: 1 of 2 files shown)


devices/YarpRobotLoggerDevice/include/BipedalLocomotion/YarpRobotLoggerDevice.h
```diff
@@ -14,6 +14,7 @@
 #include <unordered_set>
 #include <atomic>
 #include <fstream>
+#include <jsoncpp/json/json.h>
```
S-Dafarra, line 17: A preliminary comment. How did you include this dependency? I don't see any modification to the ``CMakeLists``, I am not sure this would be compiling.

Also, consider using ``#include <nlohmann/json.hpp>`` as in https://github.com/ami-iit/bipedal-locomotion-framework/blob/a405c4b96ce6b0265e8fa0556ff47b1d4f1e1791/src/Contacts/src/ContactListJsonParser.cpp#L11C1-L11C29 for example.
This dependency is found in https://github.com/ami-iit/bipedal-locomotion-framework/blob/a405c4b96ce6b0265e8fa0556ff47b1d4f1e1791/cmake/BipedalLocomotionFrameworkDependencies.cmake#L88-L90 and linked in https://github.com/ami-iit/bipedal-locomotion-framework/blob/a405c4b96ce6b0265e8fa0556ff47b1d4f1e1791/src/Contacts/CMakeLists.txt#L16
In this example, notealso that the ``Contact`` module is compiled only if the corresponding ``json`` dependency s found: https://github.com/ami-iit/bipedal-locomotion-framework/blob/a405c4b96ce6b0265e8fa0556ff47b1d4f1e1791/cmake/BipedalLocomotionFrameworkDependencies.cmake#L146-L148

GiulioRomualdi:
As a higher-level observation, it would be advisable not to stream the data outside of the logger. If you need to utilize it in the visualizer, a potential approach is to create a class that incorporates all the logger's logic and implements Python bindings. This class can then be used both in the ARP device for data logging and by the visualizer for real-time data checking. We can arrange a meeting to discuss this further.

nicktrem:
Closing pull request, has been replaced with additional features by #796

PR #797, opened 2024-01-19
Check that the size the gains and weights vectors are correct

GiulioRomualdi:
This should fix the issue we experienced on the robot two days ago.

### DIFF
--- a/src/IK/src/JointLimitsTask.cpp
+++ b/src/IK/src/JointLimitsTask.cpp
@@ -155,6 +155,14 @@ bool JointLimitsTask::initialize(
         return false;
     }
 
+    if (m_klim.size() != m_kinDyn->getNrOfDegreesOfFreedom())
+    {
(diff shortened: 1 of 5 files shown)


src/IK/src/JointLimitsTask.cpp
```diff
@@ -155,6 +155,14 @@ bool JointLimitsTask::initialize(
         return false;
     }
 
+    if (m_klim.size() != m_kinDyn->getNrOfDegreesOfFreedom())
+    {
+        log()->error("{} The size of the 'klim' parameter does not match with the number of "
```
S-Dafarra, line 160: Could you print also here the expected/given dimensions?

GiulioRomualdi: Done here: https://github.com/ami-iit/bipedal-locomotion-framework/pull/797/commits/05294ae9c67b460caa747c377078c7e6aebb1eab

S-Dafarra (approved):

PR #806, opened 2024-02-08
Add configuration files to log data from ergoCubGazeboV1_1

isorrentino:


GiulioRomualdi:
I was wondering if rhe failure has the same root if what we experienced in https://github.com/ami-iit/bipedal-locomotion-framework/issues/780

PR #815, opened 2024-03-02
add UnicyclePlanner unit test

mebbaid:
While using the unicycle planner with the `centroidal mpc` client app. Initially i was using the planner in `robotology`, then realized that the planner was added to bipedal (with some modifications). 

Trying to understand its structure, i typically rely on the unit tests codes and I realized there is not a unit test code for the UnicyclePlanner class. I am not sure if that omission is by design, but in case it is not, I added a  unit test. Feel free to discard or merge.


src/Planners/tests/UnicyclePlannerTest.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    const auto& lastLeftContact = output.left.rbegin()->pose;
+    REQUIRE(approxEqual(lastLeftContact, manif::SE3d({input.knots.back().x - 0.1, input.knots.back().y, 0.0}, Eigen::Quaterniond::Identity())));
+
+    if (saveDataTofile)
+    {
+        saveData(output, "UnicyclePlannerTestOutput.txt");
+    }
+
+}
```
S-Dafarra, line 115:
```suggestion
}

```

src/Planners/tests/UnicyclePlannerTest.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    return handler;
+}
+
+TEST_CASE("UnicyclePlannerTest")
+{
+    const double dT = 0.01;
+    const auto handler = params(dT);
+
+    bool saveDataTofile = true;
```
S-Dafarra, line 82:
```suggestion
    bool saveDataTofile = false;
```

S-Dafarra (commented):
Thanks a lot @mebbaid! Maybe we can disable the saving to file for the time being

mebbaid:
> Thanks a lot @mebbaid! Maybe we can disable the saving to file for the time being

Agreed, it was left there while I was testing. Resolved in https://github.com/ami-iit/bipedal-locomotion-framework/pull/815/commits/87c54e16d22c7e8477431a3edc4252bc647f5976

GiulioRomualdi (approved):

PR #816, opened 2024-03-03
Added askToStop and stop methods

rob-mau:


S-Dafarra:
Can you move `this->stop()` to the destructor?

Did this fix the issue you were having?

rob-mau:
> Did this fix the issue you were having?

The problem seems to still be present

GiulioRomualdi:
Hi @rob-mau, are you planning to work on this anytime soon?

rob-mau:
Hi @GiulioRomualdi. I've paused solving this issue. At the moment, I don't have any clues on how it could be fixed.

GiulioRomualdi:
Ok so since this might not be the solution, I'm going to close it. The issue https://github.com/ami-iit/bipedal-locomotion-framework/issues/812 will remain open since the problem is still there :)

PR #821, opened 2024-03-15
Add member function to set gravity vector in centroidal dynamics system

LoreMoretti:
This PR aims at providing a member function to set on the fly the gravity vector used in [CentroidalDynamics.h](https://github.com/ami-iit/bipedal-locomotion-framework/blob/master/src/ContinuousDynamicalSystem/include/BipedalLocomotion/ContinuousDynamicalSystem/CentroidalDynamics.h).

For example, when the reference frame is changed from the `world` one to a `local` one, and we want to write the centroidal dynamics in this new `local frame`, it might be needed to set a new gravity vector.

### DIFF
--- a/src/ContinuousDynamicalSystem/include/BipedalLocomotion/ContinuousDynamicalSystem/CentroidalDynamics.h
+++ b/src/ContinuousDynamicalSystem/include/BipedalLocomotion/ContinuousDynamicalSystem/CentroidalDynamics.h
@@ -157,7 +157,7 @@ public:
      * @param gravity the value of the gravity vector used in the system dynamics.
      * @return true in case of success, false otherwise.
      */
-    bool setGravityVector(const Eigen::Vector3d& gravity);
+    void setGravityVector(const Eigen::Ref<const Eigen::Vector3d>& gravity);
 };
 } // namespace ContinuousDynamicalSystem
 } // namespace BipedalLocomotion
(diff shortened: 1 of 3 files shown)


src/ContinuousDynamicalSystem/include/BipedalLocomotion/ContinuousDynamicalSystem/CentroidalDynamics.h
```diff
@@ -151,6 +151,13 @@ class CentroidalDynamics : public DynamicalSystem<CentroidalDynamics>
      * @return true in case of success, false otherwise.
      */
     bool setControlInput(const Input& controlInput);
+
+    /**
+     * Set the gravity vector of the dynamical system.
+     * @param gravity the value of the gravity vector used in the system dynamics.
+     * @return true in case of success, false otherwise.
+     */
+    bool setGravityVector(const Eigen::Vector3d& gravity);
```
GiulioRomualdi, line 160: This can return a void and it should take as input Eigen::Ref<const Eigen::Vector3d>

GiulioRomualdi (commented):
A minor comment :)

GiulioRomualdi:
The CI is not due to your change 😀

src/ContinuousDynamicalSystem/include/BipedalLocomotion/ContinuousDynamicalSystem/CentroidalDynamics.h
LoreMoretti: Thanks @GiulioRomualdi for the suggestion.

Done in [
6b0625e](https://github.com/ami-iit/bipedal-locomotion-framework/pull/821/commits/6b0625ebeeea900b196e2a971392c9c9cdd4007e)

GiulioRomualdi (approved):

PR #823, opened 2024-03-18
set gravity vector as an input argument of the CentroidalMPC

LoreMoretti:
The goal of this PR is to expose the gravity vector as an input argument of the `CentroidalMPC`.

In this way, the user can pass it as other input arguments (like `currentCOM`, `externalWrench`, ...).

User can leverage the `setState` member function, or the `setGravity` one, to pass the gravity vector.

Moreover, if the user does not specify it, by default it is set to `[0, 0, -9.81]`.


GiulioRomualdi (approved):

GiulioRomualdi:
Thank you @LoreMoretti! Merging!

PR #824, opened 2024-03-18
Add the possibility to set the exogenous signal for the IK::CoMTask

GiulioRomualdi:


src/IK/src/CoMTask.cpp
```diff
@@ -235,3 +245,14 @@ bool CoMTask::isValid() const
...
+bool CoMTask::setFeedback(Eigen::Ref<const Eigen::Vector3d> feedback)
+{
+    bool ok = true;
+    if (m_useExogenousFeedback)
+    {
+        ok = ok && m_R3Controller.setState(feedback);
+    }
+
+    return ok;
+}
```
S-Dafarra, line 258:
```suggestion
}

```

src/IK/src/CoMTask.cpp
```diff
@@ -235,3 +245,14 @@ bool CoMTask::isValid() const
...
+
+bool CoMTask::setFeedback(Eigen::Ref<const Eigen::Vector3d> feedback)
+{
+    bool ok = true;
+    if (m_useExogenousFeedback)
+    {
+        ok = ok && m_R3Controller.setState(feedback);
+    }
+
+    return ok;
```
S-Dafarra, line 257: Should it return false if ``m_useExogenousFeedback`` is false?

GiulioRomualdi: For the SE3Task we return true if the flag is false. Still it makes sense to return false

GiulioRomualdi: With a debug print

S-Dafarra: As discussed F2F, it may be worth returning ``true`` anyhow so that we can keep the call and just use the flag to turn the feature on and off

S-Dafarra (approved):

PR #850, opened 2024-06-12
Fix `VectorsCollectionClient.read_data(False)` to provide `collection` as output and avoid segmentation fault

FabioBergonti:
I was using `VectorsCollectionClient.read_data(False)` in Python, and when the `VectorsCollectionClient` was unable to retrieve data, it caused a segmentation fault. With this modification, the method returns `None` instead, preventing the segmentation fault.

### DIFF
--- a/bindings/python/YarpUtilities/src/VectorsCollection.cpp
+++ b/bindings/python/YarpUtilities/src/VectorsCollection.cpp
@@ -69,7 +69,10 @@ void CreateVectorsCollectionClient(pybind11::module& module)
              [](VectorsCollectionClient& impl)
                  -> BipedalLocomotion::YarpUtilities::VectorsCollectionMetadata {
                  BipedalLocomotion::YarpUtilities::VectorsCollectionMetadata metadata;
-                 impl.getMetadata(metadata);
+                 if (impl.getMetadata(metadata) == false)
+                 {
+                     throw ::pybind11::value_error("Impossible to retrieve the metadata.");
+                 }
                  return metadata;
              })
         .def("read_data",
--- a/bindings/python/YarpUtilities/src/VectorsCollection.cpp
+++ b/bindings/python/YarpUtilities/src/VectorsCollection.cpp
@@ -5,12 +5,13 @@
  * distributed under the terms of the BSD-3-Clause license.
  */
 
+#include <optional>
+#include <pybind11/eigen.h>
 #include <pybind11/pybind11.h>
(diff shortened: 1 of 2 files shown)


bindings/python/YarpUtilities/src/VectorsCollection.cpp
```diff
@@ -65,17 +66,23 @@ void CreateVectorsCollectionClient(pybind11::module& module)
...
-        [](VectorsCollectionClient& impl) -> BipedalLocomotion::YarpUtilities::VectorsCollectionMetadata
-        {
-            BipedalLocomotion::YarpUtilities::VectorsCollectionMetadata metadata;
-            impl.getMetadata(metadata);
-            return metadata;
-        })
+             [](VectorsCollectionClient& impl)
+                 -> BipedalLocomotion::YarpUtilities::VectorsCollectionMetadata {
+                 BipedalLocomotion::YarpUtilities::VectorsCollectionMetadata metadata;
+                 impl.getMetadata(metadata);
```
GiulioRomualdi, line 72: You can add a check if getMetadata returns true or false. If false you may throw an error

bindings/python/YarpUtilities/src/VectorsCollection.cpp
```diff
@@ -65,17 +66,23 @@ void CreateVectorsCollectionClient(pybind11::module& module)
...
+                bool shouldWait) -> std::optional<std::map<std::string, std::vector<double>>> {
+                 BipedalLocomotion::YarpUtilities::VectorsCollection* collection
+                     = impl.readData(shouldWait);
+                 if (collection == nullptr)
+                 {
+                     // Return an empty optional if the collection is not available
+                     return std::nullopt;
+                 }
+                 return collection->vectors;
              });
```
GiulioRomualdi, line 86: You can directly return the connection

GiulioRomualdi (commented):
Thank you @FabioBergonti !

bindings/python/YarpUtilities/src/VectorsCollection.cpp
FabioBergonti: Done in 4f975ab8fdeff6aec2c5020d51a179a33c600836
Returning the collection made it unnecessary to check if the collection was a `nullptr`. I removed the unnecessary lines. I tested it and did not experience any segmentation faults.

GiulioRomualdi (approved):

PR #860, opened 2024-06-28
add setControlModeAsync member function

LoreMoretti:
The goal of this PR is to add a member function to set the motor control mode in an asynchronous process, so that the invoking thread does not get blocked.

This feature might be handy when the switch of control mode is done within `Advanceable Runners` which have real time constraints and deadline misses checks.


S-Dafarra (approved):
Cool!

GiulioRomualdi (approved):

PR #867, opened 2024-07-23
Update blf-logger-with-audio.sh script to support different logger launch file

giotherobot:
In addition add default values to bot arguments.

### DIFF
--- a/devices/YarpRobotLoggerDevice/scripts/blf-logger-with-audio.sh
+++ b/devices/YarpRobotLoggerDevice/scripts/blf-logger-with-audio.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
-input_port_audio=$1
+input_port_audio=${1:-/icub/microphone/audio:o}
+launch_file=${2:-launch-yarp-robot-logger.xml}
 
 # Function to handle SIGINT signal
 function handle_sigint {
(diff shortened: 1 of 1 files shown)


devices/YarpRobotLoggerDevice/scripts/blf-logger-with-audio.sh
```diff
@@ -1,6 +1,32 @@
...
+        launch_file="$2"
+        shift
+        shift
+        ;;
+        *)
+        # Unknown option
+        echo "Unknown option: $1"
+        exit 1
+        ;;
+    esac
```
S-Dafarra, line 24: Could you add some comments about the available options?

S-Dafarra: Just to make myself clearer, I meant to specify which inputs are parsed, like
```
Available options:
-i, --input-port-audio, name of the yarp port to connect in order to log audio
-l, --launch-file, name of the file to launch for the logger yarprobotinterface
```
or something similar

S-Dafarra:
If I understood well, you always have to set at least one parameter, otherwise it shows the help text, right?

S-Dafarra (approved):

GiulioRomualdi (approved):

PR #868, opened 2024-07-24
Add method to get joint limits from control board

LoreMoretti:
This PR add a member function to get the joint limits of the actuated joints. Also the related python bindings are added.


src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpRobotControl.h
```diff
@@ -158,6 +158,13 @@ class YarpRobotControl : public IRobotControl
      */
     std::vector<std::string> getJointList() const final;
 
+    /**
+     * Get the actuated joints limits.
+     * @param lowerLimits vector to be filled with the lower limits of the actuated joints.
+     * @param upperLimits vector to be filled with the upper limits of the actuated joints.
+     */
+    void getJointLimits(Eigen::VectorXd& lowerLimits, Eigen::VectorXd& upperLimits) const final;
```
GiulioRomualdi, line 166:
```suggestion
    void getJointLimits(Eigen::Ref<Eigen::VectorXd> lowerLimits, Eigen::Ref<Eigen::VectorXd> upperLimits) const final;
```

src/RobotInterface/YarpImplementation/src/YarpRobotControl.cpp
```diff
@@ -820,3 +828,14 @@ bool YarpRobotControl::isValid() const
...
+void YarpRobotControl::getJointLimits(Eigen::VectorXd& lowerLimits,
+                                      Eigen::VectorXd& upperLimits) const
+{
+    lowerLimits.resize(m_pimpl->actuatedDOFs);
+    upperLimits.resize(m_pimpl->actuatedDOFs);
+    for (int i = 0; i < m_pimpl->actuatedDOFs; i++)
+    {
+        m_pimpl->controlLimitsInterface->getLimits(i, &lowerLimits[i], &upperLimits[i]);
+    }
+}
```
GiulioRomualdi, line 841:
```suggestion
}

```

src/RobotInterface/YarpImplementation/src/YarpRobotControl.cpp
```diff
@@ -820,3 +828,14 @@ bool YarpRobotControl::isValid() const
 {
     return m_pimpl->robotDevice != nullptr;
 }
+
+void YarpRobotControl::getJointLimits(Eigen::VectorXd& lowerLimits,
+                                      Eigen::VectorXd& upperLimits) const
+{
+    lowerLimits.resize(m_pimpl->actuatedDOFs);
+    upperLimits.resize(m_pimpl->actuatedDOFs);
```
GiulioRomualdi, line 836: Using the eigen::Ref you cannot resize it anymore so you should check that the size is correct.

GiulioRomualdi (changes requested):

src/RobotInterface/YarpImplementation/src/YarpRobotControl.cpp
LoreMoretti: Got it. Indeed this is why I was not using eigen::Ref.

How do you suggest to do the check?

Just an assert, like:

```
assert(lowerLimits.size() == m_pimpl->actuatedDOFs)
assert(upperLimits.size() == m_pimpl->actuatedDOFs)
```

Or you prefer to change also the return type from `void` to `bool` and raise an error if size do not match?

GiulioRomualdi: I would return a bool

LoreMoretti: done with https://github.com/ami-iit/bipedal-locomotion-framework/pull/868/commits/a2809e6c8ec68a39c3c4b80b322bd4353560420a

src/RobotInterface/YarpImplementation/include/BipedalLocomotion/RobotInterface/YarpRobotControl.h
LoreMoretti: done with https://github.com/ami-iit/bipedal-locomotion-framework/pull/868/commits/a2809e6c8ec68a39c3c4b80b322bd4353560420a

GiulioRomualdi (approved):

PR #869, opened 2024-07-25
Added missing logData for vectorSignals and TextLogging

S-Dafarra:
cc @GiulioRomualdi @carloscp3009 @giotherobot

### DIFF
--- a/devices/YarpRobotLoggerDevice/app/robots/ergoCubSN001/yarp-robot-logger.xml
+++ b/devices/YarpRobotLoggerDevice/app/robots/ergoCubSN001/yarp-robot-logger.xml
@@ -19,7 +19,6 @@ BSD-3-Clause license. -->
   <param name="port_prefix">/yarp-robot-logger</param>
   <param name="text_logging_subnames">("ergoCubSN001/yarprobotinterface")</param>
   <param name="code_status_cmd_prefixes">("ssh ergocub@10.0.2.2" "ssh ergocub@10.0.2.3")</param>
-  <param name="maximum_admissible_time_step">60.0</param>
 
   <group name="Telemetry">
     <param name="save_period">1800.0</param>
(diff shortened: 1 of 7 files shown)


S-Dafarra:
Before setting it ready for review, I want to understand why in a dataset we ended up having all the timestamps for the RGBD cameras, and only one data point for all the other signals. Maybe a race condition somewhere? 🤔

S-Dafarra:
> Before setting it ready for review, I want to understand why in a dataset we ended up having all the timestamps for the RGBD cameras, and only one data point for all the other signals. Maybe a race condition somewhere? 🤔

This was due to the check on the time difference between two acquisitions.
In any case, with https://github.com/ami-iit/bipedal-locomotion-framework/pull/869/commits/c460237de7cb0206f0f93f517477aff4afc26e44 I reduced the chances that the ``run`` method remains stuck due to the waiting to some mutex for the exogenous signals.

I wanted to check the same with the text and I noticed that:
1. in theory it is not necessary to lock any mutex. Eventual race conditions due to the connections should be handled by ``yarp`` itself internally
2. there was a missing call to save the data also here

I implemented a fix for these two points in https://github.com/ami-iit/bipedal-locomotion-framework/pull/869/commits/cb5531d7a61414dcb64bb5a113e0ce1597b4ba55

devices/YarpRobotLoggerDevice/app/robots/ergoCubSN001/yarp-robot-logger.xml
```diff
@@ -19,7 +19,7 @@ BSD-3-Clause license. -->
   <param name="port_prefix">/yarp-robot-logger</param>
   <param name="text_logging_subnames">("ergoCubSN001/yarprobotinterface")</param>
   <param name="code_status_cmd_prefixes">("ssh ergocub@10.0.2.2" "ssh ergocub@10.0.2.3")</param>
-  <param name="maximum_admissible_time_step">1.0</param>
+  <param name="maximum_admissible_time_step">60.0</param>
```
GiulioRomualdi, line 22: I would completely remove this parameter so we always get the data

S-Dafarra: Done in https://github.com/ami-iit/bipedal-locomotion-framework/pull/869/commits/11a95f79a307503389e11074145218b07ca1774c

S-Dafarra:
I think this is ready for review. There are a few additions I would like to implement (e.g. check if the metadata is ready before asking), but I prefer to do it in a subsequent PR

GiulioRomualdi (approved):

PR #887, opened 2024-09-13
Add optional initialization of base pose and feet pose in the Unicycle Trajectory Generator

LoreMoretti:
This PR add the feature to initialize the `unicycleTrajectoryGenerator` with the initial base pose and the relative pose of the feet (which mimics what is done in [TrajectoryGenerator.cpp](https://github.com/robotology/walking-controllers/blob/master/src/TrajectoryPlanner/src/TrajectoryGenerator.cpp) in the `generateFirstTrajectories` member functions)


LoreMoretti:
> Look good to me! I would suggest adding a basic test for the additions

Thanks @S-Dafarra.

Indeed, when writing the new tests and compiling them, I found out that the signatures:

```
bool Planners::UnicycleTrajectoryGenerator::initialize(
    std::weak_ptr<const ParametersHandler::IParametersHandler> handler,
    const manif::SE3d& leftToRightTransform)


bool Planners::UnicycleTrajectoryGenerator::initialize(
    std::weak_ptr<const ParametersHandler::IParametersHandler> handler,
    const Eigen::Ref<const Eigen::Vector3d>& initialBasePosition)
```

seem to be ambiguous.

The only way I found so far is to pass `initialBasePosition` as `const Eigen::Vector3d&` instead of `const Eigen::Ref<const Eigen::Vector3d>&`. I know this reduces flexibility, but I could not find another way.

S-Dafarra (approved):

GiulioRomualdi (approved):

GiulioRomualdi:
CI on conda fails with: https://github.com/ami-iit/bipedal-locomotion-framework/issues/896

PR #899, opened 2024-10-04
Add Periodic Thread class

LoreMoretti:
This class manages the creation and execution of periodic threads. 

User can define the tasks for their periodic thread just by inheriting from this class and overriding the `threadInit` and `run` function.


LoreMoretti:
Also `BipedalLocomotion::System::AdvanceableRunner` could inherit from `BipedalLocomotion::System::PeriodicThread`. But, I think it is better to eventually implement it in a separate PR.

PR #916, opened 2024-12-10
Add zmp generator and populate all outputs of the unycycle trajectory generator

LoreMoretti:
Some of the outputs of the `UnicycleTrajectoryGenerator` were not being populated.

This PR fixes it.

Moreover, the `ZMP generator` (and its outputs) is added to the `UnicycleTrajectoryGenerator`


S-Dafarra:
At the code level, it looks good. What I am a bit afraid of is that from outside it is not clear that the different trajectories are not coherent with each other. At the unicycle planner level, you are calling to different generators, so somehow you can expect them to be separate. Here instead, the output is in a single data structure. Maybe it would be useful to have a trigger to turn on or off this output?

LoreMoretti:
> At the code level, it looks good. What I am a bit afraid of is that from outside it is not clear that the different trajectories are not coherent with each other. At the unicycle planner level, you are calling to different generators, so somehow you can expect them to be separate. Here instead, the output is in a single data structure. Maybe it would be useful to have a trigger to turn on or off this output?

You're right! It makes totally sense.

With the commit [ecbd51f](https://github.com/ami-iit/bipedal-locomotion-framework/pull/916/commits/ecbd51ffa8afd51fca28cf5eca9c8b51ee7d432b) I tried to make the `dcmTrajectory` and the `zmpTrajectory` optional (and enabled by boolean flags).

I like the fact that `std::optional` makes the code cleaner in my opinion.

There's the drawback. This change makes the blf Unicycle Trajectory Generator not backward compatible. But I think the number of users is still low.

Let me know what you think.

S-Dafarra:
> There's the drawback. This change makes the blf Unicycle Trajectory Generator not backward compatible. But I think the number of users is still low.
> 
> Let me know what you think.

It should be fine by changing the version number. On the other hand, another alternative is to apply this change only for the data you added

src/Planners/src/UnicycleTrajectoryGenerator.cpp
```diff
@@ -276,6 +283,14 @@ bool Planners::UnicycleTrajectoryGenerator::initialize(
                                                           initialBasePosition,
                                                           leftToRightTransform);
 
+    // Get the generator being used and instantiate optional outputs
+    ok = ok && loadParamWithFallback("use_zmp_generator", m_useZMPGenerator, false);
+    m_useDCMGenerator = !m_useZMPGenerator;
```
S-Dafarra, line 288: I would still give the possibility of using both of them at the same time

LoreMoretti: Done with [4e1a9a5](https://github.com/ami-iit/bipedal-locomotion-framework/pull/916/commits/4e1a9a5a14c2fb267a3b37847c5cc92ca25276ce)

src/Planners/include/BipedalLocomotion/Planners/UnicycleTrajectoryPlanner.h
```diff
@@ -194,45 +210,47 @@ class BipedalLocomotion::Planners::UnicycleTrajectoryPlanner final
...
+     * | `comHeight`               |     double        |        0.70       |       -         |    CoM height in double support phase                                                |
+     * | `comHeightDelta`          |     double        |        0.01       |       -         |    Delta to add to CoM heinght in Single support phases                              |
+     * | `leftZMPDelta`            | list of 2 doubles |         -         |   (0.0  0.0)    | Local ZMP reference: delta wrt center frame of the foot                              |
+     * | `rightZMPDelta`           | list of 2 doubles |         -         |   (0.0  0.0)    | Local ZMP reference: delta wrt center frame of the foot                              |
+     * | `leftZMPInitSwitchDelta`  | list of 2 doubles |    (0.0  0.0)     |   (0.0  0.0)    | Local ZMP reference: delta wrt center frame when the switch to the other foot begin  |
+     * | `rightZMPInitSwitchDelta` | list of 2 doubles |    (0.0  0.0)     |   (0.0  0.0)    | Local ZMP reference: delta wrt center frame when the switch to the other foot begin  |
```
S-Dafarra, line 250: I would suggest specifying this is used only for the ZMP generator

LoreMoretti: Sure, done with [52e42e3](https://github.com/ami-iit/bipedal-locomotion-framework/pull/916/commits/52e42e31f0dc10775f1160cf3707d907dce23f4e)

S-Dafarra (approved):

CHANGELOG.md
```diff
@@ -36,6 +40,7 @@ All notable changes to this project are documented in this file.
 - Bug fix of `prepare_data` method calling in `joints-grid-position-tracking` (https://github.com/ami-iit/bipedal-locomotion-framework/pull/895)
 - Fix the normal force limit constraint in the `CentroidalMPC` (https://github.com/ami-iit/bipedal-locomotion-framework/pull/898)
 - Fix loading of Python bindings on Windows when installed in arbitrary directory (https://github.com/ami-iit/bipedal-locomotion-framework/pull/905)
+- Fix outputs of `UnicycleTrajectoryGenerator` (https://github.com/ami-iit/bipedal-locomotion-framework/pull/916)
```
GiulioRomualdi, line 43: I think this has to be moved up

examples/UnicycleTrajectoryGenerator/main.cpp
```diff
@@ -205,23 +206,24 @@ int main(int argc, char* argv[])
         w_H_left = leftFootPlanner.getOutput().transform;
         w_H_right = rightFootPlanner.getOutput().transform;
 
+        // get the DCM trajectory from the unicycle trajectory generator
+        auto dcmPosition = output.dcmTrajectory.position;
+        auto dcmVelocity = output.dcmTrajectory.velocity;
```
GiulioRomualdi, line 211: I think this should avoid new memory allocation 
```suggestion
        const auto& dcmPosition = output.dcmTrajectory.position;
        const auto& dcmVelocity = output.dcmTrajectory.velocity;
```

examples/UnicycleTrajectoryGenerator/main.cpp
```diff
@@ -239,8 +241,8 @@ int main(int argc, char* argv[])
             Ydcm(i) = dcmPosition[i][1];
         }
 
-        // log()->info("[main] DCM x: {}", Xdcm.transpose());
-        // log()->info("[main] DCM y: {}", Ydcm.transpose());
+        log()->debug("[main] DCM x: {}", Xdcm.transpose());
+        log()->debug("[main] DCM y: {}", Ydcm.transpose());
```
GiulioRomualdi, line 245:
```suggestion
```

src/Planners/src/UnicycleUtilities.cpp
```diff
@@ -145,8 +144,7 @@ namespace Conversions
 {
 void convertToBLF(const iDynTree::Transform& input, manif::SE3d& output)
 {
-    output.asSO3() = iDynTree::toEigen(input.getRotation().asQuaternion());
-    output.translation(iDynTree::toEigen(input.getPosition()));
+    output = BipedalLocomotion::Conversions::toManifPose(input);
```
GiulioRomualdi, line 147: Perhaps this can be be removed since there is already `BipedalLocomotion::Conversions::toManifPose`

GiulioRomualdi (commented):
Some minor changes

src/Planners/src/UnicycleUtilities.cpp
LoreMoretti: The reason why I added it, is because I used it as an helper function, in the following function templates:

https://github.com/ami-iit/bipedal-locomotion-framework/blob/c2bae6f18a6ad136c0f9cddae14027e951972146/src/Planners/include/BipedalLocomotion/Planners/UnicycleUtilities.h#L111-L124

If I remove `void convertToBLF(const iDynTree::Transform& input, manif::SE3d& output)`, I think I should modify the template function as:

- perform a check on `<From>` type.
- if it is `iDynTree::Transform&` then call `BipedalLocomotion::Conversions::toManifPose` instead of `void convertToBLF(const iDynTree::Transform& input, manif::SE3d& output)`.

Do you have other better alternatives?
Is it worth the change?

GiulioRomualdi: You convinced me!

GiulioRomualdi (approved):

PR #927, opened 2025-01-28
Enable the setting of the measured feet transform for the Unicycle Trajectory Generator

LoreMoretti:
This PR adds a member function to set the feet measured transform in the Unicycle Trajectory Generator

The feet measured transforms are used by the unicycle planner when regenerating the trajectory.

### DIFF
--- a/src/Planners/include/BipedalLocomotion/Planners/UnicycleTrajectoryGenerator.h
+++ b/src/Planners/include/BipedalLocomotion/Planners/UnicycleTrajectoryGenerator.h
@@ -175,9 +175,9 @@ public:
     std::chrono::nanoseconds getSamplingTime() const;
 
     /**
-     * @brief Set the measured transform to use when regenerating the trajectory.
+     * @brief Set the feet transforms to use when regenerating the trajectory.
      */
-    void setMeasuredTransform(const manif::SE3d& measuredTransform);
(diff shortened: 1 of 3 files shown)


src/Planners/include/BipedalLocomotion/Planners/UnicycleTrajectoryGenerator.h
```diff
@@ -174,6 +174,11 @@ class BipedalLocomotion::Planners::UnicycleTrajectoryGenerator final
      */
     std::chrono::nanoseconds getSamplingTime() const;
 
+    /**
+     * @brief Set the measured transform to use when regenerating the trajectory.
+     */
+    void setMeasuredTransform(const manif::SE3d& measuredTransform);
```
S-Dafarra, line 180: Measured transform of what?

LoreMoretti: Right question, I have modified a bit the code with [a97a477](https://github.com/ami-iit/bipedal-locomotion-framework/pull/927/commits/a97a4779c3b9f6c8f20be0a31c09513f007e7051). Now the user should pass both feet transforms.

GiulioRomualdi (approved):

PR #989, opened 2025-06-06
Added possibility to avoid logging code status, robot and cameras

S-Dafarra:
I have touched the logger code to allow the following:
- avoid logging the code status
- avoid logging the cameras with a single parameter
- avoid logging the robot.

The last point has been the trickiest, since most of the preparation work was done in the ``attachAll`` method. When not using the robot nor the cameras, it would be possible not to have any device attached. Hence, I had to make sure that the logger was properly started even when the ``attachAll`` method was not called.

I also took the occasion to introduce a few auxiliary methods to make the opening and preparation phases cleaner

PR #997, opened 2025-06-26
Add possibility to log frames in the logger device

S-Dafarra:
This is on top of #989 

This PR allows storing the value of some frames streamed from the transform server. 

The device to read the transforms is created internally and not attached, so the logger is able to start even if the transform server is not running.

The frames are obtained automatically given some specified parent frames.


GiulioRomualdi (approved):
Nice!!

PR #1005, opened 2025-10-02
Reset EMAWithLimits with the current position

GiulioRomualdi:
This pull request updates the `EMAWithLimits` reset by setting the initial value


S-Dafarra:
I suppose that the tests had to be updated too

PR #1008, opened 2025-10-14
Refactor VectorsCollection and YarpRobotLogger to enable runtime update of exogenous signals

xela-95:
This pull request introduces incremental metadata versioning and synchronization improvements for the `VectorsCollection` infrastructure, primarily to support metadata updates between clients and servers. The changes add version tracking to both metadata and data collections, enable clients to detect and retrieve only new metadata, and ensure that metadata for robot and exogenous signals is handled more accurately and efficiently in real-time (RT) logging scenarios.

Key changes include:

**Incremental Metadata Versioning and Synchronization**

* Introduced a `version` field to `VectorsCollectionMetadata` and `VectorsCollection`, updated their Python bindings to support versioning, and exposed the version as a property in the API.
* Added incremental metadata retrieval in `VectorsCollectionServer` via a new `getMetadataIncremental(fromVersion)` method, which returns only metadata added since a given version. The server now tracks finalized metadata versions and key histories.
* Enhanced `VectorsCollectionClient` to cache metadata and version numbers, detect when new metadata is available (via `isNewMetadataAvailable()`), and update only when necessary. Reading data now updates the metadata availability flag if a newer version is detected.
* Added new unit tests `VectorsCollectionTest` to ensure the correctness of the new versioning and incremental update features.

**Real-Time Logging and Metadata Handling**

* Improved logic in `YarpRobotLoggerDevice` for populating and finalizing metadata for robot and exogenous signals, ensuring that metadata is only added when available and that new metadata versions are finalized and logged appropriately. This includes more robust handling of cases where metadata is not yet available.
* Refactored and clarified the RT streaming and exogenous signal preparation code to ensure metadata is only populated for robot signals in the correct context and to avoid redundant or premature channel additions.

### DIFF
--- a/devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp
+++ b/devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp
@@ -1435,8 +1435,6 @@ bool BipedalLocomotion::YarpRobotLoggerDevice::prepareCameraLogging()
 
 bool BipedalLocomotion::YarpRobotLoggerDevice::prepareRTStreaming()
 {
-    // Only populate the metadata for robot signals (for exogenous signals this is done in the
-    // lookForExogenousSignals method)
     constexpr auto logPrefix = "[YarpRobotLoggerDevice::prepareRTStreaming]";
     if (!m_sendDataRT)
     {


xela-95:
The tests on Linux are failing due to yarp server address conflicts:

```
Error:  |yarp.os.Port|/test/server_17067/measures:o| Port /test/server_17067/measures:o failed to activate at tcp://127.0.0.1:10004/ (address conflict)
Error: 0-15 14:24:55.132] [thread: 15536] [blf] [error] [VectorsCollectionServer::initialize] Unable to open the port named /test/server_17067/measures:o.
```

Before I was initializing and destroying the yarp network for each `TEST_CASE_METHOD` through a fixture, now I'm trying to use `Catch::EventListenerBase` to have a global fixture but still doesn't work as expected. Locally the test passes.

xela-95:
> I started checking the server/client code and it looks great, well done!
> 
> I still need to check the device modifications and the test

Thanks! I was waiting to set the ready for review because I still have to solve the issue with the yarp addrress coinflicts errors in the tests I introduced, but thanks for the efficiency in the review :)

xela-95:
> The error you reported is different, as this is not related to blf code you can probably just add the suppression to https://github.com/ami-iit/bipedal-locomotion-framework/blob/master/cmake/valgrind-linux.supp .

Thanks! with the suppression added in https://github.com/ami-iit/bipedal-locomotion-framework/pull/1008/commits/b91c51713b4c7c3f16a85ff2a20a7f1bec6e9ca3 it now passes locally

xela-95:
> > The issue is that they are basically the same test, so setting a custom YARP_PORT_RANGE is not effective
> 
> A possible idea is to use the PID to decide which `YARP_PORT_RANGE` to set, or at the CMake level to something like:
> 
> ```cmake
> add_bipedal_test(
>     NAME VectorsCollectionTest
>     SOURCES VectorsCollectionTest.cpp
>     LINKS BipedalLocomotion::YarpUtilities BipedalLocomotion::VectorsCollection
>     )
> 
> if(TEST VectorsCollectionTest)
>   set_property(TEST VectorsCollectionTest APPEND PROPERTY ENVIRONMENT "YARP_PORT_RANGE=20000")
> endif()
> 
> if(TEST memcheck_VectorsCollectionTest)
>   set_property(TEST memcheck_VectorsCollectionTestAPPEND PROPERTY ENVIRONMENT "YARP_PORT_RANGE=21000")
> endif()
> ```

ok I can try that, I was going to disable the valgrind check on that test 😅

devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp
```diff
@@ -827,33 +827,55 @@ bool YarpRobotLoggerDevice::addChannel(const std::string& nameKey,
...
+            std::string prefix = signal.signalName + treeDelim;
+            if (nameKey.rfind(prefix, 0) == 0)
             {
-                log()->error("[Realtime logging] Failed to populate the metadata for the signal {}",
-                             rtNameKey);
-                return false;
+                isExogenous = true;
+                break;
             }
+        }
```
S-Dafarra: Honestly, I did not understand this logic. All the elements of ``m_vectorsCollectionSignals`` should be exogenous signals

devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp
```diff
@@ -1624,13 +1595,45 @@ void YarpRobotLoggerDevice::lookForExogenousSignals()
                                 "metadata for the vectors collection signal named: {}",
                                 name);
 
-                    if (!signal.client.getMetadata(signal.metadata))
+                    // TODO: only get incremental metadata updates
+                    // if (!signal.client.isNewMetadataAvailable()
+                    //     && signal.client.getMetadata(signal.metadata))
```
S-Dafarra: Is this commented code a leftover?

devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp
xela-95: Agree! I'll centralize this in a single method

xela-95: Done in 1c14c1caf35ff88b37db65a05b31f9e73f346bd9

devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp
```diff
@@ -1422,6 +1435,8 @@ bool BipedalLocomotion::YarpRobotLoggerDevice::prepareCameraLogging()
 
 bool BipedalLocomotion::YarpRobotLoggerDevice::prepareRTStreaming()
 {
+    // Only populate the metadata for robot signals (for exogenous signals this is done in the
+    // lookForExogenousSignals method)
```
S-Dafarra:
```suggestion
```
I think this is no more necessary

xela-95:
With @S-Dafarra @carloscp3009 we tested this PR branch on ergocub SN002 and also ran the walking module. The visualizer and logger worked as expected.

We were able to:
-  stream data in real time
- refresh the list of signals with exogenous signals launched after the logger and visualizer started (like the walking module data)

Moreover we tested that the MAT file saved by the logger contained all the data visualized in real-time, including the cameras.

xela-95:
rebased on master

PR #1044, opened 2026-06-29
Add multi-threaded video encoding to YarpRobotLoggerDevice

GiulioRomualdi:
Implement a producer-consumer pattern for video encoding to enhance performance and manage frame processing efficiently.


S-Dafarra:
Hi @GiulioRomualdi, if I understood correctly, with this PR, the saving of frames is done in another thread using a queue. This thread is spawned inside the ``recordVideo``. This method is also in another thread: https://github.com/gbionics/bipedal-locomotion-framework/blob/master/devices/YarpRobotLoggerDevice/src/YarpRobotLoggerDevice.cpp#L1537.

If I understood correctly, if the saving is not keeping up with the generation of the frames, the frame is stored in a queue. This should be the main difference compared to before, where, if we were not keeping up, a warning was printed. For what I understood, the queue is anyhow limited, and if we fill it, we start dropping frame.

Which improvement did you notice in practice?

PR #1052, opened 2026-08-21
Set reference position for a subset of joints

davidegorbani:
This PR adds 2 methods in `DinRailRobotControl` and in `IRobotControl`  to set only a subset of joints, specifying the joint indices or the joint names.

### DIFF
--- a/src/RobotInterface/DinRailImplementation/src/DinRailRobotControl.cpp
+++ b/src/RobotInterface/DinRailImplementation/src/DinRailRobotControl.cpp
@@ -203,6 +203,12 @@ bool DinRailRobotControl::setDriver(std::shared_ptr<yarp::dev::PolyDriver> robot
     m_pimpl->stiffnessBuffer.resize(m_pimpl->actuatedDOFs, 0.0);
     m_pimpl->dampingBuffer.resize(m_pimpl->actuatedDOFs, 0.0);
 
+    m_pimpl->subsetPos.reserve(m_pimpl->actuatedDOFs);
+    m_pimpl->subsetVel.reserve(m_pimpl->actuatedDOFs);
+    m_pimpl->subsetTorque.reserve(m_pimpl->actuatedDOFs);
+    m_pimpl->subsetStiffness.reserve(m_pimpl->actuatedDOFs);
+    m_pimpl->subsetDamping.reserve(m_pimpl->actuatedDOFs);
+
     return true;
 }
 
--- a/src/RobotInterface/DinRailImplementation/src/DinRailRobotControl.cpp
+++ b/src/RobotInterface/DinRailImplementation/src/DinRailRobotControl.cpp
@@ -38,8 +38,15 @@ struct DinRailRobotControl::Impl
      */
     std::vector<double> posBuffer;
     std::vector<double> velBuffer;
+    std::vector<double> torqueBuffer;
     std::vector<double> stiffnessBuffer;
     std::vector<double> dampingBuffer;
+
+    /**
+     * True after the first successful full setImpedanceSetPoints() call.
+     * The subset overload requires the cache to be initialised before use.
+     */
+    bool cacheInitialized{false};
 };
 
 DinRailRobotControl::DinRailRobotControl()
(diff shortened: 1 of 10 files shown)


src/RobotInterface/DinRailImplementation/src/DinRailRobotControl.cpp
```diff
@@ -38,8 +38,15 @@ struct DinRailRobotControl::Impl
...
     std::vector<double> velBuffer;
+    std::vector<double> torqueBuffer;
     std::vector<double> stiffnessBuffer;
     std::vector<double> dampingBuffer;
+
+    /**
+     * True after the first successful full setImpedanceSetPoints() call.
+     * The subset overload requires the cache to be initialised before use.
+     */
+    bool cacheInitialized{false};
```
traversaro: This can be removed?

davidegorbani: Done in https://github.com/gbionics/bipedal-locomotion-framework/pull/1052/commits/134cf445658b1e7a7e1a020b1b4acf06c2af64f6.

traversaro (approved):

src/RobotInterface/YarpImplementation/src/YarpRobotControl.cpp
```diff
@@ -717,6 +733,213 @@ struct YarpRobotControl::Impl
...
+                const double jointError = std::abs(jointValues[i] - this->subsetCurrentPos[i]);
+                double scaling, minVelocity;
+                if (this->jointsTypeList[jointIndices[i]] == JointType::REVOLUTE)
+                {
+                    scaling = 180.0 / M_PI;
+                    minVelocity = 3.0; // deg/s
+                } else
+                {
+                    scaling = 1.0;
+                    minVelocity = 10.0; // mm/s
```
S-Dafarra: These are a bit magic numbers. Consider defining them in constant expressions

S-Dafarra: I am referring to the ``minVelocity`` both for the revolute and linear case

src/RobotInterface/YarpImplementation/src/YarpRobotControl.cpp
```diff
@@ -950,3 +1173,46 @@ bool YarpRobotControl::getJointLimits(Eigen::Ref<Eigen::VectorXd> lowerLimits,
...
+
+bool YarpRobotControl::setReferences(
+    Eigen::Ref<const Eigen::VectorXd> desiredJointValues,
+    const std::vector<std::string>& jointNames,
+    const IRobotControl::ControlMode& controlMode,
+    std::optional<Eigen::Ref<const Eigen::VectorXd>> currentJointValues)
+{
+    constexpr auto errorPrefix = "[YarpRobotControl::setReferences]";
+
+    std::lock_guard<std::mutex> lock(m_pimpl->subsetMutex);
```
GiulioRomualdi, line 1204: I see that you added a mutex for protecting the setReference. I see your point since it is usefull in case two instance of YarpRobotControl or the dinrail version are created. If this is the case should we add the mutex also for the classical setReferences methods?

cc @traversaro

src/RobotInterface/YarpImplementation/src/YarpRobotControl.cpp
davidegorbani: I fixed it with https://github.com/gbionics/bipedal-locomotion-framework/pull/1052/commits/a8ea1e491003534aa78b3adcb2fda7e85db1bc8c and defined them as constant expressions for both the `setReferences` and `setReferencesSubset` methods.

src/RobotInterface/YarpImplementation/src/YarpRobotControl.cpp
traversaro: I was the one that suggested the mutex for protecting the `subset[...]` buffers. But indeed, there is nothing protecting the underlying YARP device itself, so adding a mutex could definitely make sense.

PR #1053, opened 2026-08-24
Fix  IK::JointLimitsTask initialization and update logic

GiulioRomualdi:
This PR fixes the a bug in the computation of the joint position limit in the IK

cc @camillagallina @RiccardoGrieco and @LudovicaDanovaro


camillagallina (approved):
LGTM

S-Dafarra (approved):
Great catch!

PR #1056, opened 2026-09-04
Add dinrail ParametersHandler implementation

traversaro:
This add a ParametersHandler implementation that uses dinrail::Parameters class. It also ports the tests that we added in dinrail to ensure that `YarpImplementation` and `DinRailImplementation` have a consistent behaviour.


src/ParametersHandler/tests/CMakeLists.txt
```diff
@@ -28,3 +28,26 @@ if(FRAMEWORK_COMPILE_TomlImplementation)
...
+  add_bipedal_test(
+    NAME ParametersHandlerDinRail
+    SOURCES ParametersHandlerDinRailTest.cpp
+    LINKS BipedalLocomotion::ParametersHandlerDinRailImplementation)
+
+endif()
+
+if(FRAMEWORK_COMPILE_DinRailImplementation AND FRAMEWORK_COMPILE_YarpImplementation)
+
+  find_package(dinrail REQUIRED COMPONENTS yarp)
```
S-Dafarra, line 43: I think this ``find_package`` should not be here. All dependencies are handled in the `BipedalLocomotionFrameworkDependencies.cmake` file

traversaro: This is a tricky one. The  `FRAMEWORK_COMPILE_DinRailImplementation` just depends on `dinrail`, not on the `YARP` component of `dinrail`, that is only a test dependency.

S-Dafarra: Maybe we could move the logic in https://github.com/gbionics/bipedal-locomotion-framework/blob/master/cmake/BipedalLocomotionFrameworkDependencies.cmake#L124-L147 and use another CMake variable

traversaro: Can you be more specific? I am not sure what you meant by "use another CMake variable"?

S-Dafarra: Apologies. I was thinking to something similar to ``FRAMEWORK_RUN_Valgrind_tests``, maybe named ``FRAMEWORK_RUN_dinrail_yarp_tests`` that is optional depending on whether dinrail is found with the yarp component.

